PR #33, opened 2013-10-23
Added shorthand for calling predict on a single point.

calder:


andreasnoack:
@kleinschmidt How does this fit in with other `predict` methods?

kleinschmidt:
If I understand correctly, this package doesn't directly know about ModelFrame-based models (that's all in DataFrames, and soon StatsModels), so this would need to change (to not refer to `mf.intercept`) and just work with a matrix. You can predict from a single point by passing a 1-row matrix, so I'm not sure how much more convenient this is (especially if people are using the formula-based interface).

andreasnoack:
This is quite old so I'll close here. If there is still a with for this functionality, it should probably be rewritten on top of what we have now.

PR #95, opened 2015-01-23
Remove NumericExtensions

### LINKED ISSUE #94: Remove NumericExtensions
I think @lindahua is no longer advising use of NumericExtensions, so we should try to get rid of it. I'm not sure whether to do it now or wait until 0.4 is released, since call overloading may make some things a bit nicer.

simonster:
Fixes  #94. The performance on the benchmark in `perf` is very slightly faster (~2%).

This also includes some miscellaneous changes from my working copy:
- Improved type stability
- Ability to specify starting coefficients
- Better handling of steps that result in invalid eta values
- Re-export Distributions instead of bringing it into Main

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,5 +1,6 @@
 julia 0.3-
 Distributions 0.4.6-
-NumericExtensions 0.6-
 NumericFuns 0.2.1-
 StatsBase 0.3.9+
+Reexport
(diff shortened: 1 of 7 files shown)


simonster:
Any comments before I merge this?

dmbates:
Thanks for looking at that, Simon.  LGTM.

PR #96, opened 2015-01-29
wrkresid is exported but is no defined

gragusa:
Add `wrkresid`.


nalimilan:
We no longer export that function.

PR #97, opened 2015-02-26
Make predict work with offsets

lendle:
`predict` should allow an offset to be specified. I can't think of a use case where you would want to fit with an offset and predict without, or vice versa, so I am requiring an offset to be specified if the model is fit with one (and vice versa).

Also `StatsBase.predict` for dataframes ([implemented here](https://github.com/JuliaStats/DataFrames.jl/blob/267a69754cb8fed06af4f2bd3475b17d75cf1cfe/src/statsmodels/statsmodel.jl#L68-L75)) currently does not allow additional args. I think changing it to

```julia
function StatsBase.predict(mm::DataFrameRegressionModel, df::AbstractDataFrame, args...; kwargs...)
    # copy terms, removing outcome if present
    newTerms = remove_response(mm.mf.terms)
    # create new model frame/matrix
    newX = ModelMatrix(ModelFrame(newTerms, df)).m
    predict(mm, newX, args...; kwargs...)
end
```
will do the trick.

When this gets merged, I'll open an issue or pull request for DataFrames.

### DIFF
--- a/src/glmfit.jl
+++ b/src/glmfit.jl
@@ -232,15 +232,14 @@ function scale(m::GeneralizedLinearModel, sqr::Bool=false)
 end
 
 ## Prediction function for GLMs
-function predict(mm::GeneralizedLinearModel, newX::AbstractMatrix; offset=nothing)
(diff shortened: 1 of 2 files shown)


simonster:
Thanks!

PR #98, opened 2015-02-26
Remove unnecessary check, r.wts cannot have length zero

lendle:


andreasnoack:
No longer relevant with #229

PR #100, opened 2015-03-06
Fix #99 by improving precision in ProbInv and variance calculations.

### LINKED ISSUE #99: Probit regression fails to converge when it should
From [this discussion](https://groups.google.com/forum/#!topic/julia-stats/8ngcUHpcDxQ).

``` julia
julia> using DataFrames, GLM

julia> srand(10001)

julia> N=10000
10000

julia> distX = Normal(0,3)
Normal(μ=0.0, σ=3.0)

julia> X = rand(distX,N);

julia> ɛ = randn(N);

julia> Y = X.>ɛ;

julia> df = DataFrame(X=[X],Y=[Y]);

julia> fittedmodel = glm(Y ~ X, df, Binomial(), ProbitLink(), verbose=true);
1: 5082.253852706774, 0.25854214488487215
2: 4555.984395197572, 0.11551169008917996
3: NaN, NaN
<...snip...>
30: NaN, NaN
WARNING: failure to converge in 30 iterations

julia> fittedmodel.model.pp.beta0
2-element Array{Float64,1}:
 NaN
 NaN
```

Here's a plot of the data set:
plot
There's plenty of overlap near X=0, so the issue is not complete separation. Fitting converges without issue with the `LogitLink()`.

R converges with an estimate close to 1 for the coefficient of X, and a reasonable SE:

``` R

> n=10000
> X=rnorm(n,0,3)
> eps=rnorm(n)
> Y = X > eps
> fit = glm(Y~X, family=binomial(link="probit"))
Warning message:
glm.fit: fitted probabilities numerically 0 or 1 occurred
> summary(fit)

Call:
glm(formula = Y ~ X, family = binomial(link = "probit"))

Deviance Residuals:
    Min       1Q   Median       3Q      Max
-3.3259  -0.2189   0.0000   0.2104   3.3539

Coefficients:
            Estimate Std. Error z value Pr(>|z|)
(Intercept) 0.004144   0.021013   0.197    0.844
X           0.998480   0.020138  49.582   <2e-16 ***

---

(Dispersion parameter for binomial family taken to be 1)

    Null deviance: 13862.9  on 9999  degrees of freedom
Residual deviance:  4561.8  on 9998  degrees of freedom
AIC: 4565.8

Number of Fisher Scoring iterations: 7
```

andreasnoack:
The `ProbInv` function was not precise for large negative values so using another formula improved precision. In addition, the Bernoulli variance calculation `p*(1-p)` could be improved for all link functions for which it doesn't matter if we set `p=F^{-1}(x'β)` or `p=F^{-1}(-x'β)`. Finally, residual deviation calculations for `Binomial` is made more precise.

I had to comment out the prediction tests because they are running a glm with Bernoulli response on data in the [0,1] interval. Does that makes sense? I haven't seen it before.

@dmbates, @kleinschmidt

### DIFF
--- a/src/glmtools.jl
+++ b/src/glmtools.jl
@@ -41,7 +41,7 @@ evaluate{T<:FP}(::CauchInv,x::T) = convert(T,0.5) + atan(x)/pi
 type CauchME <: Functor{1} end
 evaluate{T<:FP}(::CauchME,x::T) = one(T)/(pi*(one(T) + abs2(x)))
 type CLgLgLink <: Functor{1} end
-evaluate{T<:FP}(::CLgLgLink,x::T) = log(-log(one(T) - x))
+evaluate{T<:FP}(::CLgLgLink,x::T) = log(-log1p(-x))
 type CLgLgInv <: Functor{1} end
 evaluate{T<:FP}(::CLgLgInv,x::T) = -expm1(-exp(x))
 type CLgLgME <: Functor{1} end
@@ -51,11 +51,11 @@ evaluate{T<:FP}(::InvME,x::T) = -one(T)/abs2(x)
 type LogitME <: Functor{1} end
 evaluate{T<:FP}(::LogitME,x::T) = (e = exp(-abs(x)); f = one(T) + e; e / (f * f))
 type ProbLink <: Functor{1} end
-evaluate{T<:FP}(::ProbLink,x::T) = sqrt2*erfinv(convert(T,2.0)*x-one(T))
+evaluate{T<:FP}(::ProbLink,x::T) = -sqrt2*erfcinv(2*x)
 type ProbInv <: Functor{1} end
-evaluate{T<:FP}(::ProbInv,x::T) = (one(T) + erf(x/sqrt2))/convert(T,2.0)
+evaluate{T<:FP}(::ProbInv,x::T) = erfc(-x/sqrt2)/2
 type ProbME <: Functor{1} end
-evaluate{T<:FP}(::ProbME,x::T) = exp(-x*x/convert(T,2.0))/sqrt2π
+evaluate{T<:FP}(::ProbME,x::T) = exp(-x*x*convert(T,0.5))/sqrt2π
 type SqrtME <: Functor{1} end
 evaluate{T<:FP}(::SqrtME,x::T) = 2*x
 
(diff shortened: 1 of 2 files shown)


src/glmtools.jl
```diff
@@ -107,8 +130,7 @@ mustart{T<:FP}(d::Distribution,y::Vector{T},wt::Vector{T}) = mustart!(d, similar
 
 type BinomialDevResid <: Functor{3} end
 function evaluate{T<:FP}(::BinomialDevResid,y::T,mu::T,wt::T)
-    omy = one(T)-y
-    2.wt*(xlogy(y,y/mu) + xlogy(omy,omy/(one(T)-mu)))
+    2.wt*ifelse(y == 1, -log(mu), -log1p(-mu))
```
simonster: A ternary may be faster than `ifelse` here. I'll benchmark (and this shouldn't hold up the merge).

simonster:
It's a Binomial GLM, not a Bernoulli GLM, right? So I'd think that `y` is the proportion of positive responses to total observations for the corresponding predictors (and unequal numbers of observations for the given predictors could be handled by adjusting the weights).

lendle:
I actually use (quasi)binomial GLM pretty regularly on responses in the [0, 1] interval. The negative binomial log likelihood turns out to be a valid loss function for the conditional mean of a random variable in [0, 1]. Using a binomial GLM in this way is a good way to ensure that predictions respect the same bounds.

How about 

```julia
2.wt(y == 1 ? -log(mu) :
     y == 0 ? -log1p(-mu): 
     -y * log(mu) - (1-y) * log1p(-mu))
```

for evaluate(::BinomialDevResid, y, mu, wt)?

andreasnoack:
I've enabled the tests again and extended the `BinomialDevResid` evaluation to cover `y` values in the unit interval. The function is a bit complicated to analyze for precision, but I think the version I've pushed is more precise than evaluating the ratios before the logs.

A fascinating thing about working with Julia is that you interact with people who use the same tools as you, but slightly differently, so you'll see new sides of models you thought your already knew. I've always thought that the "family" part of glms should match the distribution of the response, but now I understant that you can fit the proportions/probabilities with probit/logit link.

However, I still don't understand why the glm literature ended up with `Binomial` instead of `Bernoulli` for models with binary response.

nalimilan:
Fascinating indeed. We call the variant with a logit link "logistic regression". Wikipedia has elements about Bernouilli vs. binomial: http://en.wikipedia.org/wiki/Logistic_regression#In_terms_of_binomial_data

lendle:
When `abs(eta)` is  big, `mu` can be numerically 0 or 1, which will result in `NaN`s (or at least `Inf`s) in `devresid` when `mu != y `.  For the special case of logistic regression, computing deviance residuals based on `eta` instead of `mu` avoids this issue.  Here's some (hopefully correct) code for that:
```julia
if y == 1
    2.0*wt*log1pexp(-eta)
elseif y == 0
    2.0*wt*log1pexp(eta)
else
    2.0*wt*(y*(log(y) + log1pexp(-eta)) + (1 - y)*(log1p(-y) + log1pexp(eta)))
end
```

Is it worth it to include this? I suppose it could implemented by adding a couple of methods for `devresid!`,
```julia
devresid!{V<:FPVector}(r::GlmResp{V, Binomial, LogitLink}) =
    devresid!(Binomial(), LogitLink(), r.devresid, r.y, r.eta, r.wts)
```
and
```julia
#compute devresid for logistic regression using eta instead of mu
#which avoids problems when mu = logistic(eta) is exactly 0 or 1
function devresid!{T<:FP}(::Binomial,::LogitLink, dr::Vector{T},y::Vector{T}, eta::Vector{T},wt::Vector{T})
    ...
end
```
But that would make the code a bit more complicated.

Maybe there are similar tricks with other link functions or distributions, I'm not sure.

andreasnoack:
> When abs(eta) is big, mu can be numerically 0 or 1

Yes that is the reason for the correction in `var!`. The deviance residuals will be slightly more complicated to improve the precision on and I don't think they are as likely to cause trouble. When `y=0` or `y=1` the first two branches take care of avoiding `Inf`/`NaN`s and I tried to run the problematic data with `prevfloat(1.0)` and `1e-300` instead of `1` and `0` and there are no `0`, `NaN`, or `Inf`s in the deviance residuals.

We don't have `log1pexp` but we have precise versions of `logccdf` for some/all distributions so your approach could work, but I'd like to postpone that work. Fell free to file an issue such that we'll have it on record if `NaN`/`Inf`s show up in the future.

simonster:
Any reason not to merge?

andreasnoack:
No. I think this can be merged. However, I wanted to think a little more because the problem in #99 is not completely fixed by this, so I'll continue the discussion there.

PR #107, opened 2015-07-02
Import df_residual

matthieugomez:
`df_residual` becomes a method from StatsBase if the following pull request is merged: https://github.com/JuliaStats/StatsBase.jl/pull/120

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,6 +1,6 @@
 julia 0.3-
 Distributions 0.4.6-
 NumericFuns 0.2.1-
-StatsBase 0.3.9+
+StatsBase 0.7.1
 Reexport
 Compat 0.2.12-
--- a/src/GLM.jl
+++ b/src/GLM.jl
@@ -8,8 +8,7 @@ module GLM
     using Distributions: sqrt2, sqrt2π
 
     import Base: (\), cholfact, cor, scale, show, size
-    import StatsBase: coef, coeftable, confint, deviance, loglikelihood, nobs, stderr,
-                      vcov, residuals, predict, fit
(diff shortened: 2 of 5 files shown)


Evizero:
same thing for `model_response`

Evizero:
I know people got things to do and places to be, and that they do this in their spare time. 

But it is a bummer that I keep running into awesome Julia Packages like this all the time, where very simple PRs just like this one get no reaction whatsover and just sit there

andreasnoack:
@matthieugomez Could you update this pr with an new version requirement for `StatsBase`? I've tagged a version of `StatsBase` after I merged your pull request.

matthieugomez:
Thanks! It is failing but I'm not sure it's related to this pull request

Evizero:
Could you include `model_response` to the import statement of StatsBase while your at it. I think there was an oversight and this seems to be a good PR to include it.

Concerning the build: I don't think it is related. I am having this precompilation issues as well on 0.4 with the current package state. I think I saw a issue about that yesterday but I can't seem to find it. It said something about updating to the newest 0.4 dev version if I recall correctly, but I didn't get around to try that yet

andreasnoack:
Yes. Please also add `model_response`. I'll take a look at the error. It might be the binaries used for the testing that aren't up to date.

matthieugomez:
All right. Also replaced some `StatsBase.fit` by `fit` since the function is imported.

andreasnoack:
To avoid the procompilation error, I had to change the ordering in which the packages were loaded in the test file. Now lights are green so I'll merge. Thanks for the contribution and sorry about the delay.

matthieugomez:
Thanks @Evizero and @andreasnoack

PR #110, opened 2015-10-01
Simplify the code

simonster:
Simplify the code

 - Use dispatch instead of functors
 - Remove NumericFuns dependency
 - Perform updatemu! in a single step, rather than by traversing each individual array


andreasnoack:
:+1:

PR #112, opened 2015-10-03
Fit GLMs with sparse design matrices

simonster:


andreasnoack:
Great. I actually thought this was aleady possible.

PR #114, opened 2015-10-15
Restructuring of factorizations

simonster:
New functionality:
- Avoid performing an extra Cholesky factorization when fitting a GLM.
- Allow refitting LinearModels with the same factorization but a new `y`. (This was already possible for GeneralizedLinearModels.)
- Allow fitting LinearModels using sparse Cholesky.
- Allow fitting GeneralizedLinearModels using dense QR.
- Implement syntax like `fit(LinearModel{GLM.Chol}, ...)` and `fit(GeneralizedLinearModel{GLM.QR}, ...)` to fit models using alternative factorizations.

Changes to internals:
- Move `beta0`, `delbeta`, and `scratchbeta` (now `curbeta`) to GLM object. These were not really related to the factorizations themselves, and keeping them together with the factorizations required extra fields on each. Since the factorizations now just generate the betas and not the linear predictors, I renamed `LinPred` to `LinFact`.
- Separate `delbeta!` into `factorize!` and `\`/`solve!`. (`solve!` is kind of like `A_ldiv_B!` but may create a new array instead of returning the output, since this is necessary for QR and sparse Cholesky).
- Separate unweighted and weighted LinFact types for Cholesky, to avoid creating unnecessary temporaries for ordinary linear models.
- Require Julia 0.4.

cc @matthieugomez, who may be interested.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ os:
     - linux
     - osx
 julia:
-    - 0.3
     - 0.4
     - nightly
 notifications:
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,4 +1,4 @@
-julia 0.3
+julia 0.4
 Distributions 0.4.6-
 StatsBase 0.7.1
 Reexport
--- a/src/GLM.jl
+++ b/src/GLM.jl
@@ -14,16 +14,13 @@ module GLM
     export                              # types
         CauchitLink,
         CloglogLink,
-        DensePred,
-        DensePredQR,
-        DensePredChol,
         GeneralizedLinearModel,
         GlmResp,
         IdentityLink,
         InverseLink,
         LinearModel,
         Link,
-        LinPred,
+        LinFact,
         LinPredModel,
         LogitLink,
         LogLink,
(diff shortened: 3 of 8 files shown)


matthieugomez:
Thanks, it looks neat. I have a couple of questions
- For the QR factorization, I think you can write a `solve!` for qrfact that does not allocate a new solution, by using `A_ldiv_B!(qrfact, b)` that overwrites the first elements of `b`
- For the sparse cholesky factorization, it's costly to allocate `p.Xt*scr` at each iteration. I think you can call CHOLMOD to (i) compute the symbolic factorization for `X` once (ii) update based on a rectangular matrix `scale!(.scratch, p.sqrtwt, p.X)` (like [here](https://github.com/matthieugomez/LeastSquaresOptim.jl/blob/master/src/solver/sparse_cholesky.jl)).

andreasnoack:
@simonster What is the status here? Is it ready?

kleinschmidt:
I'm closing this for now since I think things have moved on too far.  If you disagree @simonster (or anyone else), feel free to re-open.

Issue #74, opened 2014-05-21
Functions to evaluate fit


johansigfrids:
I can't seem to find any functions for evaluating the fit for a linear model. Does GLM have ways to get things like R^2, adjusted R^2, sigma, AIC, BIC?

JonathanAnderson:
i've been looking for this too.

when i need r2 i use:

```
R²(lm) = 1-norm(lm.model.rr.y-predict(lm))/norm(lm.model.rr.y)
```

gragusa:
skype: gasugar

Giuseppe Ragusa

On Thu, Oct 29, 2015 at 2:53 PM, Jonathan Anderson <notifications@github.com

gragusa:
Sorry for the noise. I mistakenly replied to this....

Giuseppe Ragusa

nalimilan:
I've made two PRs implementing the necessary functions: https://github.com/JuliaStats/StatsBase.jl/pull/146 and https://github.com/JuliaStats/GLM.jl/pull/115. Please comment/review!

diegozea:
I was also looking for R² in GLM...

PR #115, opened 2015-11-11
Implement fit statistics functions

nalimilan:
This implements functions to be added to StatsBase (https://github.com/JuliaStats/StatsBase.jl/pull/146): `loglikelihood`, `nullloglikelihood`, `nulldeviance`, `df`. It also adds an `adjR²` function, and changes `nobs` to return the sum of weights where it makes sense. Tests for fitting models using weights are also introduced. Fixes https://github.com/JuliaStats/GLM.jl/issues/74.

Issues to tackle:
- [x] Computing the R² is only supported for `LinearModel` currently, as for GLMs we need to compute the deviance of the null model. (Could be added later.) UPDATE: Will add later in another PR, this isn't essential.
- [x] Is it OK that `nobs` returned the sum of weights when they are present? This is the only logical choice for binomial on aggregated data (see `gm11` in the tests). This is the same behavior as Stata, but R returns the number of rows instead. This only makes sense for frequency weights, and this will have to be improved when we support more weights types. (Anyway I think this whole idea of number of independent observations is hard to define correctly, and the BIC is a fragile measure because of that.) UPDATE: I've made this change since currently the code treats weights as frequency weights (contrary to R).
- [x] Should the log-likelihood of individual observations be computed directly using formulas, or is calling `logpdf` fine? I don't expect `loglikelihood` to be a bottleneck. UPDATE: Let's keep it that way, at least for now.
- [x] Should `adjR²` be defined in `StatsBase` together with `R²`? I actually think it should, but the small glitch is that there is no generalized adjusted R² which is exactly equal to the classical adjusted R² for linear regression. The closest variant, the adjusted McFadden R², is slightly different. This means a linear regression would give slightly different results when fit via `lm` than when fit via `glm`. Not a big deal, though. UPDATE: I fixed this by defining `adjR2` in StatsBase, but requiring a `variant` argument for GLMs. That way, there is no inconsistency depending on the way the model is fitted.
- [x] Add tests for weighted Gamma and Poisson.
- [x] Add tests for `nullloglikelihood`.
- [x] Export `scale` under a different name (`sigma` or `dispersion`?). UPDATE: I went with `dispersion`.

To use this, you need https://github.com/JuliaStats/StatsBase.jl/pull/146. Also, after fitting a model from a `DataFrame`, you need to call the functions on `obj.model` rather than on `obj` directly (or to add code to DataFrames.jl to delegate the functions to that object). I'll make a PR to fix that when we have agreed on the API.

(Tests will fail until https://github.com/JuliaStats/StatsBase.jl/pull/146 and https://github.com/JuliaStats/DataFrames.jl/pull/921 are merged and releases tagged.)

### DIFF
--- a/src/glmfit.jl
+++ b/src/glmfit.jl
@@ -258,9 +258,16 @@ fit{M<:AbstractGLM}(::Type{M},
 
 glm(X, y, args...; kwargs...) = fit(GeneralizedLinearModel, X, y, args...; kwargs...)
 
-## scale(m) -> estimate, s, of the scale parameter
-## scale(m,true) -> estimate, s^2, of the squared scale parameter
-function scale(m::AbstractGLM, sqr::Bool=false)
+"""
+    dispersion(m::AbstractGLM, sqr::Bool=false)
+
+    Estimated dispersion (or scale) parameter for a model's distribution,
+    generally written σ for linear models and ϕ for generalized linear models.
+    It is by definition equal to 1 for Binomial and Poisson families.
+
+    If `sqr` is `true`, the squared parameter is returned.
+"""
+function dispersion(m::AbstractGLM, sqr::Bool=false)
     wrkwts = m.rr.wrkwts
     wrkresid = m.rr.wrkresid
 
--- a/src/glmfit.jl
+++ b/src/glmfit.jl
@@ -114,7 +114,23 @@ function confint(obj::AbstractGLM, level::Real)
 end
 confint(obj::AbstractGLM) = confint(obj, 0.95)
 
-deviance(m::AbstractGLM)  = deviance(m.rr)
+deviance(m::AbstractGLM) = deviance(m.rr)
+
+function loglikelihood(m::AbstractGLM)
+    r = m.rr
+    wts = r.wts
+    y = r.y
+    mu = r.mu
+    ϕ = deviance(m)/sum(wts)
+    d = r.d
+    ll = zero(one(eltype(wts)) * one(loglik_obs(d, y[1], mu[1], wts[1], ϕ)))
+    @inbounds for i in eachindex(y, mu, wts)
+        ll += loglik_obs(d, y[i], mu[i], wts[i], ϕ)
+    end
+    ll
+end
+
+df(x::GeneralizedLinearModel) = dispersion_parameter(x.rr.d) ? length(coef(x)) + 1 : length(coef(x))
 
 function _fit!(m::AbstractGLM, verbose::Bool, maxIter::Integer, minStepFac::Real,
               convTol::Real, start)
--- a/src/linpred.jl
+++ b/src/linpred.jl
@@ -117,7 +117,7 @@ Base.cholfact!{T}(p::SparsePredChol{T}) = p.chol
 
 invchol(x::DensePred) = inv(cholfact!(x))
 invchol(x::SparsePredChol) = cholfact!(x) \ eye(size(x.X, 2))
-vcov(x::LinPredModel) = scale!(invchol(x.pp), scale(x,true))
+vcov(x::LinPredModel) = scale!(invchol(x.pp), dispersion(x, true))
 
 function cor(x::LinPredModel)
     Σ = vcov(x)
--- a/src/linpred.jl
+++ b/src/linpred.jl
@@ -115,12 +115,6 @@ end
 Base.cholfact{T}(p::SparsePredChol{T}) = copy(p.chol)
 Base.cholfact!{T}(p::SparsePredChol{T}) = p.chol
 
-coef(x::LinPred) = x.beta0
-coef(x::LinPredModel) = coef(x.pp)
-
-df_residual(x::LinPredModel) = df_residual(x.pp)
-df_residual(x::@compat(Union{DensePred,SparsePredChol})) = size(x.X, 1) - length(x.beta0)
-
 invchol(x::DensePred) = inv(cholfact!(x))
 invchol(x::SparsePredChol) = cholfact!(x) \ eye(size(x.X, 2))
 vcov(x::LinPredModel) = scale!(invchol(x.pp), scale(x,true))
@@ -182,5 +176,24 @@ model_response(obj::LinPredModel) = obj.rr.y
 fitted(m::LinPredModel) = m.rr.mu
 predict(mm::LinPredModel) = fitted(mm)
 formula(obj::LinPredModel) = ModelFrame(obj).formula
-nobs(obj::LinPredModel) = length(model_response(obj))
 residuals(obj::LinPredModel) = residuals(obj.rr)
+
+"""
+    nobs(obj::LinearModel)
+    nobs(obj::GLM)
+
+For linear and generalized linear models, returns the number of rows, or,
+when prior weights are specified, the sum of weights.
+"""
+function nobs(obj::LinPredModel)
+    if isempty(obj.rr.wts)
+        oftype(sum(one(eltype(obj.rr.wts))), length(obj.rr.y))
+    else
+        sum(obj.rr.wts)
+    end
+end
+
+coef(x::LinPred) = x.beta0
+coef(obj::LinPredModel) = coef(obj.pp)
+
+df_residual(obj::LinPredModel) = nobs(obj) - length(coef(obj))
--- a/src/linpred.jl
+++ b/src/linpred.jl
@@ -134,41 +134,6 @@ function show(io::IO, obj::LinPredModel)
     println(io, "$(typeof(obj)):\n\nCoefficients:\n", coeftable(obj))
 end
 
-## function show(io::IO, obj::GlmMod)
-##     cc = coef(obj)
-##     se = stderr(obj)
-##     zz = cc ./ se
-##     pp = 2.0 * ccdf(Normal(), abs(zz))
-##     @printf("\n%s\n\nCoefficients:\n", obj.fr.formula)
-##     @printf("         Term    Estimate  Std. Error     t value    Pr(>|t|)\n")
-##     N = length(cc)
-##     for i = 1:N
-##         @printf(" %12s%12.5f%12.5f%12.3f%12.3f %-3s\n",
-##                 obj.mm.model_colnames[i],
-##                 cc[i],
-##                 se[i],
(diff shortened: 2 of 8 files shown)


src/glmfit.jl
```diff
@@ -115,7 +115,23 @@ function confint(obj::AbstractGLM, level::Real)
 end
 confint(obj::AbstractGLM) = confint(obj, 0.95)
 
-deviance(m::AbstractGLM)  = deviance(m.rr)
+deviance(m::AbstractGLM) = deviance(m.rr)
+
+function loglikelihood(m::AbstractGLM)
```
andreasnoack: Could this be simpler? It seems a bit wasteful to loop through all the observations when the same information is already computed. The deviance is almost the loglikelihood, `deviance(y,mu) = 2loglikelihood(y,y) - 2loglikelihood(mu,y)`, so we could define `loglikelihood(f) = -deviance(f)/2` since it only deviates from the joint logpdf with an additive constant.

nalimilan: Actually, the deviance is also computed by going over all observations. But I agree simplifying the code would be good. The issue is, as I mention in the second point of https://github.com/JuliaStats/StatsBase.jl/pull/146, I've not been able to compute the log-likelihood from the deviance as I expected and using the formula you mention. Any idea why?

andreasnoack: > Actually, the deviance is also computed by going over all observations.

Yes, but the computation of the deviance is just summing the contributions. Furthermore, since this version is also computing the deviance the observations are traversed twice.

> The issue is, as I mention in the second point of JuliaStats/StatsBase.jl#146

Sorry, I missed that. I'm able to match the exact values for `Binomial(n=1)` (which has loglik(y,y)=0) and `Poisson`, but haven't tried others. However, in general, it doesn't seem cheaper to compute `loglik(y,y)` than `loglik(mu,y)` because the formulas don't seem to simplify. That is why I suggested computing the log-likelihood as `-deviance/2`. It might confuse some that it's different from the joint logpdf, but it's sound, much cheaper, and can easily be documented.

dmbates: The squared deviance residuals returned by the `devresid` method are defined as `deviance - null deviance` for each observation.  The calculation of the deviance itself should remain as it is because it is the deviance that you need to evaluate most frequently. The current methods are fast and only involve a single pass over the observations. I don't think there is a problem with evaluating the loglikelihood in a separate loop because that calculation is needed much less frequently.

nalimilan: @dmbates Makes sense. For my education, do you know how to compute the log-likelihood from the deviance and the null deviance in general? To my surprise, I could find no explicit reference about this.

@andreasnoack The point is, the log-likelihood for GLMs is quite a well-defined measure, it would be terribly confusing to return something different from other software. But as Douglas said, it's not needed that often (e.g. the R² is currently computed from the deviance).

dmbates: @nalimilan I long ago gave up trying to make sense of null deviance, etc.

nalimilan: @dmbates That's kind of annoying that nobody knows how to compute that stuff. I'm not happy with the fact that in 
https://github.com/JuliaStats/StatsBase.jl/pull/146 I documented `deviance` as being equal up to a constant to `-2 log L` (like other programs do), and `AIC` as being `-2 log L + 2k`, and yet ``-2 log L` isn't the same value in those two docs: in the latter, it's really `-2loglikelihood(m)`, but in the former it's some unknown quantity to me. I must be missing something obvious.

dmbates: @nalimilan The documentation for deviance should be rephrased although I'm not sure to what.  I was more-or-less following the R conventions when I wrote the code but following my incorrect understanding of deviance when I wrote the documentation for deviance.

nalimilan: @dmbates Funny. :-) The problem is, the idea that "deviance = -2 log L" can be found everywhere on the net, so that's not just an issue with us. Do you have any idea about where I could look to find out the formula? I'd be really interested in understanding and documenting this.

jrklasen: @nalimilan I stumbled over this  thread and remember an old R-help conversation, maybe you find this helpful:
http://r.789695.n4.nabble.com/GLM-output-for-deviance-and-loglikelihood-tt3451911.html#none

nalimilan: Thanks, I also bumped into this thread. This is interesting, but doesn't say how to relate the deviance to the log-likelihood in cases where the simple formula doesn't apply.

simonster: The log likelihood may have terms that depend only on the samples or on the dispersion parameter. These drop out when computing the deviance since, as @andreasnoack says, it is defined as -2*the difference between the log likelihoods for the fitted model and the saturated model, but to compute the full log likelihood you would need to include them. I think the only way to do this is to compare the formulas for the log PDF and the deviance residual for the specific distribution and adjust the deviance as necessary. Also, when the distribution has a dispersion parameter, the log likelihood may depend upon how that dispersion parameter is estimated (method of moments vs. ML).

nalimilan: Yeah, but even looking closely at the formulas found in various sources, I still can't figure it out. Not a big issue in practice, but still I'm curious... I'll try harder. Something that bugs me in particular is how to take into account the dispersion parameter, which isn't the same for the fitted model and for the saturated model. And yet the formulas include both under the same `σ²` term...

nalimilan:
Now that the preliminary work has been merged in StatsBase, could you address the questions remaining in the points I highlighted in the description above? I'd like to get your input in particular about the second and last bullets.

nalimilan:
Please have a look at the updated commits. I think I adressed all points listed in the description (see UPDATE mentions). Though the tests are still going to fail until a new DataFrames release is tagged.

In particular, I've retained the interpretation that prior weights are [frequency weights](https://github.com/JuliaStats/StatsBase.jl/issues/53#issuecomment-39653253), which implies that the only consistent solution is to make `nobs` return the sum of weights when present. This has currently the drawback that `nobs` returns a float since we do not enforce that frequency weights are integer (which other software does). Anyway I'd like to add different types of weights: we could add this restriction on this occasion, and only support non-integer weights for other types of weights (analytical and sampling).

@dmbates Is that right?

src/linpred.jl
```diff
@@ -182,5 +141,24 @@ model_response(obj::LinPredModel) = obj.rr.y
...
+"""
+    nobs(obj::LinearModel)
+    nobs(obj::GLM)
+
+For linear and generalized linear models, returns the number of rows, or,
+when prior weights are specified, the sum of weights.
+"""
+function nobs(obj::LinPredModel)
+    if isempty(obj.rr.wts)
+        oftype(sum(one(eltype(obj.rr.wts))), length(obj.rr.y))
```
nalimilan: What's the best way of writing this?

andreasnoack: I think I'd go for `convert(eltype(obj.rr.wts), length(obj.rr.y))`. It's not true if the weights are Booleans but that is probably not a case we need to consider. It is also a similar issue for your version because scalar `sum` is matching the type of vector `sum`
```jl
julia> sum(true)
true

julia> sum([true])
1
```

nalimilan: I'm not too concerned about booleans, but more about `Int16` and `Int32`. Anyway, weights should be reworked at some point (https://github.com/JuliaStats/GLM.jl/issues/124), and when no weights are present a `Ones` type returning series of `1::Int` would allow use to simplify all of this.

nalimilan:
@dmbates Could you double-check my interpretation of weights as frequency weights?

dmbates:
@nalimilan As far as I know the primary use of weights as frequency weights is for the `Binomial()` distribution with data that have been condensed from multiple `Bernoulli()` responses.  In that case the prior weights are set to the number of trials for each response, thus they would be frequency weights.  But other uses of weights don't need to have an interpretation as frequency weights.

It might be best to check for the `Binomial()` distribution before deciding that the weights must be frequency weights.

nalimilan:
> It might be best to check for the Binomial() distribution before deciding that the weights must be frequency weights.

Thanks for confirming. I have indeed checked that a weighted binomial model gives correct results, i.e. same ones as when passing individual 0/1 observations (see the tests). Results also seem consistent with Stata in the cases that can be replicated there.

So if everything is fine, we should be able to merge this after a new data frames release has been tagged.

nalimilan:
We can merge as soon as you're OK, since a new DataFrames has been released and tests now pass.

andreasnoack:
@dmbates Do you have further comments here or should we merge?

nalimilan:
Something about which I'm still unsure (but which can be done at any time) is whether we should export `dispersion`. `scale` (which it replaces) was implicitly exported since it's a Base function. Exporting `dispersion` sounds natural, but I wonder whether it should also go into `StatsBase`, as a method for `StatisticalModel`. Are there other model families where the dispersion is estimated?

nalimilan:
Merging this after approval from @dmbates. I've just rebased on it master and added the necessary functions and tests for `Bernoulli`.

Note that code paths for weighted linear model are not tested, as this kind of model cannot be fitted currently AFAIK. We'll have to test this thoroughly when we enable it (and clearly document the meaning of the weights).

hpoit:
Can a demonstration be made of the R2 estimate and the accuracy of R2?

nalimilan:
What do you mean?

hpoit:
Like demonstrations with examples on https://github.com/JuliaStats/GLM.jl/blob/master/README.md?

hpoit:
Can I help create the methods r2 and accuracy?

nalimilan:
Yeah, probably. The examples also would need a bit a cleanup, since OLS should be computed using `lm` or `fit(LinearModel, ...`. Also, the R examples could be removed. A PR would be welcome of course.

hpoit:
Thanks Milan. What is a PR? Can I start with the creation of r2() and accuracy_r2() on linpred.jl first? Then clean up the examples, change glm() for lm(), remove R examples?

nalimilan:
PR stands for Pull Request (that's how you can get your commits included into a GitHub project).

But I still don't understand what you want to do with `r2` and `accuracy_r2`. We already have an `R²`/`R2` function (as of the brand new 0.5.2 version). And what would `accuracy_r2` do?

PR #122, opened 2016-01-27
add inline documentation and few tests

abhijithch:
We have created a test for Identity and square root type link, which were not covered previously.

### DIFF
--- a/src/glmtools.jl
+++ b/src/glmtools.jl
@@ -11,41 +11,122 @@ type LogLink <: Link end
 type ProbitLink <: Link end
 type SqrtLink <: Link end
 
+"""
+Link function mapping mu to eta, the linear predictor (CauchitLink response model).
+"""
 linkfun(::CauchitLink, μ) = tan(pi*(μ - oftype(μ, 0.5)))
+"""
+Inverse Link function mapping mu to eta, the linear predictor (CauchitLink response model).
+"""
 linkinv(::CauchitLink, η) = oftype(η, 0.5) + atan(η)/pi
 mueta(::CauchitLink, η) = one(η)/(pi*(one(η) + abs2(η)))
 
+"""
+Link function mapping mu to eta, the linear predictor (CloglogLink response model).
+"""
 linkfun(::CloglogLink, μ) = log(-log1p(-μ))
+"""
+Inverse Link function mapping mu to eta, the linear predictor (CloglogLink response model).
+"""
 linkinv(::CloglogLink, η) = -expm1(-exp(η))
+"""
+Derivative of inverse link.
+"""
 mueta(::CloglogLink, η) = exp(η)*exp(-exp(η))
 
+"""
+Link function mapping mu to eta, the linear predictor (IdentityLink response model).
+"""
 linkfun(::IdentityLink, μ) = μ
+"""
+Inverse Link function mapping mu to eta, the linear predictor (IdentityLink response model).
+"""
 linkinv(::IdentityLink, η) = η
+"""
+Derivative of inverse link.
+"""
 mueta(::IdentityLink, η) = 1
 
+"""
+Link function mapping mu to eta, the linear predictor (InverseLink response model).
+"""
 linkfun(::InverseLink, μ) = 1/μ
+"""
+Inverse Link function mapping mu to eta, the linear predictor (InverseLink response model).
+"""
 linkinv(::InverseLink, η) = 1/η
+"""
+Derivative of inverse link.
+"""
 mueta(::InverseLink, η) = -inv(abs2(η))
 
+"""
+Link function mapping mu to eta, the linear predictor (LogitLink response model).
+"""
 linkfun(::LogitLink, μ) = logit(μ)
+"""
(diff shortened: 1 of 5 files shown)


src/glmtools.jl
```diff
@@ -11,41 +11,122 @@ type LogLink <: Link end
 type ProbitLink <: Link end
 type SqrtLink <: Link end
 
+"""
+Link function mapping mu to eta, the linear predictor (CauchitLink response model).
```
nalimilan: Thanks. But instead of repeating the description of the function for each method, you'd better document the function itself only once:
```julia
"""
Link function mapping μ to η (the linear predictor).
"""
function linkfun end
```

Then I don't think you need to document the individual methods, as their signature is explicit enough.

Also, using the greek letters sounds better to me than spelling them in full.

src/linpred.jl
```diff
@@ -113,14 +119,24 @@ end
 Base.cholfact{T}(p::SparsePredChol{T}) = copy(p.chol)
 Base.cholfact!{T}(p::SparsePredChol{T}) = p.chol
 
+"""
+Extract the estimates of the coefficients in the model
```
nalimilan: Missing ending period here and below.

src/linpred.jl
```diff
@@ -113,14 +119,24 @@ end
...
+"""
 coef(x::LinPred) = x.beta0
 coef(x::LinPredModel) = coef(x.pp)
 
+"""
+Degrees of freedom for residuals, when meaningful
```
nalimilan: No, this returns the *number of residual degrees of freedom* (I agree the name is confusing).

src/lm.jl
```diff
@@ -19,6 +19,9 @@ function updatemu!{V<:FPVector}(r::LmResp{V}, linPr::V)
 end
 updatemu!{V<:FPVector}(r::LmResp{V}, linPr) = updatemu!(r, convert(V,vec(linPr)))
 
+"""
+Measure of the model fit, weighted residual sum of squares for lm's
```
nalimilan: This definition is too general for the first part. Anyway, I think you'd better leave this out as you're documenting the specific method for `LmResp`, which is identical to the only for `LinearModel`. I've opened a PR with detailed docs for the function in StatsBase (I'll try to get this merged soon): https://github.com/JuliaStats/StatsBase.jl/pull/146/

src/linpred.jl
```diff
@@ -181,6 +200,13 @@ model_response(obj::LinPredModel) = obj.rr.y
 
 fitted(m::LinPredModel) = m.rr.mu
 predict(mm::LinPredModel) = fitted(mm)
+
+"""
+Extract the formula from a model.
+"""
 formula(obj::LinPredModel) = ModelFrame(obj).formula
+"""
```
nalimilan: I don't think you need to repeat this here either, as the definition from StatsBase is enough.

test/runtests.jl
```diff
@@ -17,6 +17,24 @@ test_show(lm1)
 @test_approx_eq vcov(lm1) Σ
 @test_approx_eq cor(lm1.model) diagm(diag(Σ))^(-1/2)*Σ*diagm(diag(Σ))^(-1/2)
 
+#1).
```
nalimilan: This line isn't very useful, it's even confusing since the object is called `gm`, not `gm1`.

test/runtests.jl
```diff
@@ -17,6 +17,24 @@ test_show(lm1)
...
+
+#2).
+data1=DataFrame(X=[1,2,3],Y=[1,0,1])
+
+gm01 = fit(GeneralizedLinearModel, Y~X,data1,Binomial(),SqrtLink())
+
+@test_approx_eq coef(gm01) [0.8164965809876493,-3.7740636138506924e-17]
+@test_approx_eq stderr(gm01) [0.44094786686024684,0.20411919904552372]
+
+######
```
nalimilan: No need for this.

nalimilan:
Thanks, docs are very useful. About the tests, why do you say the identity link wasn't covered? The `gm6` model seems to test this, doesn't it? Anyway, could you split the tests addition into a separate commit for clarity?

src/glmtools.jl
```diff
@@ -11,41 +11,122 @@ type LogLink <: Link end
...
+"""
 linkfun(::CauchitLink, μ) = tan(pi*(μ - oftype(μ, 0.5)))
+"""
```
nalimilan: Ah, and I think it would make sense to add a line break before `"""` now that each definition + docs takes several lines.

nalimilan:
Forgot to say, the current convention is to put the method(s) signature(s) at the top of the doc, with four spaces before. For example:
```julia
"""
    wrkresp(r::GlmResp)

Working response.
"""
 function wrkresp(r::GlmResp)
```

I hope this will be done automatically at some point.

abhijithch:
Hey thanks for the detailed response, will send a fresh PR.

nalimilan:
Note that you don't need to make a new PR. You can just do `git push abhijithch +master`, the `+` will allow erasing the previous commits.

PR #123, opened 2016-02-08
Use eachindex(x) instead of 1:length(x) and other cleanup

dmbates:
This PR is rather simple - primarily replacing instances of `1:length(x)` by `eachindex(x)` which, IIUC, should be better with subarrays, if we choose to use them.  I also removed the conditional code for versions less than v"0.4" because "julia 0.4" is now part of the `REQUIRE` file.

The next thing I plan to look at is flagging the canonical link function for a distribution to take advantage in the `wrkwts!` method of the fact that r.mueta == r.var for the canonical link.

### DIFF
--- a/src/glmfit.jl
+++ b/src/glmfit.jl
@@ -44,13 +44,13 @@ function updatemu!{T<:FPVector}(r::GlmResp{T}, linPr::T)
     wrkresid = r.wrkresid
     devresidv = r.devresid
 
-    if length(offset) == length(eta)
-        broadcast!(+, eta, linPr, offset)
-    else
+    if isempty(offset)
         copy!(eta, linPr)
+    else
+        broadcast!(+, eta, linPr, offset)
     end
 
-    @inbounds @simd for i = 1:length(eta)
+    @inbounds @simd for i = eachindex(eta)
         η = eta[i]
 
         # apply the inverse link function generating the mean vector (μ) from the linear predictor (η)
@@ -68,25 +68,21 @@ function updatemu!{T<:FPVector}(r::GlmResp{T}, linPr::T)
 end
 
 function wrkresp(r::GlmResp)
-    if length(r.offset) > 0
-        tmp = r.eta - r.offset
-        broadcast!(+, tmp, tmp, r.wrkresid)
-    else
-        r.eta + r.wrkresid
-    end
+    tmp = r.eta + r.wrkresid
+    isempty(r.offset) ? tmp : broadcast!(-, tmp, tmp, r.offset)
 end
 
 function wrkwt!(r::GlmResp)
     wrkwts = r.wrkwts
     mueta = r.mueta
     var = r.var
-    if length(r.wts) == 0
-        @simd for i = 1:length(r.var)
+    if isempty(r.wts)
+        @simd for i = eachindex(r.var)
             @inbounds wrkwts[i] = abs2(mueta[i])/var[i]
         end
     else
         wts = r.wts
-        @simd for i = 1:length(r.var)
+        @simd for i = eachindex(r.var)
             @inbounds wrkwts[i] = wts[i] * abs2(mueta[i])/var[i]
         end
     end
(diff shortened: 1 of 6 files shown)


src/glmtools.jl
```diff
@@ -1,4 +1,4 @@
-using Base.Cartesian
+#using Base.Cartesian
```
nalimilan: Why not remove this line altogether?

src/glmfit.jl
```diff
@@ -68,25 +68,21 @@ function updatemu!{T<:FPVector}(r::GlmResp{T}, linPr::T)
...
 end
 
 function wrkwt!(r::GlmResp)
     wrkwts = r.wrkwts
     mueta = r.mueta
     var = r.var
-    if length(r.wts) == 0
-        @simd for i = 1:length(r.var)
+    if isempty(r.wts)
+        @simd for i = eachindex(r.var)
```
nalimilan: I think it would be even more correct to write `eachindex(r.var, mueta, var)` (unless we are absolutely sure these arrays will always have the same type and structure). Same below.

src/linpred.jl
```diff
@@ -55,22 +45,22 @@ DensePredChol{T<:BlasReal}(X::Matrix{T}) =
     DensePredChol(X, zeros(T, size(X, 2)), zeros(T, size(X, 2)), zeros(T, size(X, 2)), cholfact!(X'X), similar(X))
 cholpred(X::Matrix) = DensePredChol(X)
 
-if VERSION >= v"0.4.0-dev+4356"
+# if VERSION >= v"0.4.0-dev+4356"   # GLM requires julia 0.4
```
nalimilan: Why not remove this too?

nalimilan:
+1 in general

simonster:
AFAIK SubVectors always have efficient linear indexing, so this shouldn't change anything performance-wise, but this is now the preferred way of writing these loops, +1 from me.

src/glmtools.jl
dmbates: I will do so.  I just wanted to check that commenting it out didn't cause a failure in the tests.

dmbates:
I don't know how to merge this now because @andreasnoack  modified master.  I tried to follow the instructions about "use the command line" and now everything is messed up in my repository.

Should I just forget the whole thing?

nalimilan:
If you're absolutely sure you don't have any local changes, you can remove the local branch by doing `git checkout master && git branch -D yourlocalbranch`. Then you can checkout the branch again from GitHub. Then `git rebase master` should work (but I agree it's easy to get into a total mess...).

If you like, I can try rebasing this myself.

dmbates:
@nalimilan I'm never sure about anything with git.  If you are willing to try to rebase please do - otherwise we should just throw it all out and I'll start from scratch.  I had put forward the PR in the hope that after discussing it someone would merge it.  I should have done so myself.

KristofferC:
You can look a git reflog and git reset to a good commit if you want.

andreasnoack:
I've rebased it now. Sorry for the mess. Is it okay if I push to your branch?

andreasnoack:
@dmbates I've pushed the rebase to this branch here. Could you take a look and see if I messed up something?

dmbates:
@andreasnoack The version now does not contain many of the cleanup changes that you made in master, such as adding spaces around infix operators.  Is it supposed to have those changes?

There is still the message that "This branch has conflicts that must be resolved", which is where trouble always starts for me. :8ball:

andreasnoack:
@dmbates Sorry. For some reason, I didn't paste the link to the branch https://github.com/JuliaStats/GLM.jl/tree/anj/eachindex. I decided to push the rebased version there such that I didn't destroy this version before it is ready.

dmbates:
@andreasnoack Looks good to me.  Please go ahead and merge your branch then we can delete my branch and this pull request.

Sorry to be so inept at git and cause all this hassle.

andreasnoack:
Done! The rebase was actually a bit tricky because you had more than one commit that touched the same lines as my changes. Please just ask if you have a conflict that requires a rebase. I (or some of the other here) is doing it so often that it is a routine so it is not worth your time and frustrations.

PR #126, opened 2016-02-10
Some formatting changes.

andreasnoack:
- Only a single expression per line
- Spaces around operators
- Line breaks in long lines
- Remove whitespace


dmbates:
LGTM

We do have a couple of vestigial `fill!(similar(y), zero(eltype(y))` calls that could be written more concisely as `zeros(y)`

andreasnoack:
I've changed the verbose `fill!`s as well.

PR #128, opened 2016-02-18
Use StatsBase.WeightVec for weights member.

dmbates:
As I mentioned in a julia-stats posting, I don't know how to template the WeightVec properly in the type declaration.  Suggestions would be welcome.

### DIFF
--- a/src/glmfit.jl
+++ b/src/glmfit.jl
@@ -8,8 +8,8 @@ type GlmResp{V<:FPVector,D<:UnivariateDistribution,L<:Link} <: ModResp       # r
     mueta::V                                   # derivative of mu w.r.t. eta
     offset::V                                  # offset added to linear predictor (usually 0)
     var::V                                     # (unweighted) variance at current mu
-    wts::V                                     # prior weights
-    wrkwts::V                                  # working weights
+    wts::StatsBase.WeightVec                   # prior weights
+    wrkwts::StatsBase.WeightVec                # working weights
     wrkresid::V                                # working residuals
     function GlmResp(y::V, d::D, l::L, eta::V, mu::V, off::V, wts::V)
         if isa(d, Binomial)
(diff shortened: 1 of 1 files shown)


nalimilan:
I don't think it's currently possible to make type parameters to depend on one another in the same signature. But that's not a serious problem either: you can simply add an additional parameter for that, and set it to `eltype(V)` in the `GlmResp` constructor:
```julia
type GlmResp{V<:FPVector,W<:Real,D<:UnivariateDistribution,L<:Link}
...
    wts::WeightVec{W,V}                   # prior weights
    wrkwts::WeightVec{W,V}                # working weights
...
end
```

Then it looks like you need an outer constructor to extract `W` and parameterize `GlmResp` on it. Not sure it it's possible only with an inner constructor (in my tests, it doesn't work).

But what's the interest of using `WeightVec` here? AFAIK two main advantages of this type are 1) avoid computing the sum over and over, and 2) support multiple dispatch based on the type. It doesn't look like these really apply here (we could store the sum in another field if needed), and from the diff it looks like it makes the code more verbose, not less.

dmbates:
@andreasnoack This is the specific issue I was considering in my post to julia-stats about template parameters for members of a composite type.  In some ways the easiest approach may be to change the WeightedVec template to the type of vector only, not eltype and vectortype.

@nalimilan I should have made it clearer that I was intending to extend the WeightVec type with checks on whether the weights were non-negative and, perhaps, creating a supertype with another subtype of UJnitWeightVec.  That way one could use multiple dispatch on unit weights.

At present I am extending the MixedModels package to allow for fitting generalized linear mixed models.  I plan to use the GlmResp type for the response to avoid code duplication.  A common operation in GLMMs is to change the weights then form Z'WZ where Z is an indicator matrix.  It can be expressed as a weighted count of the indicator categories.

By itself this isn't a big deal but it is an advantage to localize common operations.  Right now I am looking at whether threads can be used to enhance the performance of glm and glmm.  It is easier to isolate a particular piece of code for `@threads`  when it only occurs in one place.

nalimilan:
If you want to allow for different subtypes of, say, `AbstractWeightVector`, then you'll have to parameterize `GlmResp` on the type of the weights vector(s). Else, there would be type instability in the fields.

This would also fix the issue with type parameters you have with the current version.

PR #130, opened 2016-02-29
Allow users to use integer arrays as input (but convert them).

### LINKED ISSUE #84: lm(X,y) supports y as integer, but not X
Currently, lm(X,y) has asymmetric type support - y can be integer, X cannot. Please permit X as integer type. Example of the issue:

julia> y
3-element Array{Int64,1}:
 2
 4
 6

julia> X
3x1 Array{Int64,2}:
 1
 2
 3

julia> lm(X,y)
ERROR: `DensePredQR{T<:Union(Float64,Float32)}` has no method matching DensePredQR{T<:Union(Float64,Float32)}(::Array{Int64,2})

julia> X
3x1 Array{Float64,2}:
 1.0
 2.0
 3.0

julia> lm(X,y)
LinearModel{DensePredQR{Float64}}:

Coefficients:     Estimate   Std.Error    t value Pr(>|t|)
x1        2.0 3.84593e-16 5.20031e15   <1e-31

pkofod:
Fixes #84 .

### DIFF
--- a/src/lm.jl
+++ b/src/lm.jl
@@ -66,15 +66,15 @@ cholfact(x::LinearModel) = cholfact(x.pp)
 function fit{LmRespT<:LmResp,LinPredT<:LinPred, T<:FP}(::Type{LinearModel{LmRespT,LinPredT}},
     X::@compat(Union{Matrix{T},SparseMatrixCSC{T}}), y::FPVector)
     rr = LmRespT(y)
-    pp = LinPredT(full(X))
+    pp = LinPredT(X)
     installbeta!(delbeta!(pp, rr.y))
     updatemu!(rr, linpred(pp, 0.0))
     LinearModel(rr, pp)
 end
 function fit(::Type{LinearModel}, X::@compat(Union{Matrix,SparseMatrixCSC}), y::Vector)
-    y = float(y)
-    T = eltype(y)
-    return fit(LinearModel{LmResp{typeof(y)}, DensePredQR{T}}, float(X), y)
+    yy = float(y)
+    T = eltype(yy)
+    return fit(LinearModel{LmResp{typeof(yy)}, DensePredQR{T}}, float(X), yy)
 end
 
 lm(X, y) = fit(LinearModel, X, y)
--- a/src/lm.jl
+++ b/src/lm.jl
@@ -63,18 +63,18 @@ end
 
 cholfact(x::LinearModel) = cholfact(x.pp)
 
-function fit{LmRespT<:LmResp,LinPredT<:LinPred}(::Type{LinearModel{LmRespT,LinPredT}},
-    X::@compat(Union{Matrix,SparseMatrixCSC{FP}}), y::FPVector)
+function fit{LmRespT<:LmResp,LinPredT<:LinPred, T<:FP}(::Type{LinearModel{LmRespT,LinPredT}},
+    X::@compat(Union{Matrix{T},SparseMatrixCSC{T}}), y::FPVector)
     rr = LmRespT(y)
-    pp = LinPredT(X)
+    pp = LinPredT(full(X))
     installbeta!(delbeta!(pp, rr.y))
     updatemu!(rr, linpred(pp, 0.0))
     LinearModel(rr, pp)
 end
-function fit(::Type{LinearModel}, X::Matrix, y::Vector)
-    T = promote_type(eltype(X), eltype(y), Float64)
-    yy = convert(Vector{T}, y)
-    return fit(LinearModel{LmResp{typeof(yy)}, DensePredQR{T}}, convert(Matrix{T}, X), convert(Vector{T}, y))
+function fit(::Type{LinearModel}, X::@compat(Union{Matrix,SparseMatrixCSC}), y::Vector)
+    y = float(y)
+    T = eltype(y)
+    return fit(LinearModel{LmResp{typeof(y)}, DensePredQR{T}}, float(X), y)
 end
 
 lm(X, y) = fit(LinearModel, X, y)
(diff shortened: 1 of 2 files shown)


src/lm.jl
```diff
@@ -64,7 +64,7 @@ end
 cholfact(x::LinearModel) = cholfact(x.pp)
 
 function fit{LmRespT<:LmResp,LinPredT<:LinPred}(::Type{LinearModel{LmRespT,LinPredT}},
-    X::AbstractMatrix, y::Vector)
+    X::@compat(Union{Matrix,SparseMatrixCSC{FP}}), y::FPVector)
```
nalimilan: `SparseMatrixCSC{FP}` will only accept matrices with the abstract type `FP`, and not of any concrete subtypes. Use `T<:FP` as for GLMs.

BTW, it would be great to test sparse matrices (which would have caught this problem).

src/lm.jl
```diff
@@ -72,9 +72,9 @@ function fit{LmRespT<:LmResp,LinPredT<:LinPred}(::Type{LinearModel{LmRespT,LinPr
     LinearModel(rr, pp)
 end
 function fit(::Type{LinearModel}, X::Matrix, y::Vector)
-    T = promote_type(eltype(X), eltype(y), Float32)
+    T = promote_type(eltype(X), eltype(y), Float64)
```
nalimilan: I'm not sure what the original code is supposed to do exactly, but `yy` isn't used anywhere (except for its type, which is a waste). Why not simply use `float` as for GLMs? In general, I'd keep the two model types completely in sync unless there's a good reason not to.

nalimilan:
Thanks! I've made some comments.

pkofod:
Great comments, I will address them, and add tests for sparse matrices.

pkofod:
Is this better? Notice I'm doing `full(X)`, so it doesn't _really_ work with sparse input, but I such an addition belongs in a separate PR. I'm pretty sure several functions would have to be adapted, and the readme also states that only dense versions of the `LinPred` are supported.

nalimilan:
I don't like calling `full` automatically, as it may use too much memory without warning. It's no better than leaving the caller do the conversion manually. If it doesn't work, then let it fail for now. Does it work for GLMs?

src/lm.jl
```diff
@@ -63,18 +63,18 @@ end
...
     installbeta!(delbeta!(pp, rr.y))
     updatemu!(rr, linpred(pp, 0.0))
     LinearModel(rr, pp)
 end
-function fit(::Type{LinearModel}, X::Matrix, y::Vector)
-    T = promote_type(eltype(X), eltype(y), Float32)
-    yy = convert(Vector{T}, y)
-    return fit(LinearModel{LmResp{typeof(yy)}, DensePredQR{T}}, X, y)
+function fit(::Type{LinearModel}, X::@compat(Union{Matrix,SparseMatrixCSC}), y::Vector)
+    y = float(y)
```
nalimilan: For type stability, assign the result to `yy` (even if here the function is pretty short so it shouldn't make a big difference).

src/lm.jl
```diff
@@ -63,18 +63,18 @@ end
...
+    X::@compat(Union{Matrix{T},SparseMatrixCSC{T}}), y::FPVector)
     rr = LmRespT(y)
-    pp = LinPredT(X)
+    pp = LinPredT(full(X))
     installbeta!(delbeta!(pp, rr.y))
     updatemu!(rr, linpred(pp, 0.0))
     LinearModel(rr, pp)
 end
-function fit(::Type{LinearModel}, X::Matrix, y::Vector)
-    T = promote_type(eltype(X), eltype(y), Float32)
```
nalimilan: @andreasnoack You wrote these lines in https://github.com/pkofod/GLM.jl/commit/1ef71066b5c60e3419d632e489b00ca1c783a8a6. Why are they different from the corresponding code for GLMs? Should we keep the logic, even if it was broken (`yy` wasn't used)?

pkofod:
Alright, I'll let it fail instead. Yes it works for `glm`.

src/lm.jl
andreasnoack: I think `lm` and `glm` have been diverging slightly for a while. The change I made in that commit made the difference smaller, but I focused on a specific issue with abstract types as fields and not on getting `lm` and `glm` in sync. We should aim for getting them in sync and my guess is that `glm` has the better methods so it's probably fine to delete this method.

src/lm.jl
```diff
@@ -63,18 +63,18 @@ end
 
 cholfact(x::LinearModel) = cholfact(x.pp)
 
-function fit{LmRespT<:LmResp,LinPredT<:LinPred}(::Type{LinearModel{LmRespT,LinPredT}},
-    X::AbstractMatrix, y::Vector)
+function fit{LmRespT<:LmResp,LinPredT<:LinPred, T<:FP}(::Type{LinearModel{LmRespT,LinPredT}},
+    X::@compat(Union{Matrix{T},SparseMatrixCSC{T}}), y::FPVector)
```
andreasnoack: Why restrict the matrix type?

pkofod: I just followed the glm code.

andreasnoack: If `AbstractMatrix` doesn't cause any issues, I'd stick with that here and change `glm` instead (in a separate PR).

pkofod: Sure, I'll change it later.

pkofod:
I've "reverted" to AbstractMatrix. Notice the parametric type in the "first" `fit` method. I'm not using `T` for anything, so I'm unsure if it's non-Julian or not; it's basically just to enforce `eltype<:FP` as in `glm`.

andreasnoack:
Thanks

PR #135, opened 2016-04-26
Add Bernoulli distribution.

dmbates:
Many, if not most, cases of fitting a GLM with a `Binomial()` distribution could be written as `Bernoulli()`.  There is a lot of messiness dealing with `Binomial()`.  We should encourage the use of `Bernoulli()` instead.

The problem with `Binomial()` is that the distribution does not depend solely on `μ`, the probability of success.  You also need to know `n`, the number of trials, which is passed (incorrectly) as the case weights.  If `n` is always `1`, just call the distribution `Bernoulli()`.

I also deleted the unused `src/distributed.jl` file that I added a long time ago when I was messing around with distributed arrays.  Some trailing spaces were automatically removed by my editor.


src/glmtools.jl
```diff
@@ -45,26 +45,31 @@ linkfun(::SqrtLink, μ) = sqrt(μ)
...
     μ = linkinv(link, ifelse(η < 0, η, -η))
     μ * (1 - μ)
 end
-glmvar(::Binomial, ::Link, μ, η) = μ * (1 - μ)
+glmvar(::@compat(Union{Bernoulli,Binomial}), ::Link, μ, η) = μ * (1 - μ)
 glmvar(::Gamma, ::Link, μ, η) = abs2(μ)
 glmvar(::Normal, ::Link, μ, η) = 1
 glmvar(::Poisson, ::Link, μ, η) = μ
 
+mustart(::Bernoulli, y, wt) = 0.5y + oftype(y, 0.25)
```
nalimilan: Better write this as `y/2`. Cf. http://docs.julialang.org/en/latest/manual/style-guide/#avoid-using-floats-for-numeric-literals-in-generic-code-when-possible

src/glmtools.jl
```diff
@@ -45,26 +45,31 @@ linkfun(::SqrtLink, μ) = sqrt(μ)
...
 mustart(::Binomial, y, wt) = (wt * y + oftype(y, 0.5)) / (wt + one(y))
 mustart(::Gamma, y, wt) = y
 mustart(::Normal, y, wt) = y
 mustart(::Poisson, y, wt) = y + oftype(y, 0.1)
 
+function devresid(::Bernoulli, y, μ, wt)
+    -2wt * (y == 1 ? log(μ) : (y == 0 ? log1p(-μ) : throw(ArgumentError("$y ∉ support(Bernoulli)"))))
```
nalimilan: Why not write this as an `if... elseif... else`? That would be easier to read.

Also, the error message would probably be clearer as something like `"y should be 0 or 1 (got $y)"`.

nalimilan:
Interesting, I wonder why R doesn't provide it for simplicity.

dmbates:
@nalimilan The R `glm()` code was written by a summer student in about 1994 when Ross and Robert were the only people using R and the language was rather primitive.  The software engineering is, shall we say, "casual".  Until recently the code for the log-likelihood evaluation required an argument called `n` for every distribution even though the name was not bound in the calling environment unless the distribution family was `binomial()`.  You can get away with this in R because of lazy evaluation.   People keep threatening to rewrite the `glm` code but now there is too much other code that depends upon these quirks.

nalimilan:
OK, that's scary. :-) Do you think we should deprecate `Binomial` in favor of `Bernoulli` at some point? Or are there cases in which it is still needed?

dmbates:
`Binomial()` is still needed for condensed data sets where each response really is `n` trials and `y` successes.  To make it fit in with the exponential family you need to use the proportion of successes, `y/n` as the response and use `n` for the case weights.

The only alternative is to spread the responses out so that each row of `n, y` becomes `n` rows with `1`  's in `y` of the rows and `0`'s in the rest, but that can be expensive, although it may be worth testing exactly how expensive.

nalimilan:
OK, I see.

Looks good to me, though you could squash the two commits. I'll rebase https://github.com/JuliaStats/GLM.jl/pull/115 on top of that and merge it soon unless you have objections.

dmbates:
No objections.  Please do go ahead with the rebase and merge.

nalimilan:
OK, I've squashed you commits.

PR #143, opened 2016-06-13
Dmb/hermitian

dmbates:
@andreasnoack Could you take a look?  I'm not sure that the version at which the change in `cholfact!` was made.


andreasnoack:
Sure. I've adjusted the version, squashed a commit and pushed to this PR.

PR #145, opened 2016-06-16
Ensure non-zero starting μ for Gamma

dmbates:
A simple change to avoid starting estimates for μ being zero with a Gamma distribution.  Usually the `InverseLink` or the `LogLink` are used with a Gamma distribution.  If the starting estimate for an element of  μ is zero everything falls down.

### DIFF
--- a/src/glmfit.jl
+++ b/src/glmfit.jl
@@ -26,7 +26,7 @@ type GlmResp{V<:FPVector,D<:UnivariateDistribution,L<:Link} <: ModResp       # r
         lo = length(off)
         lo == 0 || lo == n || error("offset must have length $n or length 0")
         res = new(y,d,l,similar(y),eta,mu,similar(y),off,similar(y),wts,similar(y),similar(y))
-        updatemu!(res, eta)
+        updateμ!(res, eta)
         res
     end
 end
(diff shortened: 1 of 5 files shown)


src/glmfit.jl
```diff
@@ -26,72 +26,104 @@ type GlmResp{V<:FPVector,D<:UnivariateDistribution,L<:Link} <: ModResp       # r
...
+            wti = wt[i]
+            devr[i] *= wti
+            wkwt[i] *= wti
+        end
+    end
+end
+
+
+function updateμ!{T<:FPVector, D<:Union{Bernoulli, Binomial}, L<:LogitLink}(
+    m::GlmResp{T,D,L}, η::T)
```
andreasnoack: Maybe just use `LogitLink` instead of `L` here since `LogitLink` is concrete.

src/glmfit.jl
```diff
@@ -26,72 +26,104 @@ type GlmResp{V<:FPVector,D<:UnivariateDistribution,L<:Link} <: ModResp       # r
...
+    η, offset, wrkresid = m.η, m.offset, m.wrkresid
+    @inbounds @simd for i in eachindex(η)
+        v[i] = wrkresid[i] + η[i] - offset[i]
+    end
+    v
+end
+
+function updateμ!{T<:FPVector}(r::GlmResp{T}, linPr::T)
+    y, d, l, η, μ = r.y, r.d, r.l, r.eta, r.mu
+    @show d, l, typeof(μ), typeof(η)
```
simonster: Doesn't look like this should be here

andreasnoack:
Great that only requires so few changes to get this fast path for logit with canonical link. I think I've mentioned it before but a more extensive change that might be able to speed up the evaluations, even more, could be to use a gradient based convergence criterion. That would avoid the many logarithm computations during the iterations.

src/glmfit.jl
```diff
@@ -26,72 +26,104 @@ type GlmResp{V<:FPVector,D<:UnivariateDistribution,L<:Link} <: ModResp       # r
...
+    end
+end
+
+
+function updateμ!{T<:FPVector, D<:Union{Bernoulli, Binomial}, L<:LogitLink}(
+    m::GlmResp{T,D,L}, η::T)
+
+    y, μ, wrkres, wrkwt, dres = m.y, m.mu, m.wrkresid, m.wrkwts, m.devresid
+
+    @inbounds Threads.@threads for i in 1:length(μ)
```
nalimilan: Maybe add `using Base.Threads` at the top to avoid repeating `Threads.`? Really cool that we can use several cores!

src/glmfit.jl
```diff
@@ -137,37 +169,45 @@ function _fit!(m::AbstractGLM, verbose::Bool, maxIter::Integer, minStepFac::Real
     m.fit && return m
     maxIter >= 1 || error("maxIter must be positive")
     0 < minStepFac < 1 || error("minStepFac must be in (0, 1)")
-
+@show minStepFac
```
nalimilan: Should remove these.

src/glmtools.jl
```diff
@@ -63,7 +63,7 @@ glmvar(::Poisson, ::Link, μ, η) = μ
 
 mustart(::Bernoulli, y, wt) = (y + oftype(y, 0.5)) / 2
 mustart(::Binomial, y, wt) = (wt * y + oftype(y, 0.5)) / (wt + one(y))
-mustart(::Gamma, y, wt) = y
+mustart(::Gamma, y, wt) = y == 0 ? oftype(y, 0.1) : y  # add a test
```
nalimilan: Is the comment a FIXME?

src/glmfit.jl
```diff
@@ -26,72 +26,104 @@ type GlmResp{V<:FPVector,D<:UnivariateDistribution,L<:Link} <: ModResp       # r
...
+            devr[i] *= wti
+            wkwt[i] *= wti
+        end
+    end
+end
+
+
+function updateμ!{T<:FPVector, D<:Union{Bernoulli, Binomial}, L<:LogitLink}(
+    m::GlmResp{T,D,L}, η::T)
+
```
nalimilan: Remove blank line here (and below).

dmbates:
Replaced by #155

PR #150, opened 2016-09-12
Fix deprecation warnings for Binomial and vectorized functions

andreasnoack:


test/runtests.jl
```diff
@@ -209,7 +209,7 @@ test_show(gm17)
 srand(1)
 X = sprand(1000, 10, 0.01)
 β = randn(10)
-y = Bool[rand() < x for x in logistic(X * β)]
+y = Bool[rand() < x for x in @compat logistic.(X * β)]
```
nalimilan: More simply and more efficiently written as `rand() < logistic(x)`? (Not that it matters a lot...)

nalimilan:
For the record, note that the `Binomial` change means non-integer weights are no longer accepted.

test/runtests.jl
andreasnoack: Yes. That is better. I'll update.

PR #152, opened 2016-09-22
Require julia 0.5, remove dependency on Compat

dmbates:


src/glmfit.jl
```diff
@@ -250,7 +250,7 @@ function fit{M<:AbstractGLM,T<:FP,V<:FPVector}(::Type{M},
 end
 
 fit{M<:AbstractGLM}(::Type{M},
-    X::@compat(Union{Matrix,SparseMatrixCSC}),
```
ararslan: Hah, these `@compat`s are actually left over from 0.3/0.4 changes. They could have been removed a long time ago.

andreasnoack:
LGTM

ararslan:
Since this is bumping to Julia 0.5, would you mind changing the stuff that's deprecated in 0.5? In particular the implicit vectorization is causing a bunch of warnings.

There also seems to be a DataArrays issue on nightly... Not sure yet what that's about.

dmbates:
@ararslan Will do the changes on implicit vectorization.

dmbates:
I haven't been able to check this on the dev version of Julia as my compilation is failing.  I will download a pre-compiled nightly to test on.

dmbates:
This is okay on my local version of nightly except for the problem in `DataArrays`.  Considering the switch to `CategoricalArrays` and `NullableArrays` it may not be worthwhile tracking down a problem in `DataArrays`/

nalimilan:
Do we really need to require 0.5? Doesn't Compat provide all the needed features?

nalimilan:
Nevermind, I hadn't noted the discussion on the other issue regarding multithreading.

dmbates:
The checks are green on this PR now.  Can I go ahead and merge?

PR #155, opened 2016-10-01
Specialized update methods for canonical links.

dmbates:
- Define methods for `updateμ!` specific to Distribution/Link combinations for the canonical link.
- Having specific methods removes some of the numerical issues in evaluating the working weights.
- Remove the unused parameter `wt` in `devresid` methods.  All calls to these methods were with `wt = 1` and it was used as a multiplier.
- Allow for multithreaded evaluation of `updateμ!` with canonical links.
- Switch tests to use `@testset` and replace calls to `@test_approx_eq a b` to `@test isapprox(a,b)`, allowing for more flexibility in setting `atol` or `rtol`

### DIFF
--- a/src/glmfit.jl
+++ b/src/glmfit.jl
@@ -12,7 +12,7 @@ type GlmResp{V<:FPVector,D<:UnivariateDistribution,L<:Link} <: ModResp       # r
 end
 
 function GlmResp{V<:FPVector, D, L}(y::V, d::D, l::L, η::V, μ::V, off::V, wts::V)
-    if d == Binomial
+    if d == Binomial()
         for yy in y
             0. <= yy <= 1. || throw(ArgumentError("$yy in y is not in [0,1]"))
         end
--- a/src/glmfit.jl
+++ b/src/glmfit.jl
@@ -119,7 +119,7 @@ function updateμ!{T,D,L}(r::GlmResp{T,D,L})
         dμdη = mueta(L(), ηi)
         yi = y[i]
         wrkres[i] = (yi - μi)/dμdη
-        dres[i] = devresid(r.d, yi, μi, one(yi))
+        dres[i] = devresid(r.d, yi, μi)
         wrkwt[i] = abs2(dμdη) / max(eps(), glmvar(r.d, L(), μi, ηi))
     end
 end
--- a/src/glmfit.jl
+++ b/src/glmfit.jl
@@ -26,7 +26,7 @@ type GlmResp{V<:FPVector,D<:UnivariateDistribution,L<:Link} <: ModResp       # r
         lo = length(off)
         lo == 0 || lo == n || error("offset must have length $n or length 0")
         res = new(y,d,l,similar(y),eta,mu,similar(y),off,similar(y),wts,similar(y),similar(y))
-        updatemu!(res, eta)
+        updateμ!(res, eta)
         res
     end
 end
@@ -34,7 +34,7 @@ end
 # returns the sum of the squared deviance residuals
 deviance(r::GlmResp) = sum(r.devresid)
 
-function updatemu!{T<:FPVector}(r::GlmResp{T}, linPr::T)
+function updateμ!{T<:FPVector}(r::GlmResp{T}, linPr::T)
     y = r.y
     dist = r.d
     link = r.l
@@ -144,11 +144,11 @@ function _fit!(m::AbstractGLM, verbose::Bool, maxIter::Integer, minStepFac::Real
         copy!(p.beta0, start)
         fill!(p.delbeta, 0)
         linpred!(lp, p, 0)
-        updatemu!(r, lp)
+        updateμ!(r, lp)
     else
         delbeta!(p, wrkresp(r), wrkwt!(r))
         linpred!(lp, p)
-        updatemu!(r, lp)
+        updateμ!(r, lp)
         installbeta!(p)
     end
     devold = deviance(m)
@@ -158,7 +158,7 @@ function _fit!(m::AbstractGLM, verbose::Bool, maxIter::Integer, minStepFac::Real
         try
             delbeta!(p, r.wrkresid, wrkwt!(r))
             linpred!(lp, p)
-            updatemu!(r, lp)
+            updateμ!(r, lp)
             dev = deviance(m)
         catch e
             isa(e, DomainError) ? (dev = Inf) : rethrow(e)
@@ -167,7 +167,7 @@ function _fit!(m::AbstractGLM, verbose::Bool, maxIter::Integer, minStepFac::Real
             f /= 2.
             f > minStepFac || error("step-halving failed at beta0 = $(p.beta0)")
             try
-                updatemu!(r, linpred(p, f))
+                updateμ!(r, linpred(p, f))
                 dev = deviance(m)
             catch e
                 isa(e, DomainError) ? (dev = Inf) : rethrow(e)
@@ -216,7 +216,7 @@ function StatsBase.fit!(m::AbstractGLM, y; wts=nothing, offset=nothing, dofit::B
     isa(wts, Void) || copy!(r.wts, wts)
     isa(offset, Void) || copy!(r.offset, offset)
     initialeta!(r.d, r.l, r.eta, r.y, r.wts, r.offset)
-    updatemu!(r, r.eta)
+    updateμ!(r, r.eta)
     fill!(m.pp.beta0, 0)
     m.fit = false
     if dofit
--- a/src/glmfit.jl
+++ b/src/glmfit.jl
@@ -31,14 +31,18 @@ function GlmResp{V<:FPVector, D, L}(y::V, d::D, l::L, η::V, μ::V, off::V, wts:
     res
 end
 
-# returns the sum of the squared deviance residuals
+"""
+    deviance(r::GlmResp)
+
+Return the sum of the squared deviance residuals
+"""
 deviance(r::GlmResp) = sum(r.devresid)
 
 """
     wtscale!{T<:FPVector}(devr::T, wkwt::T, wt::T)
 
 Scale the deviance residuals, `devr`, and the working weights, `wkwt`, by `wt`,
-if `wt` is nonempty.
+when `wt` is nonempty.
 """
 function wtscale!{T<:FPVector}(devr::T, wkwt::T, wt::T)
     if !isempty(wt)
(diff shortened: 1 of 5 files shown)


src/glmfit.jl
```diff
@@ -1,96 +1,95 @@
...
-    l::L
     devresid::V                                # (squared) deviance residuals
     eta::V                                     # linear predictor
     mu::V                                      # mean response
     mueta::V                                   # derivative of mu w.r.t. eta
     offset::V                                  # offset added to linear predictor (usually 0)
-    var::V                                     # (unweighted) variance at current mu
     wts::V                                     # prior weights
-    wrkwts::V                                  # working weights
+    wrkwt::V                                   # working weights
```
nalimilan: Why remove the final `s` if you keep it for `wts` and for `wrkwts` below?

src/glmfit.jl
```diff
@@ -12,7 +12,7 @@ type GlmResp{V<:FPVector,D<:UnivariateDistribution,L<:Link} <: ModResp       # r
 end
 
 function GlmResp{V<:FPVector, D, L}(y::V, d::D, l::L, η::V, μ::V, off::V, wts::V)
-    if d == Binomial
+    if d == Binomial()
```
nalimilan: So this was broken before, and not caught by the tests?

src/glmfit.jl
```diff
@@ -31,14 +31,18 @@ function GlmResp{V<:FPVector, D, L}(y::V, d::D, l::L, η::V, μ::V, off::V, wts:
     res
 end
 
-# returns the sum of the squared deviance residuals
+"""
+    deviance(r::GlmResp)
+
+Return the sum of the squared deviance residuals
```
nalimilan: Missing final dot.

nalimilan: Actually, I don't think we should add a docstring for this method, as we already have one for `StatisticalModel` and for `LinearModel`. `GlmResp` sounds like an internal type to me, which should not be shown to the user when calling `?deviance`.

src/glmfit.jl
```diff
@@ -52,35 +56,76 @@ end
...
+end
+
+function updateμ!{T<:FPVector,D<:Normal}(r::GlmResp{T,D,IdentityLink})
+    y, η, μ, wrkres, wrkwt, dres = r.y, r.eta, r.mu, r.wrkresid, r.wrkwt, r.devresid
+
+    @inbounds Threads.@threads for i in eachindex(η)
+        μi = μ[i] = η[i]
+        wrkwt[i] = 1
+        yi = y[i]
+        wrkresi = wrkres[i] = (yi - μi)
```
nalimilan: I don't think using a temporary `wrkresi` variable makes any difference in performance, especially not inside an `@inbounds` block.

src/glmfit.jl
```diff
@@ -52,35 +56,76 @@ end
...
+        dres[i] = devresid(r.d, yi, μi, one(yi))
         wrkwt[i] = abs2(dμdη) / max(eps(), glmvar(r.d, L(), μi, ηi))
     end
-    wtscale!(devres, wrkwt, wts)
-    r
 end
 
 """
-    wrkresp(r::GlmResp)
+    wrkresp(r::GlmResp){}
```
nalimilan: Typo?

src/glmfit.jl
```diff
@@ -89,6 +134,23 @@ function wrkresp(r::GlmResp)
...
+    broadcast!(+, v, r.eta, r.wrkresid)
+    isempty(r.offset) ? v : broadcast!(-, v, v, r.offset)
+end
+
+"""
+    wrkwt(r::GlmResp)
+
+Return the working weights.
+
+The result is `r.wrkwt`, not a copy of `r.wrkwt`
```
nalimilan: Missing final dot here and below.

nalimilan: Also it's a bit weird to say this is `r.wrkwt` in the docs: if that's guaranteed to be the case, why provide an accessor at all?

test/runtests.jl
```diff
@@ -8,257 +8,314 @@ end
...
-@test_approx_eq aicc(gm1) 76.76131840195768
-@test_approx_eq bic(gm1) 57.74744128863877
-@test_approx_eq coef(gm1)[1:3] [3.044522437723423,-0.45425527227759555,-0.29298712468147375]
+form = DataFrame(Any[[0.1,0.3,0.5,0.6,0.7,0.9],[0.086,0.269,0.446,0.538,0.626,0.782]],
+    [:Carb, :OptDen])
+
+@testset "lm" begin
+    lm1 = fit(LinearModel, OptDen ~ Carb, form)
+    test_show(lm1)
+    @test isapprox(coef(lm1), collect(linreg(form[:Carb], form[:OptDen])))
```
nalimilan: FWIW, you know you can use `≈` (`\approx[tab]` at the REPL)?

test/runtests.jl
```diff
@@ -214,12 +214,14 @@ admit_agr2 = DataFrame(Any[[61., 151, 121, 67], [33., 54, 28, 12], pool([1, 2, 3
...
 
+## The model matrix here is singular so tests like the deviance are just round off error
 @testset "Binomial LogitLink aggregated" begin
     gm15 = fit(GeneralizedLinearModel, p ~ rank, admit_agr2, Binomial(), wts=admit_agr2[:count])
     test_show(gm15)
     @test df(gm15) == 4
     @test nobs(gm15) == 400
-    @test isapprox(deviance(gm15), -2.4424906541753456e-15, rtol = 1e-7)
+# The model matrix is singular so the deviance is essentially round-off error
+#    @test isapprox(deviance(gm15), -2.4424906541753456e-15, rtol = 1e-7)
```
nalimilan: That doesn't mean it shouldn't be tested. Maybe increase tolerance a bit?

test/runtests.jl
ararslan: And if you're on a Mac you can do <kbd>option</kbd>+<kbd>x</kbd>

src/glmtools.jl
```diff
@@ -39,7 +39,7 @@ mueta(::LogLink, η) = exp(η)
 
 linkfun(::ProbitLink, μ) = -sqrt2 * erfcinv(2μ)
 linkinv(::ProbitLink, η) = erfc(-η / sqrt2) / 2
-mueta(::ProbitLink, η) = exp(-η^2 / 2) / sqrt2π
+mueta(::ProbitLink, η) = exp(-abs2(η) / 2) / sqrt2π
```
nalimilan: FWIW, both versions give exactly the same native code for `Int` and `Float64`.

src/glmtools.jl
```diff
@@ -67,26 +67,26 @@ mustart(::Gamma, y, wt) = y
 mustart(::Normal, y, wt) = y
 mustart(::Poisson, y, wt) = y + oftype(y, 0.1)
 
-function devresid(::Bernoulli, y, μ, wt)
+function devresid(::Bernoulli, y, μ)
     if y == 1
-        return -2 * wt * log(μ)
```
nalimilan: Aren't these needed to support weighted model fits?

src/glmfit.jl
```diff
@@ -52,35 +56,76 @@ end
...
+    updateμ!(r)
+    if !isempty(r.wts)
+        r.devresid .*= r.wts
+        r.wrkwt .*= r.wts
+    end
+    r
+end
+
+function updateμ!{T<:FPVector,D<:Union{Bernoulli,Binomial}}(r::GlmResp{T,D,LogitLink})
+    y, η, μ, wrkres, wrkwt, dres = r.y, r.eta, r.mu, r.wrkresid, r.wrkwt, r.devresid
```
nalimilan: Unrelated, but we could probably make `GlmResp` immutable, which would avoid the need to defining temporary variables to ensure performance.

nalimilan:
Looks like a nice cleanup overall!

src/glmfit.jl
dmbates: Yes.

src/glmfit.jl
dmbates: Okay.  I will omit this docstring.

src/glmfit.jl
dmbates: Is there a usage of `wrkwts` below?  I thought I had eliminated them all.

Perhaps the `wts` name should be replaced by `weights`.  I think I was too cryptic when I first started writing Julia code.

This still would not resolve `wrkwt` vs `weights` regarding singular or plural.  Most of the names of vectors, `wrkresid`, `wrkwt`, `devresid`, `y`, ... are singular.  For consistency it might be best to make them all singular.

src/glmfit.jl
dmbates: Okay.  Is it worthwhile checking if these temporary variables within the loops are redundant?

src/glmfit.jl
dmbates: Yes.  Thanks for catching that.

src/glmfit.jl
dmbates: To me, accessing a member of a type directly is poor form.  It is better to use an extractor function if available.  However this one is so trivial that it cold easily be removed.

test/runtests.jl
dmbates: The reason for using `isapprox` is because the optional `atol` and `rtol` arguments are available. Sometimes the same calculation on different platforms can produce slightly different results and tolerances are helpful.

src/glmtools.jl
dmbates: I find it easier to read this way.  `-η^2` has too many operators to parse for me.

src/glmfit.jl
dmbates: Okay, will do.

src/glmtools.jl
dmbates: No.  The weights are applied after `updateμ!` now.

src/glmfit.jl
nalimilan: As you like. ;-)

src/glmfit.jl
nalimilan: I just think you don't need to mention that it doesn't make a copy, that's a common pattern in Julia unless otherwise indicated.

test/runtests.jl
nalimilan: Yes, but here you don't specify the tolerance. Not a big deal anyway.

src/glmfit.jl
```diff
@@ -1,14 +1,28 @@
...
     d::D
-    devresid::V                                # (squared) deviance residuals
-    eta::V                                     # linear predictor
-    mu::V                                      # mean response
-    mueta::V                                   # derivative of mu w.r.t. eta
-    offset::V                                  # offset added to linear predictor (usually 0)
-    wts::V                                     # prior weights
-    wrkwt::V                                   # working weights
-    wrkresid::V                                # working residuals
+    "`devresid`: the squared deviance residuals"
```
nalimilan: I could be wrong, but I don't think you need to repeat the name of the field in the docstring.

nalimilan:
Looks good to me, at least for what I understand.

PR #156, opened 2016-10-06
Update df to dof to accomodate StatsBase change

ararslan:
This bumps the StatsBase requirement to 0.11.0, which deprecates `df` in favor of `dof`, and changes all occurrences of `df` to `dof` to accomodate that change.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,5 +1,5 @@
 julia 0.5
 Distributions 0.10.0
-StatsBase 0.8.3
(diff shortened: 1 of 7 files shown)


ararslan:
Looks like the `dof` methods are actually defined in DataFrames for some reason. I've submitted a PR there. Once that's merged and tagged, I'll update this PR to require the patched version of DataFrames, then things should be good.

PR #158, opened 2016-10-12
rename df -> dof

simonbyrne:


simonbyrne:
Ah, I forgot to pull new changes.

PR #160, opened 2016-10-20
updateμ! method for Gamma/Inverse, properly name wrkresp!

dmbates:
The mutating function `wrkresp!` was accidentally written as a method for `wrkresp`.

This PR also includes a specialized `updateμ!` method for the Gamma distribution and its canonical link `InverseLink`.

The reason for writing the specialized `updateμ!` methods is because terms that may be close to zero cancel in the evaluation of the `wrkresid` for the canonical link.

### DIFF
--- a/src/glmfit.jl
+++ b/src/glmfit.jl
@@ -74,6 +74,20 @@ function updateμ!{T<:FPVector,D,L}(r::GlmResp{T,D,L}, linPr::T)
     r
 end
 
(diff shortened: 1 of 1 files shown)


dmbates:
The travis failure on the push seems to be the ongoing problem with cloning DataFrames in the nightlies.

dmbates:
Any objections to my merging this?

PR #161, opened 2016-11-15
Eliminate memory allocation for weights adjustments

dmbates:
Using memory profiling I found that memory was being allocated in a convenience function that was called in only one place.

### DIFF
--- a/src/glmfit.jl
+++ b/src/glmfit.jl
@@ -56,7 +56,7 @@ function updateμ!{T<:FPVector,D,L}(r::GlmResp{T,D,L}, linPr::T)
     updateμ!(r)
     if !isempty(r.wts)
         map!(*, r.devresid, r.devresid, r.wts)
-	map!(*, r.wrkwt, r.wrkwt, r.wts)
+        map!(*, r.wrkwt, r.wrkwt, r.wts)
     end
     r
 end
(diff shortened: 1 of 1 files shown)


src/glmfit.jl
```diff
@@ -68,8 +55,8 @@ function updateμ!{T<:FPVector,D,L}(r::GlmResp{T,D,L}, linPr::T)
     isempty(r.offset) ? copy!(r.eta, linPr) : broadcast!(+, r.eta, linPr, r.offset)
     updateμ!(r)
     if !isempty(r.wts)
-        r.devresid .*= r.wts
-        r.wrkwt .*= r.wts
+        map!(*, r.devresid, r.devresid, r.wts)
+	map!(*, r.wrkwt, r.wrkwt, r.wts)
```
ararslan: This one line is indented with a tab, whereas the surrounding code is indented by multiples of 4 spaces. If you could change it to spaces for consistency that would be great.

dmbates:
@ararslan Thanks for catching that.  I had noticed the strange indentation in one diff but then I didn't see it when I looked at the file on github.  Now fixed.

ararslan:
@nalimilan Do you mean `.*=`? I've not seen `.=*`. The former exists in 0.5 and works for any iterable that supports indexed assignment (AFAICT).

dmbates:
@nalimilan I believe that in v0.5 modifying operators like `.*=` are not yet in-place.

ararslan:
Oh sorry, "in-place" in the sense of an in-place operation, not in the sense of existence. Yes, they won't be in place until 0.6.

PR #162, opened 2016-11-30
Make linear model easier to extend to new array types.

andreasnoack:
Introduce `fit!` and create `LinearModel` instance before fitting of the linear model.

Use Cholesky by default.

These changes are motivated by an attempt to extend `GLM.jl` to other array types and they shouldn't have an impact on normal usage of the package.

I think it was cumbersome to construct the correct `LmResp` and `LinPred` data types without constructing the instances which the old version of `fit` required. Here I instead instantiate the types, create the `LinearModel` and call `fit!`.

The change to using Cholesky is not strictly necessary but I think that a Cholesky is implemented slightly more than the QR for other matrix types. It is also much faster to use Cholesky and my view is that the error due to numerical precision loss are insignificant relative to the statistical error, i.e. if `X'X` is so badly conditioned that using the QR actually matters, then the variance of the estimates is enormous.

### DIFF
--- a/src/lm.jl
+++ b/src/lm.jl
@@ -16,6 +16,11 @@ end
 convert{V<:FPVector}(::Type{LmResp{V}}, y::V) =
     LmResp{V}(zeros(y), similar(y, 0), similar(y, 0), y)
 
+function convert{T<:Real}(::Type{LmResp}, y::AbstractVector{T})
+    yy = float(y)
+    convert(LmResp{typeof(yy)}, yy)
+end
+
 function updateμ!{V<:FPVector}(r::LmResp{V}, linPr::V)
     n = length(linPr)
     length(r.y) == n || error("length(linPr) is $n, should be $(length(r.y))")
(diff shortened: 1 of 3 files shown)


src/lm.jl
```diff
@@ -100,23 +105,15 @@ end
...
-    pp = LinPredT(X)
-    installbeta!(delbeta!(pp, rr.y))
-    updateμ!(rr, linpred(pp, 0.0))
-    LinearModel(rr, pp)
-end
-function fit(::Type{LinearModel}, X::AbstractMatrix, y::Vector)
-    yy = float(y)
-    T = eltype(yy)
-    return fit(LinearModel{LmResp{typeof(yy)}, DensePredQR{T}}, float(X), yy)
+function StatsBase.fit!{LmRespT<:LmResp, LinPredT<:LinPred}(obj::LinearModel{LmRespT,LinPredT})
```
andreasnoack: I've removed them now.

PR #164, opened 2016-12-09
"df_residual" -> "dof_residual" in README

maximsch2:
Edit README to point to the non-deprecated function.


andreasnoack:
Thanks. I'll merge when https://github.com/JuliaLang/METADATA.jl/pull/7246 is in to avoid that the test banner turns red.

PR #170, opened 2016-12-13
Stop Travis testing on OS X

andreasnoack:
It is unlikely that anything in the package would fail on OS X only and the OS X queue is always long

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: julia
 os:
     - linux
-    - osx
 julia:
     - 0.5
     - nightly


nalimilan:
I'd rather keep it, it's not like we're often in a hurry. If sometimes you really want to merge something without waiting for the OS X build to complete, just go ahead.

andreasnoack:
I won't fight much for this but since Travis can't keep up with the OS X demand I think it is reasonable to disable OS X testing for projects where platform specific test failures are unlikely. Then the queue can be shorter for those who need  OS X testing.

.travis.yml
```diff
@@ -1,7 +1,6 @@
 language: julia
 os:
     - linux
-    - osx
```
ararslan: What I would do is just comment out this line and add a comment with a reference to this PR. That way it's clearer why it was removed.

ararslan:
I had been thinking about this as well and I agree with Andreas. Though greater platform coverage is always desirable, the OS X Travis backlog hinders more rapid development. We can always try moving the OS X building to CircleCI, if anyone wants to test that out, but in the meantime this seems like an unfortunate but overall good idea to me.

nalimilan:
FWIW, Tony just said this on another issue:
> Travis has said they should be announcing increased OS X capacity soon which would be good news.

https://github.com/JuliaLang/julia/pull/19586#issuecomment-266971272

So maybe wait for a bit more and see how it evolves.

andreasnoack:
Ok. Let's wait and see if things improve.

Issue #101, opened 2015-03-11
Interval functionality in predict() function?


Mokubyow:
In R the predict function has an interval parameter where you can specify and retrieve a 95% confidence or prediction interval. Is that possible with the predict() function in Julia? If not it would be a good feature add.

kleinschmidt:
It doesn't look like it's currently possible, but I agree it would be nice to have.  As always, pull requests are encouraged, and this should be pretty straightforward to implement if you're interested in taking a crack at it.

PR #171, opened 2016-12-14
confidence intervals in predict()

mkborregaard:
Adds basic support for confidence intervals in predict() (see https://github.com/JuliaStats/GLM.jl/issues/101)

I have so far only added to the `lm()` function, but it should be fairly easy to transfor to the `glm()` method. Also, the `predict` call is through `DataFrames`, but I have not touched that package, so it only works inside the module (I use Atom’s ability to set the working module). Also, it is not extensively tested. Oh, and because the intervals return a `Matrix`, this makes the `predict` function not be type stable, so some thought is needed here.
But it is something to build upon.

A quick example (notice I have to jump through some hoops because I bypass DataFrames):
```julia
using DataFrames, RCall
a = randn(100); b = randn(100); c = a .+ b .+ 0.8randn(100);
df = DataFrame(a = a, b = b, c = c);
mod = lm(c ~ a + b, df)

predict(LinearModel(mod.model), hcat(ones(100),Matrix(df[:,1:2])), :confint)
R"""
mod <- lm(c ~ a + b, $(df))
predict(mod, interval = "confidence")
"""
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,6 +72,15 @@ julia> predict(OLS)
  4.33333
  6.83333
 
+# Prediction with new data and confidence values
+ julia> newX = DataFrame(X=[2,3,4]);
+ julia> predict(OLS, newX, :confint)
+ 3×3 Array{Float64,2}:
+  4.33333  1.33845   7.32821
+  6.83333  2.09801  11.5687
+  9.33333  1.40962  17.257  
+# The columns of the matrix are prediction, lower and upper confidence interval
+
 ```
 
 Probit Regression:
--- a/README.md
+++ b/README.md
@@ -79,7 +79,7 @@ julia> predict(OLS)
   4.33333  1.33845   7.32821
   6.83333  2.09801  11.5687
   9.33333  1.40962  17.257  
-# The columns of the matrix are prediction, lower and upper confidence interval
+# The columns of the matrix are prediction, 95% lower and upper confidence bounds
 
 ```
 
--- a/src/lm.jl
+++ b/src/lm.jl
@@ -158,7 +158,7 @@ end
 
 predict(mm::LinearModel, newx::AbstractMatrix) = newx * coef(mm)
 
-function predict(mm::LinearModel, newx::AbstractMatrix, interval_type::Symbol, level = 0.95)
+function predict(mm::LinearModel, newx::AbstractMatrix, interval_type::Symbol; level = 0.95, alpha = 1 - level)
     retmean = newx * coef(mm)
     interval_type == :confint || error("only :confint is currently implemented") #:predint will be implemented
 
@@ -166,7 +166,7 @@ function predict(mm::LinearModel, newx::AbstractMatrix, interval_type::Symbol, l
     residvar = (ones(size(newx,2),1) * deviance(mm)/dof_residual(mm))
     retvariance = (newx/R).^2 * residvar
 
-    interval = quantile(TDist(dof_residual(mm)), (1 - level)/2) * sqrt.(retvariance)
+    interval = quantile(TDist(dof_residual(mm)), alpha/2) * sqrt.(retvariance)
     retmean, retmean .+ interval, retmean .- interval
 end
 
--- a/src/lm.jl
+++ b/src/lm.jl
@@ -156,7 +156,23 @@ function coeftable(mm::LinearModel)
               ["x$i" for i = 1:size(mm.pp.X, 2)], 4)
 end
 
-predict(mm::LinearModel, newx::Matrix) =  newx * coef(mm)
+# Different methods of prediction for the different factorization types
+getR(x::DensePredChol) = UpperTriangular(x.chol.factors)
+getR(x::DensePredQR) = x.qr[:R]
+
+function predict(mm::LinearModel, newx::Matrix; error::Symbol = :none)
+    prediction = newx * coef(mm)
+    error == :none && return prediction
+    error == :confint || error("specify error as :none or :confint") #:predint will be implemented
+    R = getR(mod.model.pp)
+
+    tmp = newx * (R\Diagonal(ones(3,3)))
+    tmp = tmp.^2 * (ones(3,1) * deviance(mod)/dof_residual(mod))
+    tval = quantile(TDist(dof_residual(mod)), 0.025)
+    interval = tval * sqrt(tmp)
+
+    hcat(prediction, prediction + interval, prediction - interval)
+end
 
 function confint(obj::LinearModel, level::Real)
     hcat(coef(obj),coef(obj)) + stderr(obj) *
--- a/src/lm.jl
+++ b/src/lm.jl
@@ -160,15 +160,15 @@ end
 getR(x::DensePredChol) = UpperTriangular(x.chol.factors)
 getR(x::DensePredQR) = x.qr[:R]
 
-function predict(mm::LinearModel, newx::Matrix; error::Symbol = :none)
+function predict(mm::LinearModel, newx::Matrix; error::Symbol = :none, level = 0.95)
     prediction = newx * coef(mm)
     error == :none && return prediction
     error == :confint || error("specify error as :none or :confint") #:predint will be implemented
-    R = getR(mod.model.pp)
+    R = getR(mm.pp)
 
     tmp = newx * (R\Diagonal(ones(3,3)))
-    tmp = tmp.^2 * (ones(3,1) * deviance(mod)/dof_residual(mod))
-    tval = quantile(TDist(dof_residual(mod)), 0.025)
+    tmp = tmp.^2 * (ones(3,1) * deviance(mm)/dof_residual(mm))
+    tval = quantile(TDist(dof_residual(mm)), 1. -level/2)
     interval = tval * sqrt(tmp)
 
     hcat(prediction, prediction + interval, prediction - interval)
--- a/src/lm.jl
+++ b/src/lm.jl
@@ -156,30 +156,20 @@ function coeftable(mm::LinearModel)
               ["x$i" for i = 1:size(mm.pp.X, 2)], 4)
 end
 
-# Different methods of prediction for the different factorization types
-
 predict(mm::LinearModel, newx::AbstractMatrix) = newx * coef(mm)
 
 function predict(mm::LinearModel, newx::AbstractMatrix, interval_type::Symbol, level = 0.95)
-    prediction = newx * coef(mm)
+    retmean = newx * coef(mm)
     interval_type == :confint || error("only :confint is currently implemented") #:predint will be implemented
 
     R = cholfact!(mm.pp)[:U] #get the R matrix from the QR factorization
-    residvar = (ones(3,1) * deviance(mm)/dof_residual(mm))'
+    residvar = (ones(size(newx,2),1) * deviance(mm)/dof_residual(mm))
+    retvariance = (newx/R).^2 * residvar
 
-    function _local(x::StridedVector)
-        Ac_ldiv_B!(R,x)
-        residvar * (x⋅x)
-    end
-
-    interval = similar(prediction)
-    for i in 1:length(interval)
-        interval[i] = _local(newx[i,:])[1]
-    end
-
-    interval = quantile(TDist(dof_residual(mm)), (1 - level)/2) * sqrt(interval)
-    prediction, prediction .+ interval, prediction .- interval
+    interval = quantile(TDist(dof_residual(mm)), (1 - level)/2) * sqrt(retvariance)
+    retmean, retmean .+ interval, retmean .- interval
 end
+
 
 function confint(obj::LinearModel, level::Real)
     hcat(coef(obj),coef(obj)) + stderr(obj) *
--- a/src/lm.jl
+++ b/src/lm.jl
@@ -157,22 +157,30 @@ function coeftable(mm::LinearModel)
 end
 
 # Different methods of prediction for the different factorization types
-getR(x::DensePredChol) = UpperTriangular(x.chol.factors)
-getR(x::DensePredQR) = x.qr[:R]
 
-predict((::LinearModel, newx::Matrix)) = newx * coef(mm)
+predict(mm::LinearModel, newx::AbstractMatrix) = newx * coef(mm)
 
-function predict(mm::LinearModel, newx::Matrix, interval_type::Symbol, level = 0.95)
+function predict(mm::LinearModel, newx::AbstractMatrix, interval_type::Symbol, level = 0.95)
     prediction = newx * coef(mm)
-    interval_type == :confint || error("only :confint is currently implemented") #:predint will be implemented
-    R = getR(mm.pp)
+    #interval_type == :confint || error("only :confint is currently implemented") #:predint will be implemented
 
-    tmp = newx * (R\Diagonal(ones(3,3)))
-    tmp = tmp.^2 * (ones(3,1) * deviance(mm)/dof_residual(mm))
-    tval = quantile(TDist(dof_residual(mm)), 1. -level/2)
-    interval = tval * sqrt(tmp)
+    R = cholfact!(mm.pp)[:U] #get the R matrix from the QR factorization
+    # get the multipliers for the final interval
+    residvar = (ones(3,1) * deviance(mm)/dof_residual(mm))'
+    tval = quantile(TDist(dof_residual(mm)), (1 - level)/2)
 
-    hcat(prediction, prediction + interval, prediction - interval)
+    function _local(x::StridedVector)
+        Ac_ldiv_B!(R,x)
+        residvar * (x⋅x)
+    end
+
+    interval = similar(prediction)
+    for i in 1:length(interval)
+        interval[i] = _local(newx[i,:])[1]
+    end
+
+    interval = tval * sqrt(interval)
+    prediction, prediction .+ interval, prediction .- interval
 end
 
 function confint(obj::LinearModel, level::Real)
--- a/src/lm.jl
+++ b/src/lm.jl
@@ -159,18 +159,14 @@ end
 predict(mm::LinearModel, newx::AbstractMatrix) = newx * coef(mm)
 
 """
-For linear models, specifying `interval_type` will return a 3-tuple with the
-predicted values, the lower and the higher confidence bound. Confidence intervals
-delimit the uncertainty of the estimate of the predicted values, prediction
-intervals delimit the estimated bounds for any new data points from the
-population.
-
-# Arguments
-* `interval_type::Symbol`: one of `:confint` or `:predint`
-* `level=0.95`: the level of the confidence interval
-* `alpha=1-level`: an alternative way of specifying the level of the interval
+    predict(mm::LinearModel, newx::AbstractMatrix, interval_type::Symbol, level = 0.95)
+    
+Specifying `interval_type` will return a 3-column matrix with the prediction and
+the low and high confidence bounds for a given `level` (0.95 equates alpha = 0.05).
+Valid values of `interval_type` are `:confint` delimiting the  uncertainty of the
+predicted relationship, and `:predint` delimiting estimated bounds for new data points.
 """
-function predict(mm::LinearModel, newx::AbstractMatrix, interval_type::Symbol; level = 0.95, alpha = 1 - level)
+function predict(mm::LinearModel, newx::AbstractMatrix, interval_type::Symbol, level = 0.95)
     retmean = newx * coef(mm)
     interval_type == :confint || error("only :confint is currently implemented") #:predint will be implemented
 
@@ -178,8 +174,8 @@ function predict(mm::LinearModel, newx::AbstractMatrix, interval_type::Symbol; l
     residvar = (ones(size(newx,2),1) * deviance(mm)/dof_residual(mm))
     retvariance = (newx/R).^2 * residvar
 
-    interval = quantile(TDist(dof_residual(mm)), alpha/2) * sqrt.(retvariance)
-    retmean, retmean .+ interval, retmean .- interval
+    interval = quantile(TDist(dof_residual(mm)), (1 - level)/2) * sqrt.(retvariance)
+    hcat(retmean, retmean .+ interval, retmean .- interval)
 end
 
 
--- a/src/lm.jl
+++ b/src/lm.jl
@@ -166,7 +166,7 @@ function predict(mm::LinearModel, newx::AbstractMatrix, interval_type::Symbol, l
     residvar = (ones(size(newx,2),1) * deviance(mm)/dof_residual(mm))
     retvariance = (newx/R).^2 * residvar
 
-    interval = quantile(TDist(dof_residual(mm)), (1 - level)/2) * sqrt(retvariance)
+    interval = quantile(TDist(dof_residual(mm)), (1 - level)/2) * sqrt.(retvariance)
     retmean, retmean .+ interval, retmean .- interval
 end
 
--- a/src/lm.jl
+++ b/src/lm.jl
@@ -169,6 +169,7 @@ predicted relationship, and `:predint` delimiting estimated bounds for new data 
 function predict(mm::LinearModel, newx::AbstractMatrix, interval_type::Symbol, level::Real = 0.95)
     retmean = newx * coef(mm)
     interval_type == :confint || error("only :confint is currently implemented") #:predint will be implemented
+    length(mm.rr.wts) == 0 || error("prediction with confidence intervals not yet implemented for weighted regression")
 
     R = cholfact!(mm.pp)[:U] #get the R matrix from the QR factorization
     residvar = (ones(size(newx,2),1) * deviance(mm)/dof_residual(mm))
--- a/src/lm.jl
+++ b/src/lm.jl
@@ -159,14 +159,14 @@ end
 predict(mm::LinearModel, newx::AbstractMatrix) = newx * coef(mm)
 
 """
-    predict(mm::LinearModel, newx::AbstractMatrix, interval_type::Symbol, level = 0.95)
-    
+    predict(mm::LinearModel, newx::AbstractMatrix, interval_type::Symbol, level::Real = 0.95)
+
 Specifying `interval_type` will return a 3-column matrix with the prediction and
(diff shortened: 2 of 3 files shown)


mkborregaard:
I'm a little uncomfortable with how I get R from the Cholesky factorization, but I seem to have no control over which one `lm` chooses to use (DensePredQR or DensePredChol)

nalimilan:
To avoid the type instability, you could only return the confidence interval if a third argument `level` is provided.

mkborregaard:
Absolutely - but am I right in assuming these functions are currently never called by the user, but always called from within DataFrames? I guess then that would be where to implement different methods for different signatures.

nalimilan:
I think people can still call these methods manually if they want to enter the data in matrix form rather than in tabular form. DataFrames (or StatsModels in the future) should only implement the wrapper from tabular to matrix.

mkborregaard:
Right, I've updated the code to do that.

src/lm.jl
```diff
@@ -156,7 +156,24 @@ function coeftable(mm::LinearModel)
...
 end
 
-predict(mm::LinearModel, newx::Matrix) =  newx * coef(mm)
+# Different methods of prediction for the different factorization types
+getR(x::DensePredChol) = UpperTriangular(x.chol.factors)
+getR(x::DensePredQR) = x.qr[:R]
+
+predict((::LinearModel, newx::Matrix)) = newx * coef(mm)
+
+function predict(mm::LinearModel, newx::Matrix, interval_type::Symbol, level = 0.95)
```
ararslan: From my experience with other statistical software, it's more common to specify alpha than the confidence level, with the implicit understanding that the confidence level is then 1 - alpha.

src/lm.jl
```diff
@@ -156,7 +156,24 @@ function coeftable(mm::LinearModel)
...
+    prediction = newx * coef(mm)
+    interval_type == :confint || error("only :confint is currently implemented") #:predint will be implemented
+    R = getR(mm.pp)
```
andreasnoack: This can be done with `cholfact!(mm.pp)[:U]`

src/lm.jl
```diff
@@ -156,7 +156,24 @@ function coeftable(mm::LinearModel)
...
+
+    tmp = newx * (R\Diagonal(ones(3,3)))
```
andreasnoack: probably `tmp = newx/R` but I'll suggest that this methods is written for `x::StridedVector` such that this becomes `tmp = R'\x`. Actually, I think it would make sense to make a non-allocating version where `x` is destructed during the computation. Then it would be `Ac_ldiv_B!(R, x)`.

src/lm.jl
```diff
@@ -156,7 +156,24 @@ function coeftable(mm::LinearModel)
...
+    tmp = tmp.^2 * (ones(3,1) * deviance(mm)/dof_residual(mm))
```
andreasnoack: Maybe `varx = (tmp⋅tmp)*deviance(mm)/dof_residual(mm)` which would make it non-allocating. Using `tmp` for two different types would make the code harder to optimize for the compiler.

src/lm.jl
```diff
@@ -156,7 +156,24 @@ function coeftable(mm::LinearModel)
...
+    tval = quantile(TDist(dof_residual(mm)), 1. -level/2)
```
andreasnoack: Just use integer `1` here then this would also do the right thing when we support `Float32` in the future.

src/lm.jl
```diff
@@ -156,7 +156,24 @@ function coeftable(mm::LinearModel)
...
+    interval = tval * sqrt(tmp)
+
+    hcat(prediction, prediction + interval, prediction - interval)
```
andreasnoack: Again, if this is written for vector input then it doesn' have to be allocating. Then we can have another version written for matrix input that just loops over the rows and `X` and calls the method for vector.

src/lm.jl
mkborregaard: Right, I am happy to change it to comply with any standards set by this community. Which software does this, though? I come from R, where the convention is to define `level` for confidence intervals, so I think of that as the most intuitive. In support of this is also naming - this is called "the 95% confidence interval", not the "0.05 confidence interval".

src/lm.jl
mkborregaard: Nice! - done

src/lm.jl
mkborregaard: Not quite sure what you mean here - do you mean to start with `x = view(newx,:,:)`? If I do that, neither`Ac_ldiv_B!(R,x)` or `tmp = R'\x` will run. Sorry, I am not super-sharp on julias facilities for low-level linear algebra

src/lm.jl
andreasnoack: Most of the statistics code used to be hard coded for `Float64` but we are trying to make it more generic.

src/lm.jl
mkborregaard: I've changed this code a bit now.

src/lm.jl
mkborregaard: I've tried a slightly different tack now with a local function, see if you think that would work.

src/lm.jl
```diff
@@ -156,7 +156,30 @@ function coeftable(mm::LinearModel)
...
+# Different methods of prediction for the different factorization types
+
+predict(mm::LinearModel, newx::AbstractMatrix) = newx * coef(mm)
+
+function predict(mm::LinearModel, newx::AbstractMatrix, interval_type::Symbol, level = 0.95)
+    prediction = newx * coef(mm)
+    interval_type == :confint || error("only :confint is currently implemented") #:predint will be implemented
+
+    R = cholfact!(mm.pp)[:U] #get the R matrix from the QR factorization
+    residvar = (ones(3,1) * deviance(mm)/dof_residual(mm))'
```
andreasnoack: The `ones(3,1)` part doesn't seem to do anything here.

src/lm.jl
```diff
@@ -156,7 +156,30 @@ function coeftable(mm::LinearModel)
...
+        Ac_ldiv_B!(R,x)
+        residvar * (x⋅x)
+    end
+
+    interval = similar(prediction)
+    for i in 1:length(interval)
+        interval[i] = _local(newx[i,:])[1]
+    end
+
+    interval = quantile(TDist(dof_residual(mm)), (1 - level)/2) * sqrt(interval)
```
andreasnoack: In the current form, this creates two temporary arrays: `sqrt(interval)` and then again for `quantile`. Hence is would be better to move these inside the loop since that would avoid the temporaries. I think the easiest way to accomplish that would be to structure the code as I suggested. In that model, there would mainly be computations on scalars and then, in a separate method, a single loop over the rows of the `X`s. Another reason to favor that solution is that it exposes a non-allocating prediction method which some users might wat to make use of. That would be possible in the current model.

mkborregaard: I'll try to do something like what you suggested - the reason I didn't do it just as you wrote it is that I didn't want to calculate `tvalue`, `residvar`, `prediction` and `R` repeatedly inside a loop (and the suggestion was to pass the inner version to  a loop, right)? So I defined a local function inside the scope that had access to these variables. I can also pass them explicitly to a `predict(x::StridedVector,...)` (I am doing that now) but I don't see how a user would use that?

mkborregaard: Another reason for having the `sqrt` outside the loop is that if I want to create a `predint` instead (a prediction interval) i need to multiply by the predictive variance before taking the square root. I can put a conditional in the inner function as well, of course.

src/lm.jl
```diff
@@ -156,7 +156,30 @@ function coeftable(mm::LinearModel)
               ["x$i" for i = 1:size(mm.pp.X, 2)], 4)
 end
 
-predict(mm::LinearModel, newx::Matrix) =  newx * coef(mm)
+# Different methods of prediction for the different factorization types
+
+predict(mm::LinearModel, newx::AbstractMatrix) = newx * coef(mm)
+
+function predict(mm::LinearModel, newx::AbstractMatrix, interval_type::Symbol, level = 0.95)
```
nalimilan: Could you add a docstring to document what `interval_types` values are possible for `LinearModel` and what they mean?

It would also be great to add a generic docstring for `predict` [in StatsBase](https://github.com/JuliaStats/StatsBase.jl/blob/af2c1279b55c179bc3130203d1bf3614ca23ccf4/src/statmodels.jl#L153), to standardize on the order of arguments. The description of `newx`, which is common to all model types, should live there.

nalimilan:
Also needs tests.

src/lm.jl
mkborregaard: I don't think I am doing it right (`predict_old` is the current, `predict` my attempt at your suggestion):
```julia
predict(mm::LinearModel, newx::AbstractMatrix) = newx * coef(mm)

function predict(mm::LinearModel, x::StridedVector, tvalue, residvar, R, interval_type)
    Ac_ldiv_B!(R,x)
    tmp = residvar * (x⋅x) # ready for modifying if I need a prediction interval
    tvalue * sqrt(tmp)[1]
end

function predict(mm::LinearModel, newx::AbstractMatrix, interval_type::Symbol, level = 0.95)
    prediction = newx * coef(mm)
    interval_type == :confint || error("only :confint is currently implemented") #:predint will be implemented

    R = cholfact!(mm.pp)[:U] #get the R matrix from the QR factorization
    residvar = (ones(size(newx,2),1) * deviance(mm)/dof_residual(mm))'
    tvalue = quantile(TDist(dof_residual(mm)), (1 - level)/2)
    interval = similar(prediction)

    for i in 1:length(interval)
        interval[i] = predict(mm, newx[i,:], tvalue, residvar, R, interval_type)
    end

    prediction, prediction .+ interval, prediction .- interval
end

function predict_old(mm::LinearModel, newx::AbstractMatrix, interval_type::Symbol, level = 0.95)
    prediction = newx * coef(mm)
    interval_type == :confint || error("only :confint is currently implemented") #:predint will be implemented

    R = cholfact!(mm.pp)[:U] #get the R matrix from the QR factorization
    residvar = (ones(size(newx,2),1) * deviance(mm)/dof_residual(mm))'


    function _local(x::StridedVector)
        Ac_ldiv_B!(R,x)
        residvar * (x⋅x)
    end

    interval2 = similar(prediction)
    for i in 1:length(interval2)
        interval2[i] = _local(newx[i,:])[1]
    end

    interval2 = quantile(TDist(dof_residual(mm)), (1 - level)/2) * sqrt(interval2)
    prediction, prediction .+ interval2, prediction .- interval2
end
```
skaermbillede 2016-12-16 kl 13 39 00

mkborregaard:
Absolutely

src/lm.jl
mkborregaard: Here is a version very close to what you suggested, I think, but that is slower:
```julia
function predict2(mm::LinearModel, x::StridedVector, interval_type::Symbol, level = 0.95)
    prediction = x' * coef(mm)
    interval_type == :confint || error("only :confint is currently implemented") #:predint will be implemented

    R = cholfact!(mm.pp)[:U] #get the R matrix from the QR factorization
    residvar = deviance(mm)/dof_residual(mm)
    tvalue = quantile(TDist(dof_residual(mm)), (1 - level)/2)

    Ac_ldiv_B!(R,x)
    tmp = residvar * (x⋅x) # ready for modifying if I need a prediction interval
    interval = tvalue * sqrt(tmp)

    prediction[1], (prediction .+ interval)[1], (prediction .- interval)[1]
end

function predict2(mm::LinearModel, newx::AbstractMatrix, interval_type::Symbol, level = 0.95)
    y = similar(newx[:,1])
    yl, yh = similar(y), similar(y)

    for i in 1:size(y,1)
        y[i], yh[i], yl[i] = predict2(mm, newx[i,:], interval_type, level)
    end

    y, yh, yl
end
```
skaermbillede 2016-12-16 kl 14 55 05

mkborregaard: Right, now I tested essentially the version you suggest under 1 and the one that uses the dot product. I also for comparison tested the one I originally wrote (except that my `getR` function is implemented with `cholfact!` as suggested), which is half the number of lines, and in my mind much more intuitive (see below). 
I think I have spent enough time now playing around with the implementation. If you think it works now I would be willing to write up the tests, docstrings, prediction intervals and take a stab at the GLM code, when I find some time again.
```julia
function predict_suggested(mm::LinearModel, newx::AbstractMatrix, interval_type::Symbol, level = 0.95)

    interval_type == :confint || error("only :confint is currently implemented") #:predint will be implemented

    R = cholfact!(mm.pp)[:U] #get the R matrix from the QR factorization
    residvar = (ones(size(newx,2),1) * deviance(mm)/dof_residual(mm))'

    function _local(x::StridedVector)
        pred = x' * coef(mm)
        Ac_ldiv_B!(R,x)
        pred[1], (residvar * (x⋅x))[1]
    end

    n = size(newx, 1)
    retmean, retvariance = similar(newx, n), similar(newx, n)
    for i in 1:length(retmean)
        retmean[i], retvariance[i] = _local(newx[i,:])
    end

    interval = quantile(TDist(dof_residual(mm)), (1 - level)/2) * sqrt(retvariance)
    retmean, retmean .+ interval, retmean .- interval
end

function predict_suggested_with_dot(mm::LinearModel, newx::AbstractMatrix, interval_type::Symbol, level = 0.95)
    interval_type == :confint || error("only :confint is currently implemented") #:predint will be implemented

    R = cholfact!(mm.pp)[:U] #get the R matrix from the QR factorization
    residvar = (ones(size(newx,2),1) * deviance(mm)/dof_residual(mm))'

    function _local(x::StridedVector)
        pred = x ⋅ coef(mm)
        Ac_ldiv_B!(R,x)
        pred[1], (residvar * (x⋅x))[1]
    end

    n = size(newx, 1)
    retmean, retvariance = similar(newx, n), similar(newx, n)
    for i in 1:length(retmean)
        retmean[i], retvariance[i] = _local(newx[i,:])
    end

    interval = quantile(TDist(dof_residual(mm)), (1 - level)/2) * sqrt(retvariance)
    retmean, retmean .+ interval, retmean .- interval
end

function predict_original_PR(mm::LinearModel, newx::AbstractMatrix, interval_type::Symbol, level = 0.95)
    retmean = newx * coef(mm)
    interval_type == :confint || error("only :confint is currently implemented") #:predint will be implemented

    R = cholfact!(mm.pp)[:U] #get the R matrix from the QR factorization
    residvar = (ones(size(newx,2),1) * deviance(mm)/dof_residual(mm))
    tmp = (newx/R).^2
    retvariance = tmp * residvar #(newx/R).^2

    interval = quantile(TDist(dof_residual(mm)), (1 - level)/2) * sqrt(retvariance)
    retmean, retmean .+ interval, retmean .- interval
end
```
skaermbillede 2016-12-16 kl 22 19 59

src/lm.jl
```diff
@@ -156,7 +156,20 @@ function coeftable(mm::LinearModel)
...
+    retmean = newx * coef(mm)
+    interval_type == :confint || error("only :confint is currently implemented") #:predint will be implemented
+
+    R = cholfact!(mm.pp)[:U] #get the R matrix from the QR factorization
+    residvar = (ones(size(newx,2),1) * deviance(mm)/dof_residual(mm))
+    retvariance = (newx/R).^2 * residvar
+
+    interval = quantile(TDist(dof_residual(mm)), (1 - level)/2) * sqrt(retvariance)
```
andreasnoack: `(1 - level)/2` -> `1 - level/2`
`sqrt(retvariance)` -> `sqrt.(retvariance)`, will give deprecation warning on 0.6 otherwise

mkborregaard: np. For the `level` thing, we could make it a keyword argument and also supply `alpha = 1-level`, that would make it clear and allow users use the way they are most familiar with? @ararslan 

About the `sqrt.` I had that in the code earlier but it produced an error. I must have done something else wrong back then, as it works now :-)

ararslan: Hm, I'm not entirely sure how I feel about that. On the one hand it's quite convenient for users, but on the other hand it breeds inconsistency in code. But maybe it's okay if they're both kwargs and we ensure they agree?

mkborregaard: I have no particular inclination, will leave it to you :-)

ararslan: Uhhhh, ¯\\\_(ツ)\_/¯ What do you think, @andreasnoack and @nalimilan (and anyone else reading this)?

src/lm.jl
```diff
@@ -156,7 +156,32 @@ function coeftable(mm::LinearModel)
...
+For linear models, specifying `interval_type` will return a 3-tuple with the
+predicted values, the lower and the higher confidence bound. Confidence intervals
+delimit the uncertainty of the estimate of the predicted values, prediction
+intervals delimit the estimated bounds for any new data points from the
+population.
+
+# Arguments
+* `interval_type::Symbol`: one of `:confint` or `:predint`
+* `level=0.95`: the level of the confidence interval
+* `alpha=1-level`: an alternative way of specifying the level of the interval
```
nalimilan: I think providing two arguments for the same thing is actually more confusing than useful. Computing `level` from `alpha` is really simple anyway.

Then better document arguments directly in the description paragraph since they are only two. They could very naturally be inserted into existing sentences.

src/lm.jl
```diff
@@ -156,7 +156,32 @@ function coeftable(mm::LinearModel)
               ["x$i" for i = 1:size(mm.pp.X, 2)], 4)
 end
 
-predict(mm::LinearModel, newx::Matrix) =  newx * coef(mm)
+predict(mm::LinearModel, newx::AbstractMatrix) = newx * coef(mm)
+
+"""
+For linear models, specifying `interval_type` will return a 3-tuple with the
```
nalimilan: Please add the signature first (with a four-space indent and a line break after it).

src/lm.jl
```diff
@@ -156,7 +156,32 @@ function coeftable(mm::LinearModel)
...
+function predict(mm::LinearModel, newx::AbstractMatrix, interval_type::Symbol; level = 0.95, alpha = 1 - level)
+    retmean = newx * coef(mm)
+    interval_type == :confint || error("only :confint is currently implemented") #:predint will be implemented
+
+    R = cholfact!(mm.pp)[:U] #get the R matrix from the QR factorization
+    residvar = (ones(size(newx,2),1) * deviance(mm)/dof_residual(mm))
+    retvariance = (newx/R).^2 * residvar
+
+    interval = quantile(TDist(dof_residual(mm)), alpha/2) * sqrt.(retvariance)
+    retmean, retmean .+ interval, retmean .- interval
```
nalimilan: Why not return a matrix instead of a tuple of vectors?

test/runtests.jl
```diff
@@ -306,6 +306,12 @@ end
 
     newd = convert(DataFrame, newX)
     predict(gm13, newd)
+
+    Ylm = X * [0.8, 1.6] + 0.8randn(10)
+    mm = fit(LinearModel, X, Ylm)
+    pm, lo, hi = predict(mm, newX, :confint)
+    @test isapprox(lo[5], 0.8543142404183606)
```
nalimilan: I would test all returned values, just to be sure.

src/lm.jl
mkborregaard: No reason in particular, the first version in this PR actually did that. @andreasnoack gave an example that returned a tuple of vectors at some point so I just changed it. Do you prefer a matrix?

mkborregaard:
Note that one thing the PR does not do yet is to take into account weights for a weighted prediction.

test/runtests.jl
mkborregaard: Done

src/lm.jl
mkborregaard: Again, this is not the advice from the Julia documentation, but fixed as suggested.

src/lm.jl
nalimilan: It doesn't: on the contrary, it says "Only provide an argument list when really necessary.". :-)

src/lm.jl
nalimilan: Where do you see the docs saying the contrary?

src/lm.jl
mkborregaard: Ha ha! - Sorry, I see now it says "When a function accepts many keyword arguments", I read it as "When a function accepts **any** keyword arguments". Thanks for clarifying!

src/lm.jl
nalimilan: I find a matrix more natural, since it allows you to extract a row easily. As Andreas noted, if you want only one prediction, the function could avoid allocating by returning a tuple of scalars. But I think this should be a different method which would dispatch on `newX::AbstractVector` instead. It could be added if it turns out to be useful.

README.md
```diff
@@ -72,6 +72,15 @@ julia> predict(OLS)
...
  6.83333
 
+# Prediction with new data and confidence values
+ julia> newX = DataFrame(X=[2,3,4]);
+ julia> predict(OLS, newX, :confint)
+ 3×3 Array{Float64,2}:
+  4.33333  1.33845   7.32821
+  6.83333  2.09801  11.5687
+  9.33333  1.40962  17.257  
+# The columns of the matrix are prediction, lower and upper confidence interval
```
nalimilan: "95% lower and upper confidence bounds"

nalimilan:
> Note that one thing the PR does not do yet is to take into account weights for prediction from weighted regressions.

Is that something you intend to support? If not, is a clear error raised in that case?

src/lm.jl
```diff
@@ -156,7 +156,28 @@ function coeftable(mm::LinearModel)
...
+
+"""
+    predict(mm::LinearModel, newx::AbstractMatrix, interval_type::Symbol, level = 0.95)
+    
+Specifying `interval_type` will return a 3-column matrix with the prediction and
+the low and high confidence bounds for a given `level` (0.95 equates alpha = 0.05).
+Valid values of `interval_type` are `:confint` delimiting the  uncertainty of the
+predicted relationship, and `:predint` delimiting estimated bounds for new data points.
+"""
+function predict(mm::LinearModel, newx::AbstractMatrix, interval_type::Symbol, level = 0.95)
```
nalimilan: `level::Real`

src/lm.jl
```diff
@@ -156,7 +156,28 @@ function coeftable(mm::LinearModel)
...
 end
 
-predict(mm::LinearModel, newx::Matrix) =  newx * coef(mm)
+predict(mm::LinearModel, newx::AbstractMatrix) = newx * coef(mm)
+
+"""
+    predict(mm::LinearModel, newx::AbstractMatrix, interval_type::Symbol, level = 0.95)
+    
+Specifying `interval_type` will return a 3-column matrix with the prediction and
+the low and high confidence bounds for a given `level` (0.95 equates alpha = 0.05).
```
nalimilan: "lower and upper" sounds like the conventional terms for this.

mkborregaard:
> Is that something you intend to support?

It should, shouldn't it? Not going to happen this weekend, though. Right now it raises no error.

src/lm.jl
mkborregaard: fixed

src/lm.jl
mkborregaard: fixed

nalimilan:
I just meant that if support for weights is not added in this PR when it's merged, an error should be raised, rather than returning incorrect results.

src/lm.jl
nalimilan: OK, thanks, that example needs fixing. Indeed `?*` shows you does not even match what happens in Base.

mkborregaard:
I agree completely - it was to highlight this that I made the comment, so it wouldn't get merged accidentally without this.

src/lm.jl
nalimilan: I'd personally prefer changing HypothesisTests, whose docs even say "with coverage 1-`alpha`", implying that it's more natural to talk about `level`. But we could do a small survey of other software to check what's more common.

mkborregaard: in R: `confint(object, parm, level = 0.95, ...)`
in Matlab: `confint(fitresult,level)`, where the default value of `level` is 0.95
in Python (scipy): `stats.norm.interval(alpha, loc=0, scale=1)`, where alpha is confusingly named, but should have the value 0.95 to give the 95% confidence interval

mkborregaard:
I'll try to add support for weights, prediction intervals, and GLM before asking that this be merged. That is likely to be after the basic infrastructure is moved to StatModels - then I or anyone else can make a related PR on StatsModels at the time of merging. 
Just a question - am I allowed to use RCall in the tests section, or is that 'cheating'? That would make it easy to try out a number of different data and model structures quite easily.

nalimilan:
> Just a question - am I allowed to use RCall in the tests section, or is that 'cheating'? That would make it easy to try out a number of different data and model structures quite easily.

I think you can simply check that the results are consistent with R and save the literal values in the tests. That will make it easier to run the tests than installing R separately.

src/lm.jl
simonster: At the moment, HypothesisTests is not even consistent with the `StatsBase` documentation. So perhaps we should change HypothesisTests. Of course, the difficulty is that this is a positional argument. But maybe we can do something based on the assumption that few people use confidence intervals with coverage <0.5.

andreasnoack:
> and GLM before asking that this be merged

Feel free to do that but I think it would be fine to merge tje `LinearModel` part and then extend to GLMs in a separate PR.

andreasnoack:
What is the status here? Would you still like to add GLM support or would it be an idea to wrap this up and extend the functionality in a new PR?

mkborregaard:
Sorry for be unresponsive, I have been away from keyboard. I think it is a good idea to wrap up - also to make it easier for others to build on. Will add the exception on weights hopefully today, and then signal it should be ready for merge.

mkborregaard:
I have added an error on predicting from regressions with weights, and reviewed the conversation here. IMO this is ready to merge. 

A question: Should the docstring here https://github.com/JuliaStats/StatsBase.jl/pull/227 then be merged as well? There was a bit of discussion about the specificity of newX etc. What would talk against is that the predict-form where newX is a DataFrame is not yet implemented, but that is the one most likely to be used by users.

mkborregaard:
In my view, this PR is ready to merge.

andreasnoack:
Thanks for bumping and for all the work.

PR #173, opened 2017-01-31
Stop if deviance is zero

### LINKED ISSUE #117: GLM breaks when data is exactly linear
Take any exactly linear data and feed it to GLM
(arguably not the right function when you know the data is linear)

```
x=[1,2,3,4]
y=[24,34,44,54]
data=DataFrame(X=x,Y=y)
```

```
OLS = glm(Y~X, data, Normal(), IdentityLink())
```

anyway one gets the complaint

```
failure to converge in 30 iterations
```

which seems unfortunate.

andreasnoack:
Fixes #117

### DIFF
--- a/src/glmfit.jl
+++ b/src/glmfit.jl
@@ -233,10 +233,11 @@ function _fit!(m::AbstractGLM, verbose::Bool, maxIter::Integer, minStepFac::Real
         installbeta!(p, f)
         crit = (devold - dev)/dev
         verbose && println("$i: $dev, $crit")
-        if crit < convTol
+        if crit < convTol || dev == 0
             cvg = true
             break
         end
+        @assert isfinite(crit)
         devold = dev
     end
     cvg || throw(ConvergenceException(maxIter))
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -321,7 +321,6 @@ end
     @test isapprox(pred[3,3], 3.6617479283005894)
     @test isapprox(pred[4,3], 0.6477623101170038)
     @test isapprox(pred[5,3], 2.564532433982956)
-    
 end
 
 @testset "Issue 118" begin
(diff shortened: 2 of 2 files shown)


src/glmfit.jl
```diff
@@ -233,10 +233,13 @@ function _fit!(m::AbstractGLM, verbose::Bool, maxIter::Integer, minStepFac::Real
...
         crit = (devold - dev)/dev
         verbose && println("$i: $dev, $crit")
-        if crit < convTol
+        if crit < convTol || dev == 0
             cvg = true
             break
         end
+        if !isfinite(crit)
+            error("convergence criterion is $crit") # maybe change ConvergenceException to allow for custom error
+        end
```
ararslan: Using a `StatsBase.ConvergenceException` sounds like a good idea here (this is the kind of thing it was made for), but unfortunately that's only set up to capture the number of iterations rather than the convergence criterion, so the error message might not be as informative...

nalimilan: We should add a second field to that exception then (with backward compatibility via a default value).

ararslan: Or we could just change the internal field to a string and let people write their own messages, akin to base exception types.

nalimilan: Yeah, but it's not great if you need to automate the check to find out why convergence failed. Another possibly useful extension at some point would be to return the values of coefficients (this is e.g. what the gnm R package does and that's useful to find out which variables are problematic).

ararslan: That seems useful, though that could similarly be accomplished by just putting the coefficients in the error string, e.g.
```julia
throw(ConvergeException("convergence failed with criterion $crit, coefficients were $(join(coefs, " "))"))
```

andreasnoack: I think I've changed my mind on this particular error. I don't think it should ever happen that we get `NaN`s or `Inf`s here so this should probably just be an `@assert`. The `ConvergeException` should be thrown for cases that could possibly happen for certain datasets.

nalimilan: @ararslan That's not really practical when you have 100 or more coefficients, which has happened to me many times (with log-mutiplicative models). Anyway it's always better to be able to access the vector object rather than just a string (with precision loss, etc.).

But indeed if that can't happen then there's no hurry to add this (though in general adding the criterion to `ConvergenceException` wouldn't be a bad idea).

andreasnoack:
Any additional comments?

PR #177, opened 2017-03-22
Updates for Julia 0.6

ararslan:


src/GLM.jl
```diff
@@ -54,14 +55,14 @@ module GLM
         updateμ!,      # update the response type from the linear predictor
         wrkresp         # working response
 
-    typealias FP AbstractFloat
-    typealias FPVector{T<:FP} AbstractArray{T,1}
+    const FP = AbstractFloat
+    @compat const FPVector{T<:FP} = AbstractArray{T,1}
```
nalimilan: Technically I think `const` isn't needed, but that's so subtle that I don't think we care.

src/linpred.jl
```diff
@@ -15,20 +15,19 @@ function installbeta!(p::LinPred, f::Real=1.)
...
-typealias BlasReal Union{Float32,Float64}
-
 type DensePredQR{T<:BlasReal} <: DensePred
     X::Matrix{T}                  # model matrix
     beta0::Vector{T}              # base coefficient vector
     delbeta::Vector{T}            # coefficient increment
     scratchbeta::Vector{T}
     qr::QRCompactWY{T}
-    function DensePredQR(X::Matrix{T}, beta0::Vector{T})
+    @compat function (::Type{DensePredQR{T}}){T}(X::Matrix{T}, beta0::Vector{T})
```
nalimilan: Is `@compat` really needed? Same below.

andreasnoack: I don't think it is

nalimilan: Then please remove it before merging, these annotations are annoying. :-)

ararslan: I thought it was needed for the type parameter on `new`?

src/GLM.jl
ararslan: I think it's the preferred style, as it's what's suggested by the deprecation warning and is what's used in Base.

src/linpred.jl
ararslan: Turns out I was wrong--not needed for `new{T}` on 0.5. ¯\\\_(ツ)\_/¯

Issue #181, opened 2017-05-17
Implement ANOVAs for lms


LewisHein:
This is an attempt to revive #70 and resolve the issues around it. The purpose of this issue is to decide exactly how we should resolve some people's reservations about the original implementation.

If I understand the preceeding in issue #65 and PR #70 discussion correctly, ANOVA tables are objectionable because they give false impressions, and we want a truly general way of comparing two models, null and alternative based on their F statistics. I have a couple issues with this previous discussion.

1) @dmbates Your comments opened my eyes to a new way of thinking about ANOVA models. I think it would be great if Julia could finally have an implementation of ANOVAs that is general and done right, but let's not lose perspective. There truly is a place in this world for the default alternative hypothesis of "The response differs in at least one possible case", because that's the one we want 99.9% of the time.

1) I feel like the discussion went down a rabbit hole, big-time. As mentioned above, It was a fascinating, useful, and productive rabbit hole, but I don't think rabbits are our highest priority right now. Again, as above, it would be great to make Julia have a general, well-done implementation of ANOVA, BUT, and this is a really big BUT, we've got to have the simple interface that everyone is used to thinking about. We can call the shots to some extent for how people do things in Julia, but the way people are used to thinking about ANOVA (as a test of a null hypthothesis that the response variable is the same across all possible values of predictors) isn't going to go away anytime soon. It is perfectly correct, as far as it goes, but not very general. My fear is that if we take users coming from this paradigm and suddenly drop them into the fully general case, they won't know how to formulate the special case they are used to. A few will push ahead and figure it out, but many will throw up their hands in frustration and walk back to R or SPSS. This would be a shame.

@dmbates You and most of the other people who write GLM.jl are statisticians and know statistical theory inside and out. I am not. I am a mathematically-minded scientist who is more stubborn about trying to figure things out than some others I know. You guys, with your intimate knowledge of stats theory, are making GLM.jl one of the best stats packages out there. But again, none of us represent the majority of the intended end users. These include many people who just had a few stats classes for their degrees and were never taught these details we are discussing. If we don't give them a familiar interface, we will scare them away, and be left with only professional statisticians and extra-stubborn scientists in our user base. Everyone else will go back to R and Julia won't have conquered the world.

[...]

PR #182, opened 2017-05-17
Add F test for comparing two models

LewisHein:
Here is a rough first draft of what I'm thinking for the F-test for model comparison. If you all think I am on the right track, I'll write up some docs. The addition to the test suite tests a single ANOVA run against the p-value from R's implementation. This is for addressing #181


nalimilan:
Thanks. I think the function should return a table similar to `CoefTable` rather than a list of objects when more than two models are passed. That would be more convenient for printing as well as to extract values.

src/ftest.jl
```diff
@@ -0,0 +1,75 @@
+type FTestResult
+    mod1::RegressionModel
```
nalimilan: Should probably not store the original models, that can take a lot of space and can be tracked separately if needed.

src/ftest.jl
```diff
@@ -0,0 +1,75 @@
...
+    end
+end
+
+function show(io::IO, x::MultiFTestResult)
+    for res in x.results
+	show(io, res)
+    end
+end
+
+export ftest
```
nalimilan: Exports should go to StatsBase.jl.

src/ftest.jl
```diff
@@ -0,0 +1,75 @@
...
+    fstat::Float64 # f statistic
+    pval::Float64 # p value
+end
+
+type MultiFTestResult
+    results::Array{FTestResult, 1}
+end
+
+function ftest(mod1::RegressionModel, mod2::RegressionModel)
+    SSR1 = deviance(mod1.model.rr)
```
nalimilan: Should call `deviance` on `mod1` to be fully generic.

src/ftest.jl
```diff
@@ -0,0 +1,75 @@
...
+    SSR2 = deviance(mod2.model.rr)
+
+    nparams1 = length(mod1.model.pp.beta0)
```
nalimilan: Use `dof(mod1)`.

src/ftest.jl
```diff
@@ -0,0 +1,75 @@
...
+    MSR2::Float64 # mean of squared residuals, model 2
+    fstat::Float64 # f statistic
+    pval::Float64 # p value
+end
+
+type MultiFTestResult
+    results::Array{FTestResult, 1}
+end
+
+function ftest(mod1::RegressionModel, mod2::RegressionModel)
```
nalimilan: The F test is only valid for linear regressions, so the signature should be stricter.

src/ftest.jl
```diff
@@ -0,0 +1,75 @@
...
+
+    fstat = MSR1/MSR2
+    pval = ccdf(FDist(df2, df1), fstat)
+
+    return FTestResult(mod1, mod2, SSR1, SSR2, df1, df2, MSR1, MSR2, fstat, pval)
+end
+
+function ftest(mods::RegressionModel...)
+    nmodels = length(mods)
+    results = Array{FTestResult, 1}((nmodels^2)-nmodels)
```
nalimilan: What R's `anova` and Stata's `nestreg` do is to compare each model to the previous one. Computing all possible comparisons quickly gets very messy.

nalimilan:
It would also be nice to check that the models were fitted on the same data, at least by checking that the response vectors are equal (and maybe even that all common variables are equal).

src/ftest.jl
LewisHein: Not sure what you mean here -- I can't export ftest until I've defined it.

ararslan: Sure you can.

ararslan: Bindings in `export` statements aren't resolved immediately. As an example,
```julia
julia> baremodule Test
       export A # has no definition
       end
Test

julia> using Test # doesn't error even though A is exported and undefined

julia> Test.A
ERROR: UndefVarError: A not defined
```
Julia doesn't care until you try to call it.

LewisHein: Yup, you're right. It appears that I made a typo when I tried it the first time

LewisHein:
> It would also be nice to check that the models were fitted on the same data, at least by checking that the response vectors are equal (and maybe even that all common variables are equal).

This would be great; however, I am not sure how to get the common variables of the models. I am pretty new to the internals of GLM.jl, so please forgive my ignorance.

src/GLM.jl
```diff
@@ -53,7 +53,8 @@ module GLM
         nobs,           # total number of observations
         predict,        # make predictions
         updateμ!,      # update the response type from the linear predictor
-        wrkresp         # working response
+        wrkresp,        # working response
+	ftest		# compare two models with an F test
```
ararslan: Please indent with spaces instead of tabs

src/ftest.jl
```diff
@@ -31,23 +29,33 @@ function ftest(mod1::RegressionModel, mod2::RegressionModel)
...
+    for (i, mod1) in enumerate(mods[2:end])
+	    result = ftest(mods[i], mod1)
+	    SSR1s[i] = result.SSR1
+	    SSR2s[i] = result.SSR2
+	    df1s[i] = result.df1
+	    df2s[i] = result.df2
+	    MSR1s[i] = result.MSR1
+	    MSR2s[i] = result.MSR2
+	    fstats[i] = result.fstat
+	    pvals[i] = result.pval
```
ararslan: Same for these. Indents should be 4 spaces each.

src/ftest.jl
```diff
@@ -31,23 +29,33 @@ function ftest(mod1::RegressionModel, mod2::RegressionModel)
...
     end
 
-    return MultiFTestResult(results)
+    return CoefTable([SSR1s, SSR2s, df1s, df2s, MSR1s, MSR2s, fstats, pvals], ["Model 1 SSR", "Model 2 SSR", "Model 1 df", "Model 2 df", "Model 1 MSR", "Model 2 MSR", "F statistic", "p-value"], ["Model $(i-1):$i" for i in 2:nmodels])
```
ararslan: It would be good to break this long line into multiple lines for readability.

LewisHein:
Also, switching to LinPredModel has caused my tests to fail, with an error message about having no method matching `ftest(::DataFrames.DataFrameRegressionModel{GLM.LinearModel{GLM.LmResp{Array{Float64,1}},GLM.DensePredQR{Float64}},Array{Float64,2}}, ::DataFrames.DataFrameRegressionModel{GLM.LinearModel{GLM.LmResp{Array{Float64,1}},GLM.DensePredQR{Float64}},Array{Float64,2}})`.

How do I fix this?

src/ftest.jl
```diff
@@ -0,0 +1,84 @@
...
+        SSR2s[i] = result.SSR2
+        df1s[i] = result.df1
+        df2s[i] = result.df2
+        MSR1s[i] = result.MSR1
+        MSR2s[i] = result.MSR2
+        fstats[i] = result.fstat
+        pvals[i] = result.pval
+    end
+
+    return CoefTable([SSR1s, SSR2s, df1s, df2s, MSR1s, MSR2s, fstats, pvals],
```
nalimilan: Sorry if I wasn't clear, I didn't mean to (ab)use `CoefTable` for this, but to take inspiration from it to define the `MultiFTestResult` type (could be called `FTestTable` BTW). That type would have the same fields as `FTestResult`, but those would be vectors rather than scalars (one column for each info).

I'm not even sure we need `FTestResult` for the special case of a two-model comparison. Opinions?

LewisHein: This isn't an opinion, but rather an explanation. I defined FTestResult purely so that I could define show() for it and have the result nicely printed.

Is there some more standard way to do this?

nalimilan:
> This would be great; however, I am not sure how to get the common variables of the models. I am pretty new to the internals of GLM.jl, so please forgive my ignorance.

Unfortunately, you can't get the names of the coefficients at this level, they are only available for `DataFrameRegressionModel` objects (or other models with a model frame). We could add a specific method which would check this later in DataFrames/StatsModels, but that's beyond the scope of this PR.

What can be done in GLM.jl is to check that the fields of the `LmResp` object are equal, which should catch most problems.

LewisHein:
Well, this morning with a fresh perspective, I tackled the submodel test. Tell me what you think of my solution.

src/ftest.jl
nalimilan: Not totally "standard", but `CoefTable` (in StatsBase) is a very similar case (lightweight structure to be used where R would use a data frame, which is overkill in Julia).

LewisHein:
Where?

nalimilan:
> Where?

Funny, GitHub lost the context. I referred the the `break` line inside the loop.

LewisHein:
No, not by my reckoning. That `break` is reached if and only if the column pred1[:, i] matched pred2[:, j] This means that variable i in model 1 is equal to variable j in model 2, and thus that there exists a variable in model 1 that is present in model 2.

But we can't return here. If we did, then we would return true every time model 1 contained at least one variable that was in model 2 -- Not the definition of nesting!

Have I made myself clear? Does some of this need to go into comments in the code?

nalimilan:
Got it. But then you should call `continue` rather than `break`. You also don't need `var_in_mod2`: if you reach the line after the inner loop, it means the column wasn't found and you can return false:
```julia
    @inbounds for i in 1:npreds1
        for j in 1:npreds2
            view(pred1, :, i) == view(pred2, :, j) && continue
        end
        # We have found a predictor variable in model 1 that is not in model 2
        return false 
    end
```

This also should be tested on a few error cases.

LewisHein:
I'm confused. Does this `continue` (or `break`) affect the inner `for` loop or the outer? According to my tests, it only affects the inner loop. If I am correct about this, the inner loop will terminate after running npreds2 times, and execution will proceed to `return false` in all cases.

LewisHein:
Here was my test:
```julia
    for i in 1:3
        info("i=",i)
        for j in 1:4
            info("j=",j)
            continue
        end
    end
```

With output:

```
INFO: i=1
INFO: j=1
INFO: j=2
INFO: j=3
INFO: j=4
INFO: i=2
INFO: j=1
INFO: j=2
INFO: j=3
INFO: j=4
INFO: i=3
INFO: j=1
INFO: j=2
INFO: j=3
INFO: j=4
```

nalimilan:
My bad. I should stop doing two things at the same time. Currently one needs to use `@goto` for this, which isn't simpler than using a boolean flag. Carry on.

My suggestion to use `view` still holds though.

src/ftest.jl
```diff
@@ -24,10 +24,10 @@ function issubmodel(mod1::LinPredModel, mod2::LinPredModel)
     # If model 1 has more predictors, it can't possibly be a submodel
     npreds1 > npreds2 && return false 
     
-    for i in 1:npreds1
+    @inbounds for i in 1:npreds1
         var_in_mod2 = false
-        for j in 1:npreds2
-            if pred1[:, i] == pred2[:, j]
+        @inbounds for j in 1:npreds2
```
nalimilan: Actually one `@inbounds` is enough for the whole block.

src/ftest.jl
```diff
@@ -51,15 +60,15 @@ function ftest(mod1::LinPredModel, mod2::LinPredModel)
...
     MSR2 = SSR1/df1
 
     fstat = MSR1/MSR2
     pval = ccdf(FDist(df2, df1), fstat)
 
-    return FTestResult(SSR1, SSR2, df1, df2, MSR1, MSR2, fstat, pval)
+    return FTestTable(SSR1, SSR2, df1, df2, MSR1, MSR2, fstat, pval)
 end
 
 function ftest(mods::LinPredModel...)
```
nalimilan: The methods should be defined the other way around: you don't even need to have a special method for the two-argument case, it's just a special case of this one.

src/ftest.jl
```diff
@@ -85,29 +94,25 @@ function ftest(mods::LinPredModel...)
         pvals[i] = result.pval
     end
 
-    return CoefTable([SSR1s, SSR2s, df1s, df2s, MSR1s, MSR2s, fstats, pvals],
-    ["Model 1 SSR", "Model 2 SSR", "Model 1 df", "Model 2 df", "Model 1 MSR", "Model 2 MSR", "F statistic", "p-value"],
-    ["Model $(i-1):$i" for i in 2:nmodels])
+    return FTestTable(SSR1s, SSR2s, df1s, df2s, MSR1s, MSR2s, fstats, pvals)
```
nalimilan: Showing the values for "Model 1" and "Model 2" isn't ideal in a table where there are more than two models. It would be better to have one line for each model, with a column giving the difference in SSR, df and MSR between the two. Do you see what kind of R output I'm taking as a reference here?

src/ftest.jl
```diff
@@ -85,29 +94,25 @@ function ftest(mods::LinPredModel...)
...
 end
 
 ### Utility functions to show FTestResult and MultiFTestResult
-function show(io::IO, x::FTestResult)
+function show(io::IO, x::FTestTable)
```
nalimilan: Actually, looking at the equivalent code for `CoefTable`, I realize it's a bit complex to print this kind of tables correctly. So maybe after all maybe we should reuse `CoefTable` here? If so we should rename it later to make it more general.

src/ftest.jl
LewisHein: Yes, the two-argument case is a special case of the multi-argument case, and not strictly necessary. However, I use it as a helper function to make the multi-argument code more readable.

LewisHein:
I think I have reached a compromise about how to print the FTestTable. 

```julia
julia> d = DataFrame(Treatment=[1, 1, 1, 2, 2, 2, 1, 1, 1, 2, 2, 2.], Result=[1.1, 1.2, 1, 2.2, 1.9, 2, .9, 1, 1, 2.2, 2, 2], Other=[1, 1, 2, 1, 2, 1, 3, 1, 1, 2, 2, 1])

julia> bigmod = lm(Result~Treatment+Other, d).model
julia> mod = lm(Result~Treatment, d).model
julia> nullmod = lm(Result~1, d).model

julia> ftest(mod, nullmod)
====================================================================
Fisher 2-model F test with 10 and 1 degrees of freedom
Sums of squared residuals: 0.12833333333333344 and 3.229166666666667
F* = 241.62337662337643 p = 2.481215056713184e-8


julia> ftest(bigmod, mod, nullmod)
======================================================================
Fisher 2-model F test with 9 and 1 degrees of freedom
Sums of squared residuals: 0.10383333333333342 and 0.12833333333333344
F* = 2.1235955056179776 p = 0.17903437900958952
====================================================================
Fisher 2-model F test with 10 and 1 degrees of freedom
Sums of squared residuals: 0.12833333333333344 and 3.229166666666667
F* = 241.62337662337643 p = 2.481215056713184e-8
```

How does this look for output? I am not familiar with what R spits out, although I can look if needed

nalimilan:
I would say the output would be more readable if it was formatted like a table. This is why I realized a bit late it may be more appropriate to use `CoefTable` after all.

R's `anova` function looks [like this](https://stats.stackexchange.com/questions/53312/comparing-two-models-using-anova-function-in-r). Stata's `nestreg` looks [like this](http://www.stata.com/manuals13/rnestreg.pdf) (see at the bottom of p.3). These are just possible inspirations (in particular, you can't print model formulas from GLM.jl, this would have to be done in DataFrames/StatsModels). I wonder which of the sum of squares and the R² is the best statistic to show. I'd personally prefer the latter, but maybe there are reasons to prefer the former.

src/ftest.jl
nalimilan: Well, the code would be quite simpler if you computed each stat direclty in the most general function, e.g. using comprehensions and `diff`.

LewisHein:
Well, I looked at the output of CoefTable, R's output and Stata's output. CoefTable and R both look a little messy to me, so I made something else.

Here's what I've got; see what you think
```julia
julia> ftest(mod, nullmod)
Fisher 2-model F tests (between ajdacent pairs of models)

======================================================================================
Comparison	DF1	DF2	SSR1		SSR2		F*		p(>F)
--------------------------------------------------------------------------------------
1:2		10	1	0.128333	3.229167	241.623377	<1e-7
======================================================================================


julia> ftest(bigmod, mod, nullmod)
Fisher 2-model F tests (between ajdacent pairs of models)

======================================================================================
Comparison	DF1	DF2	SSR1		SSR2		F*		p(>F)
--------------------------------------------------------------------------------------
1:2		9	1	0.103833	0.128333	2.123596	0.1790
2:3		10	1	0.128333	3.229167	241.623377	<1e-7
======================================================================================
```

src/GLM.jl
```diff
@@ -1,12 +1,13 @@
 __precompile__()
 
 module GLM
+    using ArgCheck
```
ararslan: The uses of ArgCheck here are so simple that I don't think it's worth adding a dependency for a couple of cases. (As an aside, if we were to add a dependency it would need to go in the REQUIRE file.)

src/ftest.jl
```diff
@@ -0,0 +1,130 @@
+type FTestTable
+    ntests::Int
+    SSR1::Array{Float64, 1} # sum of squared residuals, model 1
```
ararslan: This is just a stylistic thing, but typically we try to reserve caps for types and use all lowercase for variables and fields. So this should be `ssr1`, likewise for the other fields here.

src/ftest.jl
```diff
@@ -0,0 +1,130 @@
...
+    df1::Array{Int, 1} # degrees of freedom, model 1 residuals
+    df2::Array{Int, 1} # degrees of freedom, model 2 residuals
+    MSR1::Array{Float64, 1} # mean of squared residuals, model 1
+    MSR2::Array{Float64, 1} # mean of squared residuals, model 2
+    fstat::Array{Float64, 1} # f statistic
+    pval::Array{Float64, 1} # p value
+end
+
+function FTestTable(SSR1::Array{Float64, 1}, SSR2::Array{Float64, 1}, df1::Array{Int, 1}, df2::Array{Int, 1}, MSR1::Array{Float64, 1}, MSR2::Array{Float64, 1}, fstat::Array{Float64, 1}, pval::Array{Float64, 1})
+#Note: Is checking args in internal helper functions the desired behavior?
```
ararslan: I assume you mean in type constructors, since that's what this is, in which case yes.

src/ftest.jl
```diff
@@ -0,0 +1,130 @@
...
+@argcheck length(SSR1) == length(SSR2) == length(df1) == length(df2) == length(MSR1) == length(MSR2) == length(fstat) == length(pval) "F test values  must all be the same length"
```
ararslan: Should be indented since it's inside of a function. Also we try to keep the line lengths down to 92 characters at most (which is admittedly an arbitrary cutoff, but it's what's used in Base).

src/ftest.jl
```diff
@@ -0,0 +1,130 @@
...
+#=        if x.ntests == 1
+            lines = [string("Fisher 2-model F test with ", x.df1[1], " and ", x.df2[1], " degrees of freedom"), string("Sums of squared residuals: ", round(x.SSR1[1], 6), " and ", round(x.SSR2[1], 6)), string("F* = ", round(x.fstat[1], 6), " p = ", PValue(x.pval[1]))]
+            outlen = maximum(length, lines)
+
+            println(io, RepString("=", outlen))
+            for line in lines
+                println(io, line)
+            end
+        else=#
+	    nequals = 86
```
ararslan: Stray tab here

test/runtests.jl
```diff
@@ -321,3 +321,21 @@ end
     @test isapprox(lm(Xf, y).pp.beta0, ones(2))
     @test isapprox(lm(X, yf).pp.beta0, ones(2))
 end
+
+@testset "F test for model comparison" begin
+    d = DataFrame(Treatment=[1, 1, 1, 2, 2, 2, 1, 1, 1, 2, 2, 2.], Result=[1.1, 1.2, 1, 2.2, 1.9, 2, .9, 1, 1, 2.2, 2, 2], Other=[1, 1, 2, 1, 2, 1, 3, 1, 1, 2, 2, 1])
+    mod = lm(Result~Treatment, d).model
```
ararslan: All formulas should be in a `@formula()` call. The use of `~` on its own like this was removed in Julia 0.6.

src/ftest.jl
```diff
@@ -0,0 +1,130 @@
...
+Test if mod[i] fits significantly better than mod[i+1]
+
+Note: This function can easily be used to do an ANOVA. ANOVA is nothing more than a test
+to see if one model fits the data better than another.
+
+# Examples:
+As usual, we want to compare a result across two or more treatments. In the classic ANOVA
+framework, our null hypothesis is that Result~1 is a perfectly good fit to the data. 
+The alternative for ANOVA is that Result~Treatment is a better fit to the data than Result~1
+```jldoctest
```
ararslan: This doctest would actually fail, since the lines below that don't show output are not terminated with `;`.

src/ftest.jl
```diff
@@ -0,0 +1,130 @@
...
+            return false 
+        end
+    end
+
+    return true
+end
+
+"""
+    `function ftest(mod::LinPredModel...)`
+Test if mod[i] fits significantly better than mod[i+1]
```
ararslan: No `function`, no backticks, line break after signature. Also I'd word this as, "For each pair of linear predictive models in `mod`, perform an F-test to determine whether one fits significantly better than the other."

src/ftest.jl
```diff
@@ -0,0 +1,130 @@
+type FTestTable
```
ararslan: It seems to me that we should just use `CoefTable` for this rather than defining a new type.

src/ftest.jl
```diff
@@ -0,0 +1,130 @@
...
+            lines = [string("Fisher 2-model F test with ", x.df1[1], " and ", x.df2[1], " degrees of freedom"), string("Sums of squared residuals: ", round(x.SSR1[1], 6), " and ", round(x.SSR2[1], 6)), string("F* = ", round(x.fstat[1], 6), " p = ", PValue(x.pval[1]))]
+            outlen = maximum(length, lines)
+
+            println(io, RepString("=", outlen))
+            for line in lines
+                println(io, line)
+            end
+        else=#
+	    nequals = 86
+            println(io, "Fisher 2-model F tests (between ajdacent pairs of models)")
```
ararslan: I didn't even realize that Fisher was affiliated with it; I thought it was Snedecor. I think it's most common to just call it an F test.

src/ftest.jl
```diff
@@ -0,0 +1,130 @@
+type FTestTable
+    ntests::Int
```
LewisHein: True, but I was thinking that length() on the vectors was cumbersome. I'd replace this with a function ntests(t::FTestTable), but there's no point if we're going to CoefTable

src/GLM.jl
LewisHein: OK -- I think it might be good to move to ArgCheck for the whole package, but that's definitely outside the scope of what I'm doing here

src/ftest.jl
LewisHein: OK -- Maybe in a future PR I'll improve how CoefTable is printed.

LewisHein:
> I also think it's more useful to have one row per model (including the first one), with one column giving each stat for a given model, and another one giving the change compared to the previous model.

@nalimilan and @ararslan I see your point about one model per line. However, I find myself a little at a loss for a good way to implement it. I have thought of several ways, and the best so far is still not good. This way is going to be pretty confusing to interpret and require a lot of extra code just to get the arrays right. I am thinking of something like this:

```
Model      SSR    dSSR    MSR    dMSR    fstat    pval
------------------------------------------------------------------
mod1       100            50         
mod2       85      -15    33     -17     10.24     <1e6
```

Here are my complaints (as a user) with this layout:
 1) We are implying that fstat and pval only apply to model 2, which is obviously nonsensical.  I'm sure people can figure out that it actually applies to comparing model 2 to model 1, but why not make this immediately obvious in our output?.
 2) How much do we care about the actual values of SSR, MSR, etc. for each model? I would say we might care quite a bit, **but** most of us care first about which fits the data better in the comparison. If we have one comparison per line instead of one model per line, we give people less to wade through for the first thing they want to know. Then, they can do just a little more wading to find the second thing, if they care. Given that quite a few users may care only about the p-value of the comparison with just a glance at SSR, MSR, etc, this seems like a logical optimization.

As a programmer, this layout is definitely achievable, but requires a bunch of juggling around with arrays.  What are we to do about the first element of dSSR, dMSR, fstat, and pval? CoefTable requires all columns to be the same length, so we have to fill them all with null values. NaN is really wrong, and besides, some of these (like DF) are going to be integer arrays that won't take NaN.

nalimilan:
This looks OK to me. Knowing how well a model fits the data matters for the comparison with the next one. Yes, people need to know that the p-value refers to the comparison with the previous model, but that's not the end of the world and it's quite obvious from the fact that the first model does not have a p-value. Also, other software do that so people are usually accustomed to it, or can easily find documentation.

You're right that there's a printing issue with `CoefTable` though. I think we could use `NaN` for this, and have a special case for this value in the `show` method so that it's printed as empty. That could be useful in other situations.

src/GLM.jl
ararslan: I'd prefer to keep the dependency list as light as possible, and IMO ArgCheck doesn't really provide sufficient benefit to warrant adding it as a dependency.

src/ftest.jl
ararslan: TIL the F stands for Fisher. How did I not know that...

LewisHein:
Here's what I've got -- everything, I think, except @nalimilan's suggestion of testing output

test/runtests.jl
```diff
@@ -164,178 +164,178 @@ end
...
     @test dof(gm9) == 3
-    @test isapprox(deviance(gm9), 0.16260829451739)
-    @test isapprox(loglikelihood(gm9), -26.24082810384911)
-    @test isapprox(aic(gm9), 58.48165620769822)
-    @test isapprox(aicc(gm9), 63.28165620769822)
-    @test isapprox(bic(gm9), 59.07332993970688)
-    @test isapprox(coef(gm9), [5.50322528458221, -0.60191617825971])
-    @test isapprox(GLM.dispersion(gm9.model, true), 0.02435442293561081)
-    @test isapprox(stderr(gm9), [0.19030107482720, 0.05530784660144])
+@test isapprox(deviance(gm9), 0.16260829451739)
```
ararslan: The indent shouldn't be removed

LewisHein: Don't know how that happened... Sorry.

test/runtests.jl
```diff
@@ -195,18 +195,18 @@ admit_agr = DataFrame(count = [28., 97, 93, 55, 33, 54, 28, 12],
...
-@test isapprox(deviance(gm14), 474.9667184280627)
-@test isapprox(loglikelihood(gm14), -237.48335921403134)
-@test isapprox(aic(gm14), 482.96671842822883)
-@test isapprox(aicc(gm14), 483.0679842510136)
-@test isapprox(bic(gm14), 498.9325766164946)
-@test isapprox(coef(gm14),
-    [0.164303051291, -0.7500299832, -1.36469792994, -1.68672866457], atol=1e-5)
-end
+    for distr in (Binomial, Bernoulli)
+	gm14 = fit(GeneralizedLinearModel, admit ~ rank, admit_agr, distr(), wts=Array(admit_agr[:count]))
```
ararslan: 4 spaces instead of tabs

test/runtests.jl
ararslan: No problem. Git can do some pretty weird stuff sometimes.

src/ftest.jl
```diff
@@ -0,0 +1,92 @@
...
+        end
+    end
+
+    return true
+end
+
+"""
+    ftest(mod::LinPredModel...)
+
+For each sequential pair of linear predictors in `mod`, perform an F-test to determine if the first one fits significantly better than the second
```
nalimilan: Please break lines at 92 chars, even inside docstrings. Missing ending period. Also the function does more than testing, it returns a series of statistics which should be mentioned.

src/ftest.jl
```diff
@@ -0,0 +1,92 @@
...
+            # We have found a predictor variable in model 1 that is not in model 2
+            return false 
+        end
+    end
+
+    return true
+end
+
+"""
+    ftest(mod::LinPredModel...)
```
nalimilan: Actually, this function should be restricted to `LinearModel`. `LinPredModel` includes GLMs, for which the F test does not apply (would need the likelihood ratio test, which would be nice to have at some point).

src/ftest.jl
```diff
@@ -0,0 +1,92 @@
...
+
+For each sequential pair of linear predictors in `mod`, perform an F-test to determine if the first one fits significantly better than the second
+
+Note: This function can easily be used to do an ANOVA. ANOVA is nothing more than a test
+to see if one model fits the data better than another.
+
+# Examples:
+As usual, we want to compare a result across two or more treatments. In the classic ANOVA
```
nalimilan: "usual" probably does not apply to all potential users. Let's remain factual here and just present the way of doing an ANOVA. Also "perfectly good fit" is a bit casual, maybe just "fits the data"?

In the note above, also better remain more neutral and just state that this function can be used to perform an ANOVA by comparing etc.

src/ftest.jl
```diff
@@ -0,0 +1,92 @@
...
+framework, our null hypothesis is that Result~1 is a perfectly good fit to the data. 
+The alternative for ANOVA is that Result~Treatment is a better fit to the data than Result~1
+```jldoctest
+julia> d = DataFrame(Treatment=[1, 1, 1, 2, 2, 2, 1, 1, 1, 2, 2, 2.], Result=[1.1, 1.2, 1, 2.2, 1.9, 2, .9, 1, 1, 2.2, 2, 2]);
```
nalimilan: Please break the line. Also better not use one-letter names, so `df` instead of `d`.

src/ftest.jl
```diff
@@ -0,0 +1,92 @@
...
+    prepend!(Δdf1, [NaN])
+
+    R2 = collect(r2.(mods))
+    ΔR2 = diff(R2)
+    prepend!(ΔR2, [NaN])
+
+    ΔSSR = diff(SSR)
+    prepend!(ΔSSR, [NaN])
+    return CoefTable([df1, Δdf1, nparams, -df2, SSR, ΔSSR, R2, ΔR2, fstat, pval],
+                     ["Residual DOF", "ΔResidual DOF", "DOF", "Δdof", "SSR", "ΔSSR", "R²", "ΔR²", "F*", "p(>F)"],
```
nalimilan: Hmm, the table is going to be too wide for many terminals. We should be able to drop at least "ΔResidual DOF", which is redundant with "Δdof" (which BTW should be written in capitals). You can also take the minus of delta columns so that they are positive: this matches R and Stata, and will save one character. Also rename "Residual DOF" to "Res. DOF".

Then I guess we could tweak the printing of `CoefTable` to print less decimal digits, but that's for later.

Please also cut the line at 92 chars.

src/ftest.jl
```diff
@@ -0,0 +1,92 @@
...
+```
+"""
+function ftest(mods::LinPredModel...)
+    nmodels = length(mods)
+    for i in 2:nmodels
+        issubmodel(mods[i], mods[i-1]) || 
+        throw(ArgumentError("F test is only valid if model 2 is nested in model 1"))
+    end
+
+    SSR = collect(deviance.(mods))
```
nalimilan: Is `collect` really needed here and below?

src/ftest.jl
```diff
@@ -0,0 +1,92 @@
...
+Model 2          11.0           1.0 2.0 -1.0  3.22917 3.10083 -2.22045e-16 -0.960258 241.623 2.48122e-8
+
+
+```
+"""
+function ftest(mods::LinPredModel...)
+    nmodels = length(mods)
+    for i in 2:nmodels
+        issubmodel(mods[i], mods[i-1]) || 
+        throw(ArgumentError("F test is only valid if model 2 is nested in model 1"))
```
nalimilan: Would be great to give the number of the problematic model. Something like `"each model must be nested in the previous one for F test (found for model $i)"`.

test/runtests.jl
```diff
@@ -176,7 +176,7 @@ end
 
 @testset "Gamma IdentityLink" begin
     gm10 = fit(GeneralizedLinearModel, lot1 ~ 1 + u, clotting, Gamma(), IdentityLink(),
-        convTol=1e-8)
```
nalimilan: Unrelated space change here and below.

test/runtests.jl
```diff
@@ -321,3 +321,25 @@ end
...
+    nullmod = lm(@formula(Result~1), d).model
+    @test GLM.issubmodel(nullmod, mod)
+    @test !GLM.issubmodel(othermod, mod)
+    
+    @test_throws ArgumentError ftest(mod, othermod)
+    
+    ft = ftest(mod, nullmod)
+    @test isapprox(ft.cols[end][2],  2.481215056713184e-8)
+    # Test output
+    @test sprint(show, ftest(mod, nullmod)) == "         Residual DOF ΔResidual DOF DOF Δdof      SSR    ΔSSR           R²       ΔR²      F*      p(>F)\nModel 1          10.0           NaN 3.0  NaN 0.128333     NaN     0.960258       NaN     NaN        NaN\nModel 2          11.0           1.0 2.0 -1.0  3.22917 3.10083 -2.22045e-16 -0.960258 241.623 2.48122e-8\n"
```
nalimilan: Use the triple-quoted syntax for this (`"""`).

test/runtests.jl
```diff
@@ -321,3 +321,25 @@ end
     @test isapprox(lm(Xf, y).pp.beta0, ones(2))
     @test isapprox(lm(X, yf).pp.beta0, ones(2))
 end
+
+@testset "F test for model comparison" begin
+    d = DataFrame(Treatment=[1, 1, 1, 2, 2, 2, 1, 1, 1, 2, 2, 2.], Result=[1.1, 1.2, 1, 2.2, 1.9, 2, .9, 1, 1, 2.2, 2, 2], Other=[1, 1, 2, 1, 2, 1, 3, 1, 1, 2, 2, 1])
```
nalimilan: Break at 92 chars.

src/ftest.jl
LewisHein: Yes -- `diff` and `view` don't work on tuples

test/runtests.jl
LewisHein: Actually this is in response to the review above that I had de-indented most of runtests.jl

LewisHein:
I vote to have a generic name from the start, because that's where we know we want to end up in the long run. It makes more sense (to me) to do things right the first time as far as possible

src/GLM.jl
```diff
@@ -54,7 +54,7 @@ module GLM
         predict,        # make predictions
         updateμ!,      # update the response type from the linear predictor
         wrkresp,        # working response
-        ftest           # compare two models with an F test
+        ftest           # compare  models with an F test
```
nalimilan: This is minor, but you have a double space here.

test/runtests.jl
nalimilan: Hmm, I'm not sure which comment you refer to exactly, but in general PRs should avoid making changes to unrelated parts of the code, and this one in particular does not use a correct alignment. Please revert whitespace changes outside of the code you need to change.

src/ftest.jl
```diff
@@ -77,16 +84,21 @@ function ftest(mods::LinPredModel...)
...
+    ΔR2 = -diff(R2)
     prepend!(ΔR2, [NaN])
 
-    ΔSSR = diff(SSR)
+    ΔSSR = -diff(SSR)
     prepend!(ΔSSR, [NaN])
-    return CoefTable([df1, Δdf1, nparams, -df2, SSR, ΔSSR, R2, ΔR2, fstat, pval],
-                     ["Residual DOF", "ΔResidual DOF", "DOF", "Δdof", "SSR", "ΔSSR", "R²", "ΔR²", "F*", "p(>F)"],
+    return CoefTable([df1, nparams, df2, SSR, ΔSSR, R2, ΔR2, fstat, pval],
+                     ["Res. DOF",
```
nalimilan: OTOH, one name per line is a bit too much. Just break the line into two (or three) parts somewhere before 92 chars.

test/runtests.jl
```diff
@@ -335,11 +337,11 @@ end
     ft = ftest(mod, nullmod)
     @test isapprox(ft.cols[end][2],  2.481215056713184e-8)
     # Test output
-    @test sprint(show, ftest(mod, nullmod)) == "         Residual DOF ΔResidual DOF DOF Δdof      SSR    ΔSSR           R²       ΔR²      F*      p(>F)\nModel 1          10.0           NaN 3.0  NaN 0.128333     NaN     0.960258       NaN     NaN        NaN\nModel 2          11.0           1.0 2.0 -1.0  3.22917 3.10083 -2.22045e-16 -0.960258 241.623 2.48122e-8\n"
+    @test sprint(show, ftest(mod, nullmod)) == """         Res. DOF DOF ΔDOF      SSR     ΔSSR           R²      ΔR²      F*      p(>F)\nModel 1      10.0 3.0  NaN 0.128333      NaN     0.960258      NaN     NaN        NaN\nModel 2      11.0 2.0  1.0  3.22917 -3.10083 -2.22045e-16 0.960258 241.623 2.48122e-8\n"""
```
nalimilan: Actually, what I meant is that with triple-quoted string you can replace `\n` with actual line breaks (see the manual). Also add a line break before `"""`.

src/ftest.jl
nalimilan: You missed this occurrence when fixing the others.

src/ftest.jl
```diff
@@ -30,32 +30,39 @@ end
 """
     ftest(mod::LinPredModel...)
 
-For each sequential pair of linear predictors in `mod`, perform an F-test to determine if the first one fits significantly better than the second
+For each sequential pair of linear predictors in `mod`, perform an F-test to determine if 
+the first one fits significantly better than the second.
 
-Note: This function can easily be used to do an ANOVA. ANOVA is nothing more than a test
-to see if one model fits the data better than another.
+A table is returned containing residual degrees-of-freedom, difference in residual DOF
```
nalimilan: The difference in residual DOF isn't returned now. Would also be nice to specify DOF and SSR in parentheses after spelling them in full.

nalimilan:
> I vote to have a generic name from the start, because that's where we know we want to end up in the long run. It makes more sense (to me) to do things right the first time as far as possible

Agreed, so we need to find a good name.`modeltable`, `modtest`, `compmod`, `fittable`, `fittest`, `nesttest`? I couldn't find functions doing the same thing in other software.

test/runtests.jl
LewisHein: Here: https://github.com/JuliaStats/GLM.jl/pull/182#pullrequestreview-39608592. I think what happened is that mis-interpreted a comment by @ararslan.

LewisHein:
I like fittest best

test/runtests.jl
nalimilan: The whitespace changes are still there...

test/runtests.jl
```diff
@@ -164,160 +164,195 @@ end
...
+    @test GLM.issubmodel(nullmod, mod)
+    @test !GLM.issubmodel(othermod, mod)
+    
+    @test_throws ArgumentError ftest(mod, othermod)
+    
+    ft = ftest(mod, nullmod)
+    @test isapprox(ft.cols[end][2],  2.481215056713184e-8)
+    # Test output
+    @test sprint(show, ftest(mod, nullmod)) == 
+"""
```
nalimilan: Sorry to be a pain, but this line should have four spaces of indentation more than the last line. Triple quoted strings do not take the indentation into account.

test/runtests.jl
LewisHein: I am baffled. I went back to a version of the file before this comment and pasted in what it had above my additions, verbatim. If this didn't work, I don't know what will.

LewisHein:
I struggled with this wording. Do you have a better way to say it?

LewisHein:
Maybe "from the previous model"?

src/ftest.jl
```diff
@@ -0,0 +1,97 @@
...
+    return true
+end
+
+"""
+    ftest(mod::LinearModel...)
+
+For each sequential pair of linear predictors in `mod`, perform an F-test to determine if 
+the first one fits significantly better than the second.
+
+A table is returned containing residual degrees-of-freedom (DOF), from the last model,
```
ararslan: No hyphens between "degrees of freedom," no comma after the abbreviation. Regarding "the last model," maybe "the preceding model"? That seems at least somewhat appropriate, given that things are described as being sequential above.

src/ftest.jl
```diff
@@ -33,9 +33,9 @@ end
 For each sequential pair of linear predictors in `mod`, perform an F-test to determine if 
 the first one fits significantly better than the second.
 
-A table is returned containing residual degrees-of-freedom (DOF), from the last model, 
-degrees of freedom, difference in DOF from the last model, sum of squared residuals, (SSR)
-difference in SSR from the last model, R², difference in R² from the last model, and F
+A table is returned containing residual degrees of freedom (DOF), degrees of freedom,
+difference in DOF from the preceeding model, sum of squared residuals (SSR), difference in
```
LewisHein: Sorry, my brain was (obviously) only half engaged...

test/runtests.jl
nalimilan: Maybe try with another editor?

nalimilan:
If we want to make this function (`fittest`? sounds a bit like a superlative...) fully generic, the function itself should be defined in StatsBase (in the statsmodels.jl file) with an empty definition, so that other packages will be able to extend it without conflicting with GLM. Can you do that?

test/runtests.jl
nalimilan: A possible explanation is that there are conflicts with master on these lines; something weird is going one. Calling `git rebase -i master` and fixing the conflicts manually shoudl help.

LewisHein:
If `fittest` sounds a little presumptious (which I agree it does), why
not `testfit`?

test/runtests.jl
```diff
@@ -321,3 +321,38 @@ end
...
+    @test isapprox(ft.cols[end][2],  2.481215056713184e-8)
+    # Test output
+    @test sprint(show, ftest(mod, nullmod)) == 
+        """
+                 Res. DOF DOF ΔDOF      SSR     ΔSSR           R²      ΔR²      F*      p(>F)
+        Model 1      10.0 3.0  NaN 0.128333      NaN     0.960258      NaN     NaN        NaN
+        Model 2      11.0 2.0  1.0  3.22917 -3.10083 -2.22045e-16 0.960258 241.623 2.48122e-8
+        """
+    
+    bigmod = lm(Result~Treatment+Other, d).model
```
ararslan: Need an `@formula` here

src/GLM.jl
```diff
@@ -6,7 +6,7 @@ module GLM
     using Base.LinAlg.LAPACK: potrf!, potrs!
     using Base.LinAlg.BLAS: gemm!, gemv!
     using Base.LinAlg: QRCompactWY, Cholesky
-    using StatsBase: StatsBase, CoefTable, StatisticalModel, RegressionModel
+    using StatsBase: StatsBase, CoefTable, StatisticalModel, RegressionModel, PValue
```
ararslan: Looks like `PValue` isn't being used here, so this change isn't necessary.

ararslan:
I'm not really a fan of `fittest` nor of `testfit`, though of those two I guess I prefer the former. Perhaps something like `comparemodels`? A little long, but it makes it clear that it's specifically for comparing the fit of models, whereas `ftest` for example sounds like something that should go in HypothesisTests.jl.

LewisHein:
I'd be fine with `compmod` or `comparemodels`

andreasnoack:
Sorry for joining the party a bit late there. What about just using `test`?

ararslan:
`test` on its own seems insufficiently descriptive to me. What's it testing?

nalimilan:
I agree that `test` is too general. `comparemodels` or `modelstable` could work. BTW, the function requires models to be nested, so we should use this term in the name if we want to have another function printing statistics (but no F test) for non-nested models (this kind of table is useful in many cases too, e.g. using AIC/BIC). `nesttest`, `comparenested`, `nesttable`...

LewisHein:
> I agree that `test` is too general. `comparemodels` or `modelstable` could work. BTW, the function requires models to be nested, so we should use this term in the name if we want to have another function printing statistics (but no F test) for non-nested models (this kind of table is useful in many cases too, e.g. using AIC/BIC). `nesttest`, `comparenested`, `nesttable`...


How about `comparenested?`

ararslan:
Sure, I'd be okay with `comparenested`. @andreasnoack?

andreasnoack:
> test on its own seems insufficiently descriptive to me. What's it testing?

What else could `test(RegressionModel, RegressionModel)` mean?

Testing two nested models is probably the most common thing to do in statistics.

> Sure, I'd be okay with `comparenested`. @andreasnoack?

I don't think it is good. In statistics, you *test* models. As mentioned above, *nestedness* is the default, so it shouldn't be necessary to mention *nested*.

nalimilan:
> What else could test(RegressionModel, RegressionModel) mean?

Nothing, but that means StatsBase would have to claim the `test` function, which I'm afraid will be overloaded a lot for various stats operations, making it quite polysemic. It's hard to read documentation for a function when there are many unrelated docstrings attached to it.

> As mentioned above, nestedness is the default, so it shouldn't be necessary to mention nested.

One issue concerns what to do with non-nested models:
1) Throw an error, but that means we cannot print a model comparison table for these (with DF and AIC/BIC), even if it would be useful.
2) Automatically change the kind of table which is returned not to include a F test: this is too magical for my taste.

A solution would be to use another term than `test` for comparisons on non-nested models, since that's not really a test. I would be OK with that, but that doesn't address my first concern.

LewisHein:
So is this waiting on me now? What was the final decision on the name?

ararslan:
The last piece here is just name bikeshedding.

andreasnoack:
> Nothing, but that means StatsBase would have to claim the `test` function

As such it wouldn't be a problem. This is only a problem if another package with no dependency relation to `StatsBase` defines `test` and if there is a common dependent package for this package and `StatsBase`. Since `StatsBase` is so deep in the dependency tree, I'm not convinced (yet) it would be a problem.

> It's hard to read documentation for a function when there are many unrelated docstrings attached to it.

We'll just have to deal with this. This is a natural consequence of being able to overload functions and we want to write `testRegressionModelAgainstRegressionModel(::RegressionModel,RegressionModel)` instead of `test(::RegressionModel,RegressionModel)`. It is also difficult to search for help for `*`.

I don't think a `test` function would have to handle non-nested models. A model summary like you describe could be handled by a different function and non-nested testing be a specialized function with a different name since it is a less standard thing to do. However, I'm a bit concerned about return types here. Do you want to return different types depending on the test being, e.g. LR, LM, Wald, F etc. In that case we might need/want to split up the function name or introduce type dispatch.

nalimilan:
I'm fine with `test` if @andreasnoack thinks that's the best term, since you appear to be the only one with a strong opinion on this. Let's not block this PR too long just because of the name.

> I'm a bit concerned about return types here. Do you want to return different types depending on the test being, e.g. LR, LM, Wald, F etc. In that case we might need/want to split up the function name or introduce type dispatch.

I'm not sure it's really worth it. The returned tables will all have the same structure, and calling function's return types won't depend on the kind of test (so no type stability will happen down in the stack). If needed, callers can just look at the column names to detect the kind of test which was performed. Also we can always split this into several types later, which would all inherit from a `CoefTable`-like abstract type and share most of their implementation code to prevent duplication.

ararslan:
I really don't think `test` is the right way to go here, but I don't want to hold up this PR over that objection.

andreasnoack:
@LewisHein Sorry that I haven't shared my comments earlier. It appears that we maintainers don't fully agree on the design here. I really don't think `CoefTable` can be the long term return type solution here. The current version is a great improvemenet over what we have but if we don't provide something lighter, people won't use it when many hypotheses are tested as in e.g. statistical genomics. It should be possible to combine convenience and efficiency with a special immutable with some "extractors" defined for such a struct. I think it would be unfortunate to be stuck with an allocating version in the future so either we should merge and be okay with breaking later or work a little more on this version before we merge.

nalimilan:
@andreasnoack A previous version of the PR introduced a custom type for this, so we can probably go back to it if needed. But what kind of structure do you have in mind to avoid any allocation? One holding a series of tuples? Do you really think the cost of the allocation will be significant compared with that of fitting a model?

andreasnoack:
Yes. I'm working on a statistical genomics problem and I have more customized code than I'd like to avoid allocations during the millions of models fitted and tested. Allocations become a big problem for heavily threaded code because the GC is a sequential bottleneck. I think an (immutable) `struct` would do the job here. with scalar fields needed for the specific test. I.e. at least the value of the test statistic but probably a bit more.

nalimilan:
@LewisHein Do you feel like implementing the custom type @andreasnoack described, or do you prefer we merge this as-is and adjust it later? Sorry for the conflicting instructions here, designing APIs is hard.

> I think an (immutable) struct would do the job here. with scalar fields needed for the specific test. I.e. at least the value of the test statistic but probably a bit more.

@andreasnoack I spoke about tuples because the structure needs to hold the result of multiple tests. How would you store these only with scalars?

andreasnoack:
I get it know. Yes, maybe a tuple of the pairwise tests could work.

LewisHein:
I don't think I have time now to re-write the return type again --
other projects are pressing. Sorry.

nalimilan:
@LewisHein OK, thanks for your time.

@andreasnoack Are you OK with merging this as-is and improving it later?

andreasnoack:
As such, I don't have a problem with merging this. As mentioned, it is a great improvement over the current situation. The problem is the future deprecation tax which most likely will result in status quo here and people just implementing their own custom solutions elsewhere.

nalimilan:
> The problem is the future deprecation tax which most likely will result in status quo here and people just implementing their own custom solutions elsewhere.

AFAIK we don't usually consider the cost of deprecations as an issue if that improves API.

andreasnoack:
> AFAIK we don't usually consider the cost of deprecations as an issue if that improves API.

Often you don't see these costs. It is just PRs that don't happen. My postulate is that we miss many improvement oppertunities because people don't want to bother with tedious deprecations. Breaking things is costly for users and deprecating things is costly for developers. As the number of users grows you'll have to weight users higher and higher until you are completely stock with an API.

LewisHein:
Hmm, this doesn't sound good. Maybe I could find the time to re-do
this... Exactly what changes would be needed?

nalimilan:
> Hmm, this doesn't sound good. Maybe I could find the time to re-do this... Exactly what changes would be needed?

That would be nice. AFAICT, that would mean going back to a specific type like you had before, but which would store each column of the current `CoefTable` as field which would be a tuple of values (rather than vectors). But regarding creating this structure without doing any allocations, I'm afraid it will be a bit tedious since e.g. `diff` doesn't work for tuples (as you noted). We can probably leave that for another PR since at least the public interface would be settled.

ararslan:
> `diff` doesn't work for tuples

```julia
_diff(t::Tuple) = (d = diff(collect(x)); ntuple(i -> d[i], length(d)))
_diff(x) = diff(x)
```
😛

ararslan:
Yeah, it's horribly inefficient, but if you need to `diff` a `Tuple`, it works in a pinch 😉

LewisHein:
Finally got some progress

Tell me what you think

nalimilan:
Thanks. I think what @andreasnoack envisioned was replacing array fields with `Tuple{T, N}`, making `N` a type parameter (which would replace `ntests`).

src/ftest.jl
```diff
@@ -0,0 +1,144 @@
...
+    ΔSSR = -diff(SSR)
+    prepend!(ΔSSR, [NaN])
+    return CoefTable([df1, nparams, df2, SSR, ΔSSR, R2, ΔR2, fstat, pval],
+                     ["Res. DOF", "DOF", "ΔDOF", "SSR", "ΔSSR", "R²", "ΔR²", "F*",
+                     "p(>F)"],
+                     ["Model $i" for i in 1:nmodels])=#
+end
+
+function show(io::IO, ftr::FTestResult)
+    ssr = ftr.SSR; nparams = ftr.nparams; dof_resid = ftr.dof_resid
```
nalimilan: No need to create these local variables, since the type is immutable accessing the fields repeatedly will be as efficient (and `ftr.` isn't that long).

src/ftest.jl
```diff
@@ -0,0 +1,144 @@
...
+    Δdof = -diff(dof_resid)
+    Δssr = -diff(ssr)
+    ΔR² = -diff(R²)
+
+    nc = 10
+    nr = ftr.ntests
+    outrows = Array{String, 2}(nr+1, nc)
+    
+    outrows[1, :] = ["", "Res. DOF",  "DOF",  "ΔDOF",  "SSR",  "ΔSSR",  "R²",  "ΔR²",  "F*",  "p(>F)"]
+    outrows[2, :] = ["Model 1", @sprintf("%.4f", dof_resid[1]), @sprintf("%.4f", nparams[1]), " ", @sprintf("%.4f", ssr[1]), " ", @sprintf("%.4f", R²[1]), " ", " ", " "]
```
nalimilan: Please wrap lines at 92 chars.

LewisHein:
Unfortunately, I've hit a major snag in moving to tuples. Here's why: The code looks like 

```julia
type FTestResult{N}
    SSR::NTuple{N, Float64}
    nparams::NTuple{N, Int}
    dof_resid::NTuple{N, Int}
    R2::NTuple{N, Float64}
    fstat::NTuple{N, Float64}
    pval::NTuple{N, PValue}
end
```

The problem is that `fstat` and `pval` both have length `N-1` rather than length `N`. I either have to pad them on creation which means lots of extra code in `ftest` that just obscures what it is doing, or have something like

```julia
type FTestResult{N}
    SSR::NTuple{N, Float64}
    nparams::NTuple{N, Int}
    dof_resid::NTuple{N, Int}
    R2::NTuple{N, Float64}
    fstat::NTuple{N-1, Float64}
    pval::NTuple{N-1, PValue}
end
```

Which won't work, for reasons that aren't clear to me.

Several workarounds come to mind, such as building an `NMinusOneTuple` type for the purpose, or  padding the offending tuples with NaN values. None are optimal; What do you prefer? Is there a better way that I have overlooked?

nalimilan:
I'd just pad with `NaN`, that should't create any issues.

LewisHein:
Only the issue of obscuring the meaning of the code in `ftest`. I feel that this code is pretty clean right now and will explode if I try to pad things.

Not to say I am unwilling to do so if necessary, but it seems like there should be a better way. Something I just thought of would be to make `fstat` and `pval` into `Tuple{Vararg{Int}}` types. I don't see that losing this length information would hurt efficiency

nalimilan:
I don't see why it could make the code so obscure. Just do `pval = obj.pval[-1]` at the top of the function. I'm not sure about the performance impact of `Tuple{Vararg{Int}}`, but we could try that if you want.

src/ftest.jl
```diff
@@ -1,42 +1,48 @@
...
+
+ return true
+ end
+
+_diffn{N, T}(t::NTuple{N, T})::NTuple{N, T} =  ntuple(i->t[i]-t[i+1], N-1)
+
+_diff{N, T}(t::NTuple{N, T})::NTuple{N, T} =  ntuple(i->t[i+1]-t[i], N-1)
+
+import Base: ./
+./{N, T1, T2}(t1::NTuple{N, T1}, t2::NTuple{N, T2}) = ntuple(i->t1[i]/t2[i], N)
```
nalimilan: Unfortunately you can't do that as that will affect other packages. You can just use `ntuple` directly, or define a custom function.

src/ftest.jl
```diff
@@ -0,0 +1,155 @@
+type FTestResult{N}
+    ssr::NTuple{N, Float64}
+    nparams::NTuple{N, Int}
+    dof_resid::NTuple{N, Int}
+    r2::NTuple{N, Float64}
+    fstat::Tuple{Vararg{Float64}}
+    pval::Tuple{Vararg{PValue}}
+end
+
+ #function FTestResult{N}(SSR1::Array{Float64, 1}, fstati
```
nalimilan: Remove this comment (and others), and fix indentation below.

src/ftest.jl
```diff
@@ -0,0 +1,155 @@
...
+end
+
+function show{N}(io::IO, ftr::FTestResult{N})
+    Δdof = _diffn(ftr.dof_resid)
+    Δssr = _diffn(ftr.ssr)
+    ΔR² = _diffn(ftr.r2)
+
+    nc = 10
+    nr = N
+    outrows = Array{String, 2}(nr+1, nc)
```
nalimilan: `Matrix{String}` is slightly nicer to read.

src/ftest.jl
```diff
@@ -0,0 +1,155 @@
...
+    colwidths = length.(outrows)
+    max_colwidths = [maximum(view(colwidths, :, i)) for i in 1:nc]
+
+    for r in 1:nr+1
+        for c in 1:nc
+            cur_cell = outrows[r, c]
+            cur_cell_len = length(cur_cell)
+            
+            print(io, cur_cell)
+            print(io, RepString(" ", max_colwidths[c]-cur_cell_len+1))
```
nalimilan: Use `" "^(max_colwidths[c]-cur_cell_len+1)` as `RepString` is likely to be deprecated at some point.

src/ftest.jl
```diff
@@ -0,0 +1,155 @@
...
+    Δssr = _diffn(ftr.ssr)
+    ΔR² = _diffn(ftr.r2)
+
+    nc = 10
+    nr = N
+    outrows = Array{String, 2}(nr+1, nc)
+    
+    outrows[1, :] = ["", "Res. DOF",  "DOF",  "ΔDOF",  "SSR",
+                    "ΔSSR",  "R²",  "ΔR²",  "F*",  "p(>F)"]
+    outrows[2, :] = ["Model 1", @sprintf("%.4f", ftr.dof_resid[1]),
```
nalimilan: Please use a more regular organization, it's very hard to read as-is. Same below.

Also, DOF and number of parameters should be printed as integers, not floats.

test/runtests.jl
```diff
@@ -321,3 +321,38 @@ end
...
+    
+    @test_throws ArgumentError ftest(mod, othermod)
+    
+    ft = ftest(mod, nullmod)
+    @test isapprox(ft.pval[1].v,  2.481215056713184e-8)
+    # Test output
+    @test sprint(show, ftest(mod, nullmod)) == 
+        """
+                Res. DOF DOF    ΔDOF    SSR    ΔSSR    R²      ΔR²    F*       p(>F) 
+        Model 1 10.0000  3.0000         0.1283         0.9603
```
nalimilan: Not a big deal, but I wonder whether it would be possible to force `@printf` to align number on the decimal separator (see e.g. the R² column).

src/ftest.jl
```diff
@@ -0,0 +1,155 @@
+type FTestResult{N}
+    ssr::NTuple{N, Float64}
+    nparams::NTuple{N, Int}
```
nalimilan: Probably better call this `dof`, as it's the result of calling `dof(model)`? Also `params` is going to have a slightly different meanings from `coef` (https://github.com/JuliaStats/StatsBase.jl/pull/274/).

test/runtests.jl
LewisHein: Maybe, but I'd have to play with it a good bit to figure out how.

nalimilan: Sure, that was just an idea. I know `CoefTable` has the same issue.

src/ftest.jl
```diff
@@ -1,48 +1,46 @@
...
+    return true
+end
 
 _diffn{N, T}(t::NTuple{N, T})::NTuple{N, T} =  ntuple(i->t[i]-t[i+1], N-1)
 
 _diff{N, T}(t::NTuple{N, T})::NTuple{N, T} =  ntuple(i->t[i+1]-t[i], N-1)
 
-import Base: ./
-./{N, T1, T2}(t1::NTuple{N, T1}, t2::NTuple{N, T2}) = ntuple(i->t1[i]/t2[i], N)
+dividetuple{N, T1, T2}(t1::NTuple{N, T1}, t2::NTuple{N, T2}) = ntuple(i->t1[i]/t2[i], N)
```
nalimilan: `T1` and `T2` are not needed. Maybe call this `dividetuples`?

src/ftest.jl
```diff
@@ -121,21 +102,47 @@ function show{N}(io::IO, ftr::FTestResult{N})
...
     nr = N
-    outrows = Array{String, 2}(nr+1, nc)
+    outrows = Matrix{String}(nr+1, nc)
     
-    outrows[1, :] = ["", "Res. DOF",  "DOF",  "ΔDOF",  "SSR",
-                    "ΔSSR",  "R²",  "ΔR²",  "F*",  "p(>F)"]
-    outrows[2, :] = ["Model 1", @sprintf("%.4f", ftr.dof_resid[1]),
-                     @sprintf("%.4f", ftr.nparams[1]), " ", @sprintf("%.4f", ftr.ssr[1]),
-                     " ", @sprintf("%.4f", ftr.r2[1]), " ", " ", " "]
+    outrows[1, :] = [
```
nalimilan: I'd group these a bit more, e.g. on two or three lines.

test/runtests.jl
```diff
@@ -339,9 +339,9 @@ end
     # Test output
     @test sprint(show, ftest(mod, nullmod)) == 
         """
-                Res. DOF DOF    ΔDOF    SSR    ΔSSR    R²      ΔR²    F*       p(>F) 
-        Model 1 10.0000  3.0000         0.1283         0.9603                        
-        Model 2 11.0000  2.0000 -1.0000 3.2292 -3.1008 -0.0000 0.9603 241.6234 <1e-7 
+                Res. DOF DOF ΔDOF SSR    ΔSSR    R²      ΔR²    F*       p(>F) 
+        Model 1 10       3        0.1283         0.9603
```
nalimilan: Can you align integers on the right? That makes it easier to read by columns since digits of the same rank are aligned. (BTW, if you do the same for floats, they will be automatically aligned on the decimal separator since their are fixed to four digits after it.)

src/ftest.jl
```diff
@@ -121,21 +102,47 @@ function show{N}(io::IO, ftr::FTestResult{N})
...
+                     "ΔDOF",
+                     "SSR",
+                     "ΔSSR",
+                     "R²",
+                     "ΔR²",
+                     "F*",
+                     "p(>F)"
+                     ]
+
+    outrows[2, :] = [
```
nalimilan: Here too you can put an item on the first line, and the closing bracket with the last element. Same below.

src/ftest.jl
```diff
@@ -104,45 +102,20 @@ function show{N}(io::IO, ftr::FTestResult{N})
...
-                    @sprintf("%.4f", ftr.r2[1]),
-                    " ",
-                    " ",
-                    " "
-                    ]
+    outrows[1, :] = ["", "Res. DOF", "DOF", "ΔDOF", "SSR", "ΔSSR",
+                     "R²", "ΔR²", "F*", "p(>F)"]
+
+    outrows[2, :] = ["Model 1", @sprintf("%.0d", ftr.dof_resid[1]),
+                    @sprintf("%.0d", ftr.dof[1]), " ",
```
nalimilan: Should be aligned with `"` (as it's part of the array, so inside `[`).

src/ftest.jl
```diff
@@ -0,0 +1,140 @@
...
+        Res. DOF DOF ΔDOF    SSR    ΔSSR      R²    ΔR²       F* p(>F)
+Model 1       10   3      0.1283          0.9603                      
+Model 2       11   2   -1 3.2292 -3.1008 -0.0000 0.9603 241.6234 <1e-7
+```
+"""
+function ftest(mods::LinearModel...)
+    nmodels = length(mods)
+    for i in 2:nmodels
+        issubmodel(mods[i], mods[i-1]) || 
+        throw(ArgumentError("F test $i is only valid if model $i is nested in model $i-1"))
```
ararslan: Should this be `$(i-1)`?

src/ftest.jl
```diff
@@ -0,0 +1,140 @@
...
+                padding = " "*padding
+            end
+            
+            print(io, padding)
+            print(io, cur_cell)
+        end
+        print(io, "\n")
+    end
+end
+
```
nalimilan: Since it looks like we'll need another round of fixes, it would be nice to remove these two empty lines in your next (and hopefully last) commit.

src/ftest.jl
```diff
@@ -0,0 +1,140 @@
...
+julia> dat = DataFrame(Treatment=[1, 1, 1, 2, 2, 2, 1, 1, 1, 2, 2, 2.],
+                     Result=[1.1, 1.2, 1, 2.2, 1.9, 2, .9, 1, 1, 2.2, 2, 2]);
+julia> mod = lm(@formula(Result~Treatment), dat        );
+julia> nullmod = lm(@formula(Result~1), d);
+julia> ft = ftest(mod.model, nullmod.model)
+        Res. DOF DOF ΔDOF    SSR    ΔSSR      R²    ΔR²       F* p(>F)
+Model 1       10   3      0.1283          0.9603                      
+Model 2       11   2   -1 3.2292 -3.1008 -0.0000 0.9603 241.6234 <1e-7
+```
+"""
```
ararslan: I'd format it like this:
```julia
"""
    ftest(mod::LinearModel...)

For each sequential pair of linear predictors in `mod`, perform an F-test to determine if 
the first one fits significantly better than the next.

A table is returned containing residual degrees of freedom (DOF), degrees of freedom,
difference in DOF from the preceding model, sum of squared residuals (SSR), difference in
SSR from the preceding model, R², difference in R² from the preceding model, and F-statistic
and p-value for the comparison between the two models.

!!! note
    This function can be used to perform an ANOVA by testing the relative fit of two models
    to the data

# Examples

Suppose we want to compare the effects of two or more treatments on some result. Because
this is an ANOVA, our null hypothesis is that `Result~1` fits the data as well as
`Result~Treatment`.
 
```jldoctest
julia> dat = DataFrame(Treatment=[1, 1, 1, 2, 2, 2, 1, 1, 1, 2, 2, 2.],
                       Result=[1.1, 1.2, 1, 2.2, 1.9, 2, .9, 1, 1, 2.2, 2, 2]);

julia> mod = lm(@formula(Result~Treatment), dat);

julia> nullmod = lm(@formula(Result~1), dat);

julia> ft = ftest(mod.model, nullmod.model)
Model 1       10   3      0.1283          0.9603                      
Model 2       11   2   -1 3.2292 -3.1008 -0.0000 0.9603 241.6234 <1e-7
"""
```
There are a few modifications in here, too lazy to enumerate them rather than just copy-pasting and editing :P

src/ftest.jl
```diff
@@ -0,0 +1,140 @@
...
+
+        if !var_in_mod2 # We have found a predictor variable in model 1 that is not in model 2
+            return false 
+        end
+    end
+
+    return true
+end
+
+_diffn{N, T}(t::NTuple{N, T})::NTuple{N, T} =  ntuple(i->t[i]-t[i+1], N-1)
```
nalimilan: `::NTuple{N, T}` is actually wrong here, since the result has `N-1` elements. Just drop the type assertion (same below).

nalimilan: And BTW, that's the cause of the tests failure.

There's also a double space.

src/ftest.jl
LewisHein: Yes, good catch

nalimilan:
Note that I resolved conflicts with master directly via GitHub, so you'll have either to pull from the branch, or to rebase locally against master and then force-push.

LewisHein:
OK -- fingers crossed...

LewisHein:
I have no idea what happened. I ran the tests before submitting, but now runtests.jl is all messed up.

LewisHein:
Now travis says it's failing on nightly Julia. What now?

nalimilan:
Don't worry about the nightlies, the failure was already there. I'll let @ararslan merge if he's OK with the new version.

nalimilan:
Thanks @LewisHein! Sorry if the process was a bit long. If you want to implement the likelihood ratio test at some point, now that we have settled the API, that would be great!

pbastide:
Hi @LewisHein, @nalimilan and @ararslan,

Thank you for all your work on the GLM package !

We would need the `ftest` function in the [PhyloNetwoks](https://github.com/crsl4/PhyloNetworks.jl) package, that already depends on GLM. Right now we use our own naive implementation of this function, but we would be happy to rely on your release in the future.

However, if I'm correct, right now v0.7.0 does not contain the changes you made here. Are you planing on making this feature available in a release soon ? That would be great for us, so that we could require this new version, and use your function.

Sorry if this is not the right place to ask, or if the answer is already somewhere else. If so, I'd be grateful for you to point me in the right direction.

Thank you !

nalimilan:
Sure. I've just tagged a new release, see https://github.com/JuliaLang/METADATA.jl/pull/11274.

pbastide:
Thank you, that was fast !

PR #187, opened 2017-07-18
Clamp cloglog inverse link

dmbates:
See https://github.com/dmbates/MixedModels.jl/issues/92 for an example of what can happen if you don't clamp the values on the mean scale to (0,1)

If μ, the value of the inverse link applied to the linear predictor, rounds down to zero or, more commonly, rounds up to one, then the `glmvar` value is zero, resulting in a division by zero.  @andreasnoack contributed special method for `glmvar` the helps with this problem if the inverse link function is "odd" (actually f(-x) == 1 - f(x)) in this case).  In the most common case of the canonical (Logit) link, the calculation of the glmvar can be and is by-passed.  

This link doesn't have such a nice way of avoiding the numerical problems, as far as I can see.

### DIFF
--- a/src/glmtools.jl
+++ b/src/glmtools.jl
@@ -14,8 +14,12 @@ linkinv(::CauchitLink, η) = oftype(η, 0.5) + atan(η) / pi
 mueta(::CauchitLink, η) = one(η) / (pi * (one(η) + abs2(η)))
 
 linkfun(::CloglogLink, μ) = log(-log1p(-μ))
-linkinv(::CloglogLink, η) = -expm1(-exp(η))
-mueta(::CloglogLink, η) = exp(η) * exp(-exp(η))
+function linkinv(::CloglogLink, η::T) where {T<:AbstractFloat}
+    clamp(-expm1(-exp(η)), eps(T), one(T) - eps(T))
+end
+function mueta(::CloglogLink, η::T) where {T<:AbstractFloat}
+    max(eps(T), exp(η) * exp(-exp(η)))
+end
 
 linkfun(::IdentityLink, μ) = μ
 linkinv(::IdentityLink, η) = η
--- a/src/glmtools.jl
+++ b/src/glmtools.jl
@@ -14,10 +14,10 @@ linkinv(::CauchitLink, η) = oftype(η, 0.5) + atan(η) / pi
 mueta(::CauchitLink, η) = one(η) / (pi * (one(η) + abs2(η)))
 
 linkfun(::CloglogLink, μ) = log(-log1p(-μ))
-function linkinv(::CloglogLink, η::T) where {T<:AbstractFloat}
+function linkinv{T<:AbstractFloat}(::CloglogLink, η::T) 
     clamp(-expm1(-exp(η)), eps(T), one(T) - eps(T))
 end
-function mueta(::CloglogLink, η::T) where {T<:AbstractFloat}
+function mueta{T<:AbstractFloat}(::CloglogLink, η::T)
     max(eps(T), exp(η) * exp(-exp(η)))
 end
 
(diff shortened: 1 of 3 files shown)


src/glmtools.jl
```diff
@@ -14,8 +14,12 @@ linkinv(::CauchitLink, η) = oftype(η, 0.5) + atan(η) / pi
 mueta(::CauchitLink, η) = one(η) / (pi * (one(η) + abs2(η)))
 
 linkfun(::CloglogLink, μ) = log(-log1p(-μ))
-linkinv(::CloglogLink, η) = -expm1(-exp(η))
-mueta(::CloglogLink, η) = exp(η) * exp(-exp(η))
+function linkinv(::CloglogLink, η::T) where {T<:AbstractFloat}
```
ararslan: The `where` syntax for type parameters is only available in Julia 0.6 and later, but this package still supports 0.5. To continue to support 0.5, this will have to be
```julia
function linkinv{T<:AbstractFloat}(::CloglogLink, η::T)
```

dmbates:
Bump.  The Travis failures are due to Distributions not compiling on the nightlies and are unrelated to this PR.

ararslan:
Is it ever the case that `η` could be something other than a floating point value, e.g. an integer? Perhaps it would be worth keeping the existing definition as a fallback for other types just in case.

src/GLM.jl
```diff
@@ -9,6 +9,9 @@ module GLM
     using StatsBase: StatsBase, CoefTable, StatisticalModel, RegressionModel
     using StatsFuns: logit, logistic
     using Distributions: sqrt2, sqrt2π
+    if VERSION >= v"0.6"
```
ararslan: ? SpecialFunctions supports 0.5

src/glmtools.jl
```diff
@@ -14,8 +14,12 @@ linkinv(::CauchitLink, η) = oftype(η, 0.5) + atan(η) / pi
 mueta(::CauchitLink, η) = one(η) / (pi * (one(η) + abs2(η)))
 
 linkfun(::CloglogLink, μ) = log(-log1p(-μ))
-linkinv(::CloglogLink, η) = -expm1(-exp(η))
-mueta(::CloglogLink, η) = exp(η) * exp(-exp(η))
+function linkinv{T<:Real}(::CloglogLink, η::T)
+    clamp(-expm1(-exp(η)), eps(T), one(T) - eps(T))
```
ararslan: `eps` isn't defined for all `Real` subtypes, e.g. `Int`

src/GLM.jl
andreasnoack:
```julia
julia> using SpecialFunctions

julia> erfc(1.0)
WARNING: both SpecialFunctions and Base export "erfc"; uses of it in module Main must be qualified
ERROR: UndefVarError: erfc not defined
```
but I've decided to handle this with `import` instead in `anj/chol`.

src/glmtools.jl
andreasnoack: I know but this a compromise to make autodiff types go through and promotion shouldn't happen here anyway. We don't want an unclamped code path. It would be very tricky to figure out why code taking that path would fail.

ararslan: Could always add a separate method like
```julia
linkinv(::CloglogLink, η::Integer) = linkinv(::CloglogLink, float(η))
```

src/GLM.jl
ararslan: Ah, I think Steven Johnson's recent SpecialFunctions change broke that. I can submit a PR there to fix it.

ararslan: Never mind, that's not what I was thinking of. Yeah I have no idea why that's happening because SpecialFunctions shouldn't be exporting anything on 0.5. Good that `import` can get around it.

andreasnoack:
@dmbates I've pushed a test case that used to fail but works with your changes here. Will merge when tests pass.

src/glmtools.jl
dmbates: Why not do the calculation in two stages?
```
function linkinv(::CloglogLink, μ)
    μ = -expm1(-exp(η))
    ep = eps(typeof(μ))
    clamp(μ, ep, 1 - ep)
end
```

PR #190, opened 2017-07-24
RFC: Introduce complementary return value in linkinv

andreasnoack:
Introduce complementary return value in `linkinv` indicating when the complimentary
inverse link function has been evaluated instead of the inverse link function.
This greatly improves the range for which the linear predictor gives non-zero
variance components.

This PR also changes the convergence criterion to use absolute difference alongside relative difference.
This is useful when data is completely separated.

@dmbates This will greatly improve the accuracy of the `glmvar` evaluation in the `CloglogLink`
case. More generally, it also simplifies the code and avoids computing μ twice so there should be
a (slight) speedup here in the non-canonical link case. The cost is that `linkinv` returns a tuple of two arguments
which might be a little inconvenient but since it is mainly an internal helper function I guess it is okay.

I think we have discussed earlier what can be assumed about `(dμ/dη)/glmvar` but I've now convinced myself
that we can avoid clamping the values and instead assume that `0/0=0`. For the score contributions in the binomial case,
the computations are something like (sorry about ∂s. I forgot that these aren't partials)

image

so in practice, I'm setting `wrkresid` to zero when a `NaN` is detected to neutralize the score contribution. For the variance
weight, the computation is something like

image

So as long as the second derivative of μ has limit zero when η goes to infinity, we should be fine. I think that should always be the case.

Similar situations can happen for the Poisson when μ goes to zero but substituting `NaN`s with zeros should be okay
under a similar condition on the second derivative (going to negative infinity).

@dmbates I'm fine with analyzing all the relevant cases but I'm not sure which non-canonical link functions are common outside the binomial model. Do you think there are relevant cases where these assumptions won't hold?

### DIFF
--- a/docs/src/links.jmd
+++ b/docs/src/links.jmd
@@ -1,0 +1,204 @@
+---
+title: Notes on link functions for GLMs
+author: Douglas Bates
+date: July 27, 2017
+---
+
+In a generalized linear model (GLM) the *n*-dimensional random variable $\mathcal{Y}$, whose observed value is $\mathbf{y}$, the *response vector*, is modeled as independent observations from a scalar distribution family $\mathcal{D}$.
+The scalar distributions differ only in their expected values.
+If there are other parameters in the scalar distribution, $\mathcal{D}$, such as a dispersion parameter, they are assumed to be equal across the observations.
+The expected value of $\mathcal{Y}$ is modeled as
+\begin{equation}
+\mathsf{𝖤}(\mathcal{Y}) = \mathbf{\mu} = \mathbf{g}^{-1}(\mathbf{X}\mathbf{β})
+\end{equation}
+where $\mathbf{\eta}=\mathbf{X\beta}$ is the *linear predictor* formed from the $n\times p$ *model matrix*, $\mathbf{X}$, and the $p$-dimensional *coefficient vector*, $\mathbf{\beta}$, and $\mathbf{g}^{-1}$ is the vector-valued *inverse link* function.
+
+The vector-valued *link* and *inverse link* functions are component-wise applications of a scalar link, $g$, and its inverse, $g^{-1}$, which map the possibly restricted range of the mean, $\mu$, to the unrestricted range of $\eta$ and vice-versa.
+For example, if 𝒟 is the [Poisson distribution](https://en.wikipedia.org/wiki/Poisson_distribution) then μ ∈ (0, ∞) while η ∈ (-∞, ∞) and typically a *log link*,
+\begin{equation}
+\eta = g(\mu) = \log(\mu),
+\end{equation}
+with inverse,
+\begin{equation}
+\mu = g^{-1}(\eta) = \exp(\eta),
+\end{equation}
+is used.
+
+In the linear model, which is the model generalized by the GLM, the link and its inverse are the identity function so that
+\begin{equation}
+\mathcal{Y}\sim\mathcal{N}(\mathbf \mu, \sigma^2\mathbf I).
+\end{equation}
+
+The scalar link and inverse link functions should be continuous and differentiable.
+To be invertible they must also be monotone.
+Without loss of generality we will assume they are monotone increasing.
+
+The *Iteratively Reweighted Least Squares* (IRLS) algorithm is a simple, effective method of determining the maximum likelihood estimates (mle's) of the coefficients, $\mathbf\beta$, and the dispersion parameter, $\phi$, if $\mathcal{D}$ has such a parameter.
+The iterative step, given the current $\mathbf\beta$, is to evaluate $\mathbf{\eta}=\mathbf{X\beta}$ then $\mathbf{\mu} = \mathbf{g}^{-1}(\mathbf{\eta})$ and the Jacobian $\frac{d\mathbf{\mu}}{d\mathbf{\eta}}$, which is, by definition, diagonal.
+The purpose is to determine an increment, `δ𝛃`, in the linear predictor scale from information on the residuals and the variance of the response on the response scale.
+The *working residuals*, `𝐫 = (dμ/d𝛈)⁻¹(𝐲 - 𝛍)` are the residuals `𝐲 - 𝛍` on the response scale transformed back to the linear predictor scale.
+The diagonal matrix of *working weights*, `𝐖 = (d𝛍/d𝛈)²𝒱⁻¹(𝛍)` is derived from the *variance function*, `𝒱(𝛍)`, of the distribution `𝒟`, which is the part of the variance that depends on the mean, 𝛍, only.
+The increment is the weighted least squares solution of `𝐗′𝐖𝐗 δ𝛃 = 𝐗′𝐫`.
+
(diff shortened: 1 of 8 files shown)


src/glmfit.jl
```diff
@@ -208,7 +213,7 @@ function _fit!(m::AbstractGLM, verbose::Bool, maxIter::Integer, minStepFac::Real
         linpred!(lp, p, 0)
         updateμ!(r, lp)
     end
-    devold = deviance(m)
+    devold  = deviance(m)
```
dmbates: Was this spacing change accidental?

dmbates:
This looks fine to me.  It is a very good idea.

In my experience the vast majority of glm's to be fit are Bernoulli, Binomial or Poisson distributions and the only time I see non-canonical links being used is with the Bernoulli or Binomial.  Very occasionally a non-canonical link is used with Poisson or Gamma but that is often a result of misunderstanding.  Switching to a complementary inverse link in the Bernoulli and Binomial cases when it is well-defined makes good sense.

dmbates:
Tests are showing some failures due to, I think, the assumption in predict that `linkinv` returns a scalar.

The particular case of using the complementary inverse link makes sense for cases where 0 < μ < 1 is enforced and I think that only the Bernoulli and Binomial distributions do that.  What I would suggest is creating a inverselinkorcomplement function that is used only for those distributions and defining a default
```
function linkinv(L::Link, η)
    μ, c = inverselinkorcomplement(η)
    c ? μ : one(μ) - μ
end
```

All the link functions that do not map to (0, 1) just define `linkinv` directly (sloppy terminology but I think you know what I mean).

This should never end up being called if we simply define a method for `updateμ!` that is specific to
`D<:Union{Bernoulli,Binomial}` in addition to the one with an explicit `LogitLink`.  It is still worthwhile keeping an explicit method for the `D<:Union{Bernoulli,Binomial},L<:LogitLink` case because of the cancellation for a canonical link.

I'll write this up and you can see what you think.  While I am at it, I think I will remove the calls to `Threads.@threads` in `updateμ!`.  I checked on some big models and it actually slows things down to use 4 threads.

src/glmtools.jl
```diff
@@ -9,28 +9,34 @@ type LogLink      <: Link end
...
+mueta(  ::IdentityLink, η) = one(η)
 
 linkfun(::InverseLink, μ) = inv(μ)
 linkinv(::InverseLink, η) = (inv(η), false)
 mueta(  ::InverseLink, η) = -inv(abs2(η))
 
-linkfun(::LogitLink, μ) = logit(μ)
-linkinv(::LogitLink, η) = (logistic(-abs(η)), η > 0)
+linkfun(::LogitLink, μ) = log(μ / (one(μ) - μ))
+linkinv(::LogitLink, η) = (inv(xp1(exp(-abs(η)))), η > 0)
```
andreasnoack: This should be `inv(xp1(exp(abs(η))))` and is the reason why the tests fail. However, I think we should just keep using `StatsFuns` for now. It is a dependency of `Distributions` which `GLM` depends on so it is already loaded anyway.

andreasnoack:
> The particular case of using the complementary inverse link makes sense for cases where 0 < μ < 1 is enforced and I think that only the Bernoulli and Binomial distributions do that. 

Notice that all the other `linkinv`s just return `false` in the second argument so the general `updateμ!` should still be fine although it is a bit ugly with all the ternaries. I thought about defining `linkcinv` methods for the link functions for `Binomial`/`Bernoulli` but, as you mention, then we'd have to specialize `updateμ!` for these models just for this so I decided it would be nicer to hide behind dispatch in `linkinv`. It is a bit annoying, though, that the inverse link function now returns two values.

andreasnoack:
> While I am at it, I think I will remove the calls to Threads.@threads in updateμ!. I checked on some big models and it actually slows things down to use 4 threads.

We should check and see if there is a problem with `Box`ed variable in the closure. I can take look.

I might have mentioned this before, but a thing I'd also like to look into is using a gradient based criterion for the convergence since that would save us a lot of logarithms.

dmbates:
Regarding the definition of `linkinv`.  I would propose creating
```
@compat abstract type Link01 <: Link end
```
so that `LogitLink`, `ProbitLink`, `CauchitLink` and `CloglogLink` are subtypes of `Link01`.

Methods for `linkinvcomplement` are defined for these links whereas other links have methods for `linkinv`.

I have this written now in `anj/abstol` branch.  Would you prefer that I create a new branch for this or can I commit to this branch?

src/glmfit.jl
andreasnoack: I think it was accidental

andreasnoack:
> Would you prefer that I create a new branch for this or can I commit to this branch?

No, it's fine. Let's just work on this branch.

docs/src/links.jmd
```diff
@@ -0,0 +1,204 @@
...
+The `LogitLink` is the [*canonical link*](https://en.wikipedia.org/wiki/Generalized_linear_model#Link_function) when the response is assumed to have a Bernoulli distribution.
+
+### The `ProbitLink` from the `Normal()` distribution
+
+The standard normal distribution
+```{julia;label=Normal;term=true}
+Normal()
+```
+generates the `ProbitLink`.
+Its cdf and quantile function are defined in terms of the "complementary error function", `erfc`, and its inverse, `erfcinv`, from the `StatsFuns` package.
```
andreasnoack: They are actually in `SpecialFunctions` but imported in `StatsFuns`

docs/src/links.jmd
```diff
@@ -0,0 +1,248 @@
+# Numerical considerations when fitting GLMs
+
+Before discussing computational aspects of fitting Generalized Linear Models (GLMs) it is helpful to define the linear model and see which properties are being generalized.
+
+### Properties of linear models
+
+The probability model for a linear model can be written
+\begin{equation}
```
nalimilan: Is that valid Markdown? Also, should probably break lines at 92 chars, as we do it e.g. in README.md.

ararslan: The file extension is .jmd, but I'm not sure what that is.

nalimilan: AFAICT that's a Weave.jl Markdown file, but I'm not sure what this implies in terms of syntax.

src/glmtools.jl
```diff
@@ -88,17 +228,22 @@ function devresid(::Binomial, y, μ)
...
 # Whether a dispersion parameter has to be estimated for a distribution
-dispersion_parameter(::Union{Bernoulli, Binomial, Poisson}) = false
+dispersion_parameter(::Union{Bernoulli, Binomial, Poisson})   = false
 dispersion_parameter(::Union{Gamma, Normal, InverseGaussian}) = true
 
-# Log-likelihood for an observation
+"""
+    loglik_obs(D::Distribution, y, μ, wt, ϕ)
+
+Return the log-likelihood contribution for `y` under distribution `D` with mean `μ`,
```
nalimilan: Could be worth mentioning "for a single observation" or something similar.

andreasnoack:
Superseded by #192

PR #192, opened 2017-08-09
Evaluate inverse link and its derivative simultaneously

dmbates:
I am proposing these changes in place of https://github.com/JuliaStats/GLM.jl/pull/190

During the iterations of the IRLS algorithm both the inverse link function and its derivative are evaluated.  Separate evaluations can result in a small amount of repeated evaluation which can be avoided by having both values returned by the method for `inverselink`.

The more important issue is ensuring that the evaluation of the derivative and of the variance function returns a non-zero value, because the derivative appears in the denominator of the working residual and the variance function appears in the denominator of the working weights.

It turns out that this is not a problem for most combinations of distribution and link.  When the link is canonical for the distribution the variance function is the derivative.  That is 
```julia
glmvar(D, linkinv(η)) == mueta(canonicallink(D), η)
```
and cancellation yields a simpler form for the working weights.  See the use of `iscanonical(r)` in the `updateμ!` methods for `r`.

The only remaining problematic cases are the `Bernoulli` or `Binomial` distributions with non-canonical link functions because
```julia
glmvar(::Union{Bernoulli,Binomial}, μ) = μ * (1 - μ)
```
and the value of μ may round off to 1.0 even when the product, μ * (1 - μ), can be evaluated stably from η.  `inverselink` methods for subtypes of `Link01` therefore return a 3-tuple of the mean, the derivative and the variance function.

Doing so avoids most of the clamping required in other approaches.  The only exception is the `CloglogLink`.  The inverse link in a `Link01` type is a continuous, differentiable, invertible (hence monotone) function on 𝕽 into (0,1).  Without loss of generality it can be assumed to be monotone increasing which means it corresponds to the cdf of a continuous univariate distribution with support on 𝕽.  The pdf is the derivative function.  A symmetric Link01 type has a symmetric pdf.  `ProbitLink` and `CauchitLink`, corresponding to the standard `Normal` and  `Cauchy` distributions, are symmetric.  `CloglogLink`, corresponding to the extreme-value distribution, is highly asymmetric.  However, in this approach the only clamping uses `realmin()`, not `eps()`.  (I found out the hard way that this clamping was still needed.)
```julia
function inverselink(::CloglogLink, η)
    expη = exp(η)
    μ = -expm1(-expη)
    omμ = exp(-expη)   # the complement, 1 - μ
    μ, max(realmin(μ), expη * omμ), max(realmin(μ), μ * omμ)
end
```

The Travis runs are failing on a test in `predict`.  I feel that the test is at fault, not the code.  The test fits a Binomial glm to data that are not binomial responses.  The responses are random uniform variates, not fractions of successes.  This is the moral equivalent of a least squares fit of data simulated as Xβ without any random variation.  The test failure is essentially due to round off because the test is phrased as `==` not `isapprox`.

I think it is the test that should be fixed.  The simplest fix is to replace the `==` with an approximate comparison.  I feel it would be better to go further and create a simulated value that actually corresponds to the model.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ os:
     - linux
     - osx
 julia:
-    - 0.5
     - 0.6
     - nightly
 notifications:
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ os:
 julia:
     - 0.5
     - 0.6
-    - nightly
+#    - nightly   Still a problem with StatsFuns package on nightly
 notifications:
     email: false
 sudo: false
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ os:
 julia:
     - 0.5
     - 0.6
-#    - nightly   Still a problem with StatsFuns package on nightly
+    - nightly
 notifications:
     email: false
 sudo: false
--- a/docs/mkdocs.yml
+++ b/docs/mkdocs.yml
@@ -1,0 +1,30 @@
+site_name:        GLM.jl
+repo_url:         https://github.com/JuliaStats/GLM.jl
+site_description: Fit and examine generalized linear models
+site_author:      JuliaStats
+
+theme: material
+
+extra:
+  palette:
+    primary: 'indigo'
+    accent:  'blue'
+
+extra_css:
+  - assets/Documenter.css
+
+extra_javascript:
+  - https://cdn.mathjax.org/mathjax/latest/MathJax.js?config=TeX-AMS_HTML
+  - assets/mathjaxhelper.js
+
+markdown_extensions:
+  - codehilite
+  - extra
+  - tables
+  - fenced_code
+  - mdx_math
+
+docs_dir: 'build'
+
+pages:
+- Home: index.md
--- a/src/glmtools.jl
+++ b/src/glmtools.jl
@@ -44,6 +44,13 @@ The canonical [`Link`](@ref) for [`Distributions.Gamma`](@ref) distribution, def
 type InverseLink  <: Link end
 
 """
+    InverseSquareLink
+
+The canonical [`Link`](@ref) for [`Distributions.InverseGaussian`](@ref) distribution, defined as `η = inv(abs2(μ))`.
+"""
+type InverseSquareLink  <: Link end
+
+"""
     LogitLink
 
 The canonical [`Link01`](@ref) for [`Distributions.Bernoulli`](@ref) and [`Distributions.Binomial`](@ref).
@@ -147,6 +154,14 @@ function inverselink(::InverseLink, η)
     μ, -abs2(μ), oftype(μ, NaN)
 end
 
+linkfun(::InverseSquareLink, μ) = inv(abs2(μ))
+linkinv(::InverseSquareLink, η) = inv(sqrt(η))
+mueta(::InverseSquareLink, η) = -inv(2η*sqrt(η))
+function inverselink(::InverseSquareLink, η)
+    μ = inv(sqrt(η))
+    μ, -μ / (2η), oftype(μ, NaN)
+end
+
 linkfun(::LogitLink, μ) = logit(μ)
 linkinv(::LogitLink, η) = logistic(η)
 function mueta(::LogitLink, η)
@@ -186,17 +201,19 @@ inverselink(::SqrtLink, η) = abs2(η), 2η, oftype(η, NaN)
 canonicallink(::Bernoulli) = LogitLink()
 canonicallink(::Binomial) = LogitLink()
 canonicallink(::Gamma) = InverseLink()
+canonicallink(::InverseGaussian) = InverseSquareLink()
 canonicallink(::Normal) = IdentityLink()
 canonicallink(::Poisson) = LogLink()
 
 glmvar(::Union{Bernoulli,Binomial}, μ) = μ * (1 - μ)
 glmvar(::Gamma, μ) = abs2(μ)
+glmvar(::InverseGaussian, μ) = μ^3
 glmvar(::Normal, μ) = one(μ)
 glmvar(::Poisson, μ) = μ
 
 mustart(::Bernoulli, y, wt) = (y + oftype(y, 1/2)) / 2
 mustart(::Binomial, y, wt) = (wt * y + oftype(y, 1/2)) / (wt + one(y))
-mustart(::Gamma, y, wt) = y == 0 ? oftype(y, 1/10) : y
+mustart(::Union{Gamma, InverseGaussian}, y, wt) = y == 0 ? oftype(y, 1/10) : y
 mustart(::Normal, y, wt) = y
 mustart(::Poisson, y, wt) = y + oftype(y, 1/10)
 
@@ -218,6 +235,7 @@ function devresid(::Binomial, y, μ)
     end
 end
 devresid(::Gamma, y, μ) = -2 * (log(y / μ) - (y - μ) / μ)
+devresid(::InverseGaussian, y, μ) = abs2(y - μ) / (y * abs2(μ))
 devresid(::Normal, y, μ) = abs2(y - μ)
 devresid(::Poisson, y, μ) = 2 * (xlogy(y, y / μ) - (y - μ))
 
@@ -228,6 +246,7 @@ dispersion_parameter(::Union{Gamma, Normal, InverseGaussian}) = true
 # Log-likelihood for an observation
 loglik_obs(::Bernoulli, y, μ, wt, ϕ) = wt*logpdf(Bernoulli(μ), y)
 loglik_obs(::Binomial, y, μ, wt, ϕ) = logpdf(Binomial(Int(wt), μ), Int(y*wt))
-loglik_obs(::Gamma, y, μ, wt, ϕ) = wt*logpdf(Gamma(1/ϕ, μ*ϕ), y)
+loglik_obs(::Gamma, y, μ, wt, ϕ) = wt*logpdf(Gamma(inv(ϕ), μ*ϕ), y)
+loglik_obs(::InverseGaussian, y, μ, wt, ϕ) = wt*logpdf(InverseGaussian(μ, inv(ϕ)), y)
 loglik_obs(::Normal, y, μ, wt, ϕ) = wt*logpdf(Normal(μ, sqrt(ϕ)), y)
 loglik_obs(::Poisson, y, μ, wt, ϕ) = wt*logpdf(Poisson(μ), y)
--- a/src/glmtools.jl
+++ b/src/glmtools.jl
@@ -1,17 +1,118 @@
+"""
+    Link
+
+An abstract type whose subtypes determine methods for [`linkfun`](@ref), [`linkinv`](@ref),
+[`mueta`](@ref), and [`inverselink`](@ref).
+"""
 @compat abstract type Link end     # Link types define linkfun!, linkinv!, and mueta!
 
+"""
+    CauchitLink
+
+A [`Link`](@ref) corresponding to the standard Cauchy distribution,
+[`Distributions.Cauchy`](@ref).
+"""
 type CauchitLink <: Link end
+
+"""
+    CloglogLink
+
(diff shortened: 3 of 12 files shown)


test/runtests.jl
```diff
@@ -368,27 +368,27 @@ end
...
-    
+
     ft = ftest(mod, nullmod)
     @test isapprox(ft.pval[1].v,  2.481215056713184e-8)
     # Test output
-    @test sprint(show, ftest(mod, nullmod)) == 
+    @test sprint(show, ftest(mod, nullmod)) ==
         """
                 Res. DOF DOF ΔDOF    SSR    ΔSSR      R²    ΔR²       F* p(>F)
-        Model 1       10   3      0.1283          0.9603
```
ararslan: The trailing whitespace here will need to be preserved since it's inside of a string. Below (line 391) as well.

.travis.yml
```diff
@@ -5,7 +5,7 @@ os:
 julia:
     - 0.5
     - 0.6
-    - nightly
+#    - nightly   Still a problem with StatsFuns package on nightly
```
andreasnoack: This was just fixed minutes ago when the Compat.jl PR to METADATA got merged

dmbates: Thanks.  I'll wait to see if this version gets through CI cleanly.

dmbates: On my machine the package passes its tests on nightly with the Compat PR merged.

dmbates:
@andreasnoack @ararslan Can this be merged now?

andreasnoack:
I'll try to get through this PR over the weekend.

ararslan:
Likewise

ararslan:
I was able to do a preliminary look over the changes and they look good in general. However, the code blocks in the docstrings are annotated as `jldoctest` likely won't pass as tests without an extra newline between output and the next prompt; Documenter expects `jldoctest` blocks to look identical to a REPL session. I'm also not familiar with an mkdocs-based Documenter setup so I can't really speak to the correctness of that part (I typically do the native HTML generation from Documenter) but it looks reasonable at least.

src/glmtools.jl
```diff
@@ -20,57 +180,157 @@ end
...
 linkfun(::InverseLink, μ) = inv(μ)
 linkinv(::InverseLink, η) = inv(η)
 mueta(::InverseLink, η) = -inv(abs2(η))
+function inverselink(::InverseLink, η)
+    μ = inv(η)
+    μ, -abs2(μ), oftype(μ, NaN)
+end
+
+linkfun(::InverseSquareLink, μ) = inv(abs2(μ))
+linkinv(::InverseSquareLink, η) = inv(sqrt(η))
```
andreasnoack: I'm a little confused by the theory here. In Table 2.1 of McCullagh and Nelder, they state that while the canonical link of the Inverse Gaussian is indeed `1/μ²`, the "inverse" is `(-2θ)^(-1/2)` which is not the inverse of `θ(μ)=1/μ²` but has the property that `dμ/dθ=μ³`. The current version here doesn't have that property so it appears to me that the optimization in `updateμ!` is invalid for the Inverse Gaussian with the canonical link with the current version of `linkinv `

dmbates:
@ararslan Thanks for the review.  Regarding the `jldoctest` blocks,  I used `showcompact`, which does not add a blank line when printing a vector.  I think you will find that they do pass doctest - at least on my machine they do.  Or you can just cut-and-paste those statements into the REPL.

src/glmtools.jl
dmbates: Thanks for the careful review.

I used the expressions returned by `inverse.gaussian()` in R as the basis for my code.
```julia
julia> R"print(names(fam <- inverse.gaussian()))";
 [1] "family"     "link"       "linkfun"    "linkinv"    "variance"
 [6] "dev.resids" "aic"        "mu.eta"     "initialize" "validmu"
[11] "valideta"   "simulate"

julia> R"fam$linkfun"
RCall.RObject{RCall.ClosSxp}
function (mu)
1/mu^2
<environment: namespace:stats>


julia> R"fam$linkinv"
RCall.RObject{RCall.ClosSxp}
function (eta)
1/sqrt(eta)
<environment: namespace:stats>


julia> R"fam$mu.eta"
RCall.RObject{RCall.ClosSxp}
function (eta)
-1/(2 * eta^1.5)
<environment: namespace:stats>


julia> R"fam$variance"
RCall.RObject{RCall.ClosSxp}
function (mu)
mu^3
<bytecode: 0x560013a5a328>
<environment: 0x560013a5c4a8>
```
and it looks as if you are correct - the variance function differs from `dμ/dη` by a factor of `-1/2`.  I guess this must be the case because the variance function should be positive but this link and its inverse is monotone decreasing.

Of course the same is true for `InverseLink` which is the canonical link for the `Gamma` distribution.

This raises the interesting question of how the examples using these distributions and canonical links worked in the first place.

I'll see what I can find.

dmbates: So the bad news is that my `iscanonical` methods don't work the way that I thought (I know, if I had written unit tests I would have found this out earlier). 
```jl
julia> using DataFrames, GLM

julia> clotting = DataFrame(u = log.([5,10,15,20,30,40,60,80,100]),
                            lot1 = [118,58,42,35,27,25,21,19,18])
9×2 DataFrames.DataFrame
│ Row │ u       │ lot1 │
├─────┼─────────┼──────┤
│ 1   │ 1.60944 │ 118  │
│ 2   │ 2.30259 │ 58   │
│ 3   │ 2.70805 │ 42   │
│ 4   │ 2.99573 │ 35   │
│ 5   │ 3.4012  │ 27   │
│ 6   │ 3.68888 │ 25   │
│ 7   │ 4.09434 │ 21   │
│ 8   │ 4.38203 │ 19   │
│ 9   │ 4.60517 │ 18   │

julia> gm8 = fit(GeneralizedLinearModel, @formula(lot1 ~ 1 + u), clotting, Gamma(), verbose=true)
1: 0.016729715388848446, 0.0021177181489758666
2: 0.016729715178483907, 1.2574304824954926e-8
DataFrames.DataFrameRegressionModel{GLM.GeneralizedLinearModel{GLM.GlmResp{Array{Float64,1},Distributions.Gamma{Float64},GLM.InverseLink},GLM.DensePredChol{Float64,Base.LinAlg.Cholesky{Float64,Array{Float64,2}}}},Array{Float64,2}}

Formula: lot1 ~ 1 + u

Coefficients:
               Estimate   Std.Error  z value Pr(>|z|)
(Intercept)  -0.0165544 0.000927542 -17.8476   <1e-70
u             0.0153431 0.000414958  36.9751   <1e-99


julia> gm8.model.rr
GLM.GlmResp{Array{Float64,1},Distributions.Gamma{Float64},GLM.InverseLink}([118.0, 58.0, 42.0, 35.0, 27.0, 25.0, 21.0, 19.0, 18.0], Distributions.Gamma{Float64}(α=1.0, θ=1.0), [0.00160669, 0.00746689, 0.00240188, 0.000843898, 0.00147963, 1.23783e-6, 0.000823452, 0.00141054, 0.000695503], [0.00813941, 0.0187744, 0.0249955, 0.0294095, 0.0356306, 0.0400445, 0.0462656, 0.0506796, 0.0541033], [122.859, 53.2639, 40.0071, 34.0026, 28.0658, 24.9722, 21.6143, 19.7318, 18.4832], Float64[], [1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0], [15094.3, 2837.04, 1600.57, 1156.18, 787.688, 623.611, 467.179, 389.345, 341.628], [0.000321911, -0.00166938, -0.0012451, -0.000862636, 0.00135305, -4.45692e-5, 0.00131496, 0.00187963, 0.00141432])

julia> GLM.iscanonical(gm8.model.rr)
false
```

The good news is that the apparently redundant calculations in the working weights are not being short circuited and the desired answer was obtained.

I still need to figure out why the methods for `iscanonical` weren't providing the answers I thought they would and how to properly evaluate the working weights.

dmbates: It looks as if I need to define the methods as
```jl
julia> iscanonical{V,T}(::GlmResp{V,Gamma{T},InverseLink}) = true
iscanonical (generic function with 1 method)

julia> iscanonical(gm8.model.rr)
true
```

andreasnoack: Ah yes. I've made that mistake many times. There is a slightly more convenient syntax for this now. You should be able to do `iscanonical(::GlmResp{<:Any,<:Gamma},InverseLink}) = true`

dmbates: I have pushed a change in the method definitions but do not define `iscanonical` to be true for the `Gamma/InverseLink` and `InverseGaussian/InverseSquareLink` combinations.

dmbates: These changes have a small but reproducible impact on the performance for important cases like logistic regression
```sh
bates@thin206:~/.julia/v0.6/GLM⟫ julia perf/glm.jl  # prior to change
  5.074595 seconds (158 allocations: 1.136 GiB, 2.63% gc time)
bates@thin206:~/.julia/v0.6/GLM⟫ julia perf/glm.jl
  5.036847 seconds (158 allocations: 1.136 GiB, 2.65% gc time)
bates@thin206:~/.julia/v0.6/GLM⟫ julia perf/glm.jl
  5.038499 seconds (158 allocations: 1.136 GiB, 2.68% gc time)
```

dmbates: @andreasnoack I am beginning to think that it is not worthwhile supporting v0.5 any more. I keep needing to fiddle with the `iscanonical` method definitions to get a version that will work with Julia v0.5

Would it be okay if I bumped the `REQUIRE` file to `julia 0.6` and amended the `.travis.yml` file accordingly?  We could also do a femtocleaner run after that.

ararslan:
Ah, sorry for the oversight. I forgot that `showcompact` omits the newline.

src/glmtools.jl
andreasnoack: Yes. Let's just drop 0.5 support.

dmbates:
We're green.  @andreasnoack could you do the honors, please?

andreasnoack:
Am I right that there isn't a (non-perf) test for fitting an `InverseGaussian`?

src/glmtools.jl
andreasnoack: Ok, I've looked more carefully at the inverse Gaussian case now. The thing is that there is a bit of freedom in how you can choose in the canonical parameters in the inverse Gaussian model. You can e.g. set `θ = 1/μ²` or `θ = -1/(2μ²)` which have different signs. The former choice implies a negative variance function but therefore also a negative dispersion. The latter version has positive variance function and dispersion but `θ<0`.

However, it seems that you can scale the variance function with any (non-zero) constant and it will still work since it will end up as the same factor in the score and the information so it will cancel in the update. When the variance is calculated, the estimation of the dispersion will capture the "wrong" factor. Bottomline is that a fast version should also work just fine even if we keep R's choices of functions. However, it is still a bit confusing that the functions don't follow the theory here.

andreasnoack: We can always add the optimization later. It doesn't have to be in this PR.

dmbates: I changed the function name to `cancancel` and added some explanation in the documentation.

I also added tests for it and for fitting an `InverseGaussian` model.

andreasnoack: I actually still think I prefer `iscanonical`. Do you think it is better to avoid "canonical"?

dmbates: A link can be the canonical link for a distribution but not provide for cancellation of terms.  E.g. `Gamma/InverseLink` and `InverseGaussian/InverseSquareLink`  I would say that `iscanonical` should return `true` for those cases but `cancancel` returns `false`.

The `cancancel` function is only used in the evaluation of the working weights in the `updateμ!` methods.

docs/mkdocs.yml
```diff
@@ -0,0 +1,30 @@
+site_name:        GLM.jl
```
nalimilan: This file isn't needed nor used with the standard setup where the native HTML backend is used. See for example in CategoricalArrays: https://github.com/JuliaData/CategoricalArrays.jl/blob/master/docs/make.jl

Issue #186, opened 2017-07-15
Path towards GLMs with fweights, pweights, and aweights


jeffwong:
This ticket is related to this [discourse discussion](https://discourse.julialang.org/t/usage-of-different-types-of-weights/4780/13)

From what I can tell

1) We need to expand [LmResp](https://github.com/JuliaStats/GLM.jl/blob/master/src/lm.jl#L4) and [GLMResp](https://github.com/JuliaStats/GLM.jl/blob/master/src/glmfit.jl#L18) to not only contain wts, but also contain fweights, pweights, and aweights. We should allow a user to pass in multiple types of weights like both fweights and pweights. 

2) We will need to have a function that consolidates fweights, pweights, and aweights into a single vector that can be used in a more classic setting like weighted maximum likelihood. In MLE, only the relative weighting of the weights matter, not the absolute value. If someone is combining fweights and pweights I think this can be consolidated through `wts = fweights * pweights`. Second opinion welcome here!

3) The [MLE](https://github.com/JuliaStats/GLM.jl/blob/master/src/lm.jl#L68) functions already accommodate a weight vector, so nothing needs to be done here if 2) is done

4) The [nobs](https://github.com/JuliaStats/GLM.jl/blob/master/src/linpred.jl#L152) function currently returns the sum of wts, if present. This is the right behavior for fweights. For pweights and aweights, I think we need to return the number of rows in the LinPred object.

5) We will need to adapt the [vcov](https://github.com/JuliaStats/GLM.jl/blob/128024c71ef00d7c3334e28d043348ec38e8d8eb/src/linpred.jl#L119) function to return the covariance matrix according to the types of weights used. Here is a [reference](http://www.parisschoolofeconomics.eu/docs/dupraz-yannick/using-weights-in-stata(1).pdf)

ararslan:
As a reference to other readers, see also https://github.com/JuliaStats/StatsBase.jl/issues/283

nalimilan:
Thanks for writing this. I agree in general. Just a few remarks:

> We need to expand LmResp and GLMResp to not only contain wts, but also contain fweights, pweights, and aweights. We should allow a user to pass in multiple types of weights like both fweights and pweights.

Do we actually need to store the three vectors? Maybe we could only store one vector, after adjusting it to take into account the other (if they were specified), as in point 2?

[...]

PR #194, opened 2017-08-13
Draft of fweights, pweights, aweights for glms

jeffwong:
This is a code outline for https://github.com/JuliaStats/GLM.jl/issues/186. I am new to contributing to Julia, so please feel free to leave feedback on how this feature should be implemented

### DIFF
--- a/src/GLM.jl
+++ b/src/GLM.jl
@@ -14,6 +14,7 @@ module GLM
     import Base: (\), cholfact, convert, cor, show, size
     import StatsBase: coef, coeftable, confint, deviance, nulldeviance, dof, dof_residual, loglikelihood, nullloglikelihood, nobs, stderr, vcov, residuals, predict, fit, model_response, r2, r², adjr2, adjr², PValue
     import StatsFuns: xlogy
+    import StatsBase: FrequencyWeights, ProbabilityWeights, AnalyticWeights
     export coef, coeftable, confint, deviance, nulldeviance, dof, dof_residual, loglikelihood, nobs, stderr, vcov, residuals, predict, fit, fit!, model_response, r2, r², adjr2, adjr²
 
     export                              # types
--- a/src/glmfit.jl
+++ b/src/glmfit.jl
@@ -18,13 +18,19 @@ immutable GlmResp{V<:FPVector,D<:UnivariateDistribution,L<:Link} <: ModResp
     offset::V
     "`wts:` prior case weights.  Can be of length 0."
     wts::V
+
+    fweights::FrequencyWeights
+    pweights::ProbabilityWeights
+    aweights::AnalyticWeights
+
     "`wrkwt`: working case weights for the Iteratively Reweighted Least Squares (IRLS) algorithm"
     wrkwt::V
     "`wrkresid`: working residuals for IRLS"
     wrkresid::V
 end
 
-function GlmResp{V<:FPVector, D, L}(y::V, d::D, l::L, η::V, μ::V, off::V, wts::V)
+function GlmResp{V<:FPVector, D, L}(y::V, d::D, l::L, η::V, μ::V, off::V, wts::V,
+    fweights::FrequencyWeights, pweights::ProbabilityWeights, aweights::AnalyticWeights)
     if d == Binomial()
         for yy in y
             0. <= yy <= 1. || throw(ArgumentError("$yy in y is not in [0,1]"))
@@ -39,7 +45,8 @@ function GlmResp{V<:FPVector, D, L}(y::V, d::D, l::L, η::V, μ::V, off::V, wts:
         "lengths of η, μ, y and wts ($nη, $nμ, $(length(wts)), $n) are not equal"))
     lo = length(off)
     lo == 0 || lo == n || error("offset must have length $n or length 0")
-    res = GlmResp{V,D,L}(y, d, similar(y), η, μ, off, wts, similar(y), similar(y))
+    res = GlmResp{V,D,L}(y, d, similar(y), η, μ, off, wts, similar(y), similar(y),
+                         fweights, pweights, aweights)
     updateμ!(res, η)
     res
 end
--- a/src/glmfit.jl
+++ b/src/glmfit.jl
@@ -1,3 +1,4 @@
+# TODO: expand weights to include fweights, pweights, aweights
 """
     GlmResp
 
--- a/src/glmfit.jl
+++ b/src/glmfit.jl
@@ -45,7 +45,9 @@ function GlmResp{V<:FPVector, D, L}(y::V, d::D, l::L, η::V, μ::V, off::V, wts:
         "lengths of η, μ, y and wts ($nη, $nμ, $(length(wts)), $n) are not equal"))
     lo = length(off)
     lo == 0 || lo == n || error("offset must have length $n or length 0")
-    res = GlmResp{V,D,L}(y, d, similar(y), η, μ, off, wts, similar(y), similar(y),
+    res = GlmResp{V,D,L}(y, d, similar(y), η, μ, off,
+                         normalizeWeights(fweights, pweights, aweights),
+                         similar(y), similar(y),
                          fweights, pweights, aweights)
     updateμ!(res, η)
     res
--- a/src/glmtools.jl
+++ b/src/glmtools.jl
@@ -102,5 +102,16 @@ loglik_obs(::Poisson, y, μ, wt, ϕ) = wt*logpdf(Poisson(μ), y)
 # TODO: combine fweights, pweights, aweights
 function normalizeWeights(fweights::FrequencyWeights, pweights::ProbabilityWeights,
                           aweights::AnalyticWeights)
-
+    nf = length(fweights)
+    np = length(pweights)
+    na = length(aweights)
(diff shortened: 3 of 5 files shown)


src/linpred.jl
```diff
@@ -116,6 +116,8 @@ Base.cholfact!{T}(p::SparsePredChol{T}) = p.chol
 
 invchol(x::DensePred) = inv(cholfact!(x))
 invchol(x::SparsePredChol) = cholfact!(x) \ eye(size(x.X, 2))
+# TODO: support fweights, pweights, and aweights
+# This vcov represents σ^2  X'X
 vcov(x::LinPredModel) = scale!(invchol(x.pp), dispersion(x, true))
```
jeffwong: This needs to be updated to `invchol(x.pp * sqrt(wts))`

jeffwong: Need to check dispersion(x, true) will use nobs, which will use the right behavior for fweights vs pweights

src/glmfit.jl
```diff
@@ -17,13 +18,19 @@ immutable GlmResp{V<:FPVector,D<:UnivariateDistribution,L<:Link} <: ModResp
...
+    aweights::AnalyticWeights
+
     "`wrkwt`: working case weights for the Iteratively Reweighted Least Squares (IRLS) algorithm"
     wrkwt::V
     "`wrkresid`: working residuals for IRLS"
     wrkresid::V
 end
 
-function GlmResp{V<:FPVector, D, L}(y::V, d::D, l::L, η::V, μ::V, off::V, wts::V)
+function GlmResp{V<:FPVector, D, L}(y::V, d::D, l::L, η::V, μ::V, off::V, wts::V,
```
jeffwong: Maybe the constructor doesn't need a wts anymore, just fweights, pweights and aweights

src/glmtools.jl
```diff
@@ -98,3 +98,22 @@ loglik_obs(::Binomial, y, μ, wt, ϕ) = logpdf(Binomial(Int(wt), μ), Int(y*wt))
 loglik_obs(::Gamma, y, μ, wt, ϕ) = wt*logpdf(Gamma(1/ϕ, μ*ϕ), y)
 loglik_obs(::Normal, y, μ, wt, ϕ) = wt*logpdf(Normal(μ, sqrt(ϕ)), y)
 loglik_obs(::Poisson, y, μ, wt, ϕ) = wt*logpdf(Poisson(μ), y)
+
+# TODO: combine fweights, pweights, aweights
+function normalizeWeights(fweights::FrequencyWeights, pweights::ProbabilityWeights,
```
ararslan: Camel case isn't used in Base Julia and the JuliaStats packages tend to match the Base style as closely as possible. For function names, typically it's all lowercase with no separation if the name isn't too long, otherwise snake case. For this function, I'd probably go with `normalize_weights`.

src/glmtools.jl
```diff
@@ -98,3 +98,22 @@ loglik_obs(::Binomial, y, μ, wt, ϕ) = logpdf(Binomial(Int(wt), μ), Int(y*wt))
...
+                          aweights::AnalyticWeights)
+    nf = length(fweights)
+    np = length(pweights)
+    na = length(aweights)
+    if (nf != 0 && np == 0 && na == 0)
+        return fweights |> collect
```
ararslan: Another minor stylistic point, but it's generally better to use standard calling conventions rather than piping. In this case you could actually just assign the result of the conditional to a variable (e.g. `result = if ...`) then return `collect(result)`

src/linpred.jl
```diff
@@ -150,10 +152,11 @@ For linear and generalized linear models, returns the number of rows, or,
 when prior weights are specified, the sum of weights.
 """
 function nobs(obj::LinPredModel)
-    if isempty(obj.rr.wts)
+    if (length(obj.rr.fweights) == 0)
```
ararslan: Is `isempty` not defined for the weight objects in StatsBase?

src/lm.jl
```diff
@@ -1,20 +1,35 @@
...
+        ll = length(aweights)
+        ll == 0 || ll == n || error("length of aweights is $ll, must be $n or 0")
+        new{V}(mu, off, normalizeWeights(fweights, pweights, aweights), y,
+               fweights, pweights, aweights)
     end
 end
 convert{V<:FPVector}(::Type{LmResp{V}}, y::V) =
-    LmResp{V}(zeros(y), similar(y, 0), similar(y, 0), y)
+    LmResp{V}(zeros(y), similar(y, 0), similar(y, 0), y,
+    similar(y, 0), similar(y, 0), similar(y, 0))
```
ararslan: Should be aligned vertically with the arguments above it

src/lm.jl
```diff
@@ -1,20 +1,35 @@
+# TODO: expand weights to include fweights, pweights, aweights
+# Example LmResp{Array{Float64,1}}(randn(5), randn(5), randn(5), randn(5))
 type LmResp{V<:FPVector} <: ModResp  # response in a linear model
     mu::V                                  # mean response
     offset::V                              # offset added to linear predictor (may have length 0)
     wts::V                                 # prior weights (may have length 0)
+    fweights::FrequencyWeights
+    pweights::ProbabilityWeights
+    aweights::AnalyticWeights
```
ararslan: I'm not sure we should carry around three different types of weights in the response object. Perhaps just a single `AbstractWeights`?

src/GLM.jl
```diff
@@ -14,6 +14,7 @@ module GLM
     import Base: (\), cholfact, convert, cor, show, size
     import StatsBase: coef, coeftable, confint, deviance, nulldeviance, dof, dof_residual, loglikelihood, nullloglikelihood, nobs, stderr, vcov, residuals, predict, fit, model_response, r2, r², adjr2, adjr², PValue
     import StatsFuns: xlogy
+    import StatsBase: FrequencyWeights, ProbabilityWeights, AnalyticWeights
```
ararslan: You'll need to bump the lower bound on StatsBase in the REQUIRE file to the version where these types were introduced.

src/glmtools.jl
```diff
@@ -98,3 +98,22 @@ loglik_obs(::Binomial, y, μ, wt, ϕ) = logpdf(Binomial(Int(wt), μ), Int(y*wt))
...
+    elseif (nf == 0 && np != 0 && na == 0)
+        return pweights |> collect
+    elseif (nf == 0 && np == 0 && na != 0)
+        return aweights |> collect
+    elseif (nf != 0 && np != 0)
+        return (fweights .* pweights) |> collect
+    else
+        return zeros(Float64, 0)
```
nalimilan: We should probably handle this more efficiently by returning an empty vector. The existing code uses that trick.

src/lm.jl
nalimilan: Yes, we should store the weights internally as a single vector (maybe keep the existing `wts` field), and only keep track of the necessary additional information like the actual number of observations (which is either the sum of weights or the vector length depending on the weights type).

src/glmtools.jl
dmbates: Actually the empty vector gets overridden at some point by `ones(n)`.  I haven't checked exactly where or why it was done.  I suspect it was so that a quantity could be normalized by `sum(weights)`.

nalimilan: OK. Looks like a good use case for a `UnitWeights` or `Ones` type which would implement the `AbstractArray` interface and always return `1` (https://github.com/JuliaStats/StatsBase.jl/issues/135). But that's not a requirement for a first implementation.

jeffwong:
I think if we only supported Frequency Weights and Probability Weights this PR would be a lot easier for me. I am completely unfamiliar with Analytic Weights, and how they affect formulas for covariance of the glm parameters. It would also mean that we could easily simplify the API by only passing in 1 vector of `weights`. If someone needed to combine Frequency and Probability together, they could do that using `fweight * pweight / sum(fweight * pweight) * sum(fweight)`. This could then be cast as a Frequency Weight: it has the relative weighting that the pweight brings, and the sum of this new vector still sums to `sum(fweight)`, protecting the use of `nobs` in the package

jeffwong:
Retrospectively, by limiting the scope to only Fweights and Pweights, I think we only needed a way to define `nobs` as a function of the weight type. Even if someone wanted to combine Fweights and Pweights, it would ultimately still behave like an Fweight. As long as the cov(parameters) was already based on $X' wts X$ instead of $X'X$ the only modification we would need is to `nobs`

nalimilan:
Sure, let's go with the simpler version for now. I don't know of other software accepting multiple types of weights at the same time, and we can always add support for analytic weights later.

sergiocorreia:
@jeffwong about your comment on Analytic Weights:

They represent multiple observations, like fweights, but instead of representing multiple *identical* obs., they represent distinct observations that were averaged out. This happens with variables such as "average grade by classroom", "accident rate per state", etc. Since groups with more individuals have less noisy means, aweights lead to heteroskedasticity.

That said, aweights are implemented as *literally* just pweights with robust standard errors, so if the code supports both then it also supports aweights

Nosferican:
What's the status on this?

nalimilan:
Superseded by https://github.com/JuliaStats/GLM.jl/pull/487.

Issue #195, opened 2017-09-20
Unexpected behavior of `issubmodel`


pbastide:
Hi @LewisHein, @ararslan and @nalimilan.

I noticed two unexpected behaviors in function `issubmodel`, called by `ftest`:
* Due to a typo, only the first predictor of mod2 were tested.
* I'm using GLM on data normalized by a Cholesky transform. This leads to rounding problems with the predictors. Sometimes, two predictors are almost equal, up to numerical approximation, but right now are considered distinct by `issubmodel`.

I'm about to make two pr referencing this issue, and that should fix the two problems pointed above. These pr contain some added tests, that can work as MWE for these issues: the tests fail before the commit, but pass after.

I think the first one is rather straightforward, but you might want to make changes in the second one. That is why I made two separate pr.

Thanks !

pbastide:
I'm closing the issue, since the two PR have been merged. Thanks !

PR #197, opened 2017-09-20
Allow for atol=1e-10 for predictor tests in `issubmodel`.

pbastide:
When testing for equal predictors, allow for some numerical tolerance.

The added test can work as a MWE: it fails with the current version of GLM, but passes after the commit.

See #195.


andreasnoack:
I think we should consider changing this function more fundamentally since it is too restrictive. Any linear hypothesis should be fine, e.g. if `Xbig = [x1 x2 x3]` then `Xsmall = [x1-x2 x1-x3]` is a valid submodel so we should probably check that `min norm(Xbig*B - Xsmall) ≈ 0`.

nalimilan:
I'll leave the math to @andreasnoack, but regarding the fix in this PR, I wonder in what concrete cases this kind of situation can happen? I think the function was designed with a very limited goal in mind: checking that the predictors were exactly the same, i.e. they come from the same columns of the same data set.

pbastide:
Using `ftest` for [PhyloNetworks](https://github.com/crsl4/PhyloNetworks.jl), I might be able to give you two practical examples where the change proposed here and by @andreasnoack could be useful.
* In PhyloNetworks, we do some "phylogenetic regression", where the data has a known variance structure. Hence, we need to do a Cholesky transform before feeding it to GLM. This is where the example I added in the test comes from. Here, it introduces some numerical issues, that would prevent us from using `issubmodel`, as the effectively equal predictors would be considered different, just because of numerical approximations.
* We use `ftest` to test homogeneous versus heterogeneous models. For instance, you have two predictors, say A and B, and you want to test:
H1: "one single coefficient for both A and B" vs H2: "A and B each have their own coefficients".
Denoting S = A+B the sum of the two predictors, they correspond to the following formulas:
H1: "Y ~ S" vs H2: "Y ~ A + B".
These two models are nested, but if you write the formulas this way, then with the current version of `issubmodel` will not recognize it, and prevent you from doing the test. So currently you have to use the formulas:
H1: "Y ~ S" vs H2: "Y ~ S + A"
They are indeed equivalent to the previous ones, but maybe less natural to use.

nalimilan:
Thanks. I'm not familiar with these use cases, but if there's a robust way of checking that models are nested that's fine with me. I'm not sure how to choose the tolerance threshold, though.

pbastide:
I agree that choosing the tolerance might be tricky. I added an optional keyword parameter to give the control to the user.
About @andreasnoack proposition, all I can think of is something like:
```julia
isapprox(norm(pred2 * (pred2 \ pred1) - pred1), 0, atol=atol)
```
to test whether pred1 is a submodel of pred2. But it's a naive approach, and there are probably better ways to do that ?

andreasnoack:
Probably better to use an elementwise norm here (cheaper) and it will  be slightly more precise and efficient if you avoid computing the least squares coefficients. Finally, I think it is better with a relative difference measure so maybe something like
```
vecnorm(view(qrfact(pred2)[:Q]'pred1, npred2 + 1:size(pred2, 1), :)) < rtol*vecnorm(pred1)
```
The left hand side is
image

pbastide:
Thank you for the explanations, both mathematical and computational !
I tried to implement your idea. For `rtol`, I used the same default than in [`isapprox`](https://docs.julialang.org/en/latest/stdlib/math/#Base.isapprox). I kept a `atol` parameter (default to 0) in addition for flexibility (but it might not be necessary ?).
I also merged the other PR with the typo fix in `npreds2` for convenience.

andreasnoack:
Great. Thanks for fixing this.

pbastide:
Thank you, I'm glad if it helped.
Would it be possible to issue a release with this fix, so that we can use it in PhyloNetworks ? (That is what motivated my changes in the first place.)

andreasnoack:
See https://github.com/JuliaLang/METADATA.jl/pull/11321

pbastide:
Thank you !

PR #198, opened 2017-10-14
Fix stderr(OLS) output in OLS example of Readme

BenjaminBorn:


BenjaminBorn:
No idea, the regression output two lines above shows the correct values.

ararslan:
Well this is amusing.

PR #201, opened 2017-11-26
StatsModels

dmbates:
Switch to `StatsModels` v0.1.0

Tests now use `DataFrames` v0.11.0, `CSV` and `CategoricalArrays`.  Because of the change from `DataArrays::pool` to `CategoricalArrays::categorical` the order of the elements in the pool changed.  This caused with changes in the coefficients and standard errors in some models used for tests.


src/GLM.jl
```diff
@@ -1,31 +1,41 @@
...
 module GLM
-    using Reexport
-    @reexport using Distributions
+    using Distributions
     using Base.LinAlg.LAPACK: potrf!, potrs!
     using Base.LinAlg.BLAS: gemm!, gemv!
     using Base.LinAlg: QRCompactWY, Cholesky, BlasReal
     using StatsBase: StatsBase, CoefTable, StatisticalModel, RegressionModel
     using StatsFuns: logit, logistic
+    using StatsModels: @formula, Formula, ModelFrame, ModelMatrix
```
nalimilan: I wonder whether we should reexport StatsModels for now, to avoid the need to call `using StatsModels` explicitly.

test/runtests.jl
```diff
@@ -28,18 +29,18 @@ form = DataFrame(Any[[0.1,0.3,0.5,0.6,0.7,0.9],[0.086,0.269,0.446,0.538,0.626,0.
     @test isapprox(r²(lm1), 0.9990466748057584)
     @test adjr²(lm1) == adjr2(lm1)
     @test isapprox(adjr²(lm1), 0.998808343507198)
-    @test isapprox(aic(lm1), -36.409684288095946)
-    @test isapprox(aicc(lm1), -24.409684288095946)
-    @test isapprox(bic(lm1), -37.03440588041178)
+    @test isapprox(StatsBase.aic(lm1), -36.409684288095946)
```
nalimilan: `StatsBase.` shouldn't be needed AFAIK?

test/runtests.jl
```diff
@@ -69,8 +70,8 @@ admit[:rank] = pool(admit[:rank])
         @test isapprox(aicc(gm2), 470.7312329339146)
         @test isapprox(bic(gm2), 494.4662797585473)
         @test isapprox(coef(gm2),
-            [-3.9899786606380734, 0.0022644256521549043, 0.8040374535155766,
-            -0.6754428594116577, -1.3402038117481079,-1.5514636444657492])
+            [-5.3301824723861895, 0.002264425652154903, 0.8040374535155792,
```
nalimilan: I've seen your comment about why coefficients change, but that's still not clear to me. The order of levels should be the same as with PDAs AFAIK. I guess that's rather a change in StatsModels? Anyway, calling `levels!` should be enough to get the same behavior as before.

test/runtests.jl
```diff
@@ -129,8 +130,7 @@ end
 end
 
 ## Example with offsets from Venables & Ripley (2002, p.189)
-anorexia = readtable(joinpath(glm_datadir, "anorexia.csv.gz"))
-anorexia[:Treat] = pool(anorexia[:Treat])
+anorexia = CSV.read(joinpath(glm_datadir, "anorexia.csv"))
```
nalimilan: Maybe pass `categorical=true`, that way we're safe if the default changes in a future version (as it's been discussed).

test/runtests.jl
dmbates: Thanks for catching that.  It's vestigial.  Initially I wasn't including `StatsBase` and using fully qualified names instead.

src/GLM.jl
dmbates: I really dislike reexporting entire packages.  Often I will check
```jl
whos(PkgName)
```
to see what the functions and types in a package are and I hate it when everything in, say, `Distributions` is in that list.

nalimilan: Yes, I'm not sure what's the best approach here. @ararslan @kleinschmidt Ideas?

kleinschmidt: I don't see the trouble of having `using StatsModels` inside the package; if GLM users want to mess about with statsmodels stuff then I think it's reasonable to have them explicitly ask for it, right?

nalimilan: Yeah, my concern is about people who just want to fit a GLM. It would be too bad to require them to run `using StatsModels`.

kleinschmidt: They'd just need `@formula` for that though, right?  I'm not really opposed to re-exporting everything from statsmodels.

nalimilan: For the basic use case, yes. But then it sounds natural to allow using contrast types without importing StatsModels explicitly either.

A possibility would be to set up the Stats meta-package we've been talking about for some time, and reexport StatsModels from it. At least it would be simple to tell people who want to do stats to  always call  `using Stats`.

ararslan: I kind of assumed when we were making StatsModels that it would be like StatsBase in that it would typically be extended and reexported. I agree that requiring that users remember to do `using StatsModels` before being able to fit a GLM with a `@formula` seems unfortunate. How did we handle this previously with formulas, which lived in DataFrames?

kleinschmidt: The stats models stuff in dataframes provided a high-level interface to the matrix+vector type fitting methods in GLM.  AFAIK GLM didn't depend on data frames...

kleinschmidt: Ah and as far as I can tell none of the changes to GLM require StatsModels...just for tests.  I think a "batteries" package is probably needed here, something that gives you DataFrames, StatsModels, CSV, maybe Feather.  Previously you'd get the formula methods by `using DataFrames, GLM`.  So from that perspective it's not crazy to have people `using DataFrames, StatsModels, GLM`, or `using Stats, GLM`.

ararslan: > or `using Stats, GLM`.

Though I would think that GLM would be included in said meta-package.

dmbates: I misunderstood the issue @nalimilan was getting at with his original comment.  I hadn't been thinking about setting contrasts, etc.

On looking at this again, I would be fine with replacing this line with `using StatsModels` and exporting all the symbols currently in `StatsModels`.  I prefer not to use the `Reexport` package myself.  I would rather have the default be not to reexport newly added generics and types.  When using `@reexport` you commit yourself to export any new symbols exported by `StatsModels`.

kleinschmidt: I suppose that is StatsModels is totally divorced from DataFrames it may be lightweight enough to have as a dependency; but right now requiring it for GLM essentially also requires DataFrames which I think is not desirable, right?

ararslan: That's a good point. You could just manually reexport any functions you've extended plus any convenience things like `@formula`.

test/runtests.jl
dmbates: According to the documentation the default value of `categorical` is `true`.  The reason for the explicit call here is that the levels are integers.  I could use an explicit `types` argument in the call to `CSV.read` but I don't think that would gain much.

src/GLM.jl
ararslan: > but right now requiring it for GLM essentially also requires DataFrames which I think is not desirable, right?

Another good point. The original plan of having an abstract table interface seems to have fallen by the wayside, so what's the current thinking for having a DataFrames dependency in StatsModels? Seems less than ideal but there are inherently some table-specific things in there.

kleinschmidt: Yeah dealing with tabular data is kind of the whole raison d'etre of StatsModels. I think it should be doable to use something more lightweight (like the column store from DataStreams); the only thing that's needed is column names, types, and something like a view that skips rows with missing values.

test/runtests.jl
dmbates: The changes are with respect to the contrasts chosen for the levels of `admit[:rank]`, which is read as an integer then converted to categorical.  I think the `pool` conversion from a vector of integers to a PooledDataArray just uses the unique levels without sorting them whereas the `categorical` conversion does sort.  You would know better than I.

In any case, the contrasts chosen here are the sensible ones.  I would prefer not to enforce bug-for-bug compatibility with the last version.

test/runtests.jl
nalimilan: OK, I hadn't realized that `categorical=true` just means that the result will be categorical if the proportion of unique values is below a certain threshold. I guess that's OK. (I think we would have to pass `CategoricalValue` as the type to be certain we get a categorical vector.)

src/GLM.jl
nalimilan: The situation isn't worse than before at least. It will improve as soon as we create that table interface.

Stats would probably include GLM, but we need to find a general pattern which will also work for modeling packages which would not be in Stats. Honestly, I'd find it a bit ridiculous to tell people that if they do `using MyModeling package`, they also need to do `using StatsModels`. Re-exporting sounds more logical. Ideally at some point there will be a way to load a package automatically when loading another one, without reexporting.

kleinschmidt: The situation may not be worse but this PR changes the relationship between GLM and StatsModels. Before, if you just did `using GLM` you wouldn't get DataFrames or any of the StatsModeling functionality; you'd need to also do `using DataFrames`.  I'm fine with this PR pulling in StatsModels and DataFrames whenever I do `using GLM` now, because I'm never fitting a GLM to non-tabular data anyway.  But there may be people out there who don't want the added dependency on DataFrames and could be surprised by it when they update.  Just something to consider.

nalimilan: > But there may be people out there who don't want the added dependency on DataFrames and could be surprised by it when they update. Just something to consider.

Yes, hopefully we'll get rid of the dependency on DataFrames at some point. Until then I'd say it's fine if those people need to install DataFrames.

nalimilan: The question was, isn't it weird that you need to call `using StatsModels` to adjust contrasts?

dmbates: In that case I think we should go with your original suggestion of re-exporting the symbols from `StatsModels`, as there aren't very many.  I will amend the PR accordingly.

dmbates:
I guess I don't understand how the `@reexport` macro works.  Notice below that generics and macros, like `@formula` and `coefnames` are exported as `StatsModels.#@formula`, etc. but the types `ModelMatrix` and `ModelFrame` are exported as `GLM.#ModelMatrix` and `GLM.#ModelFrame`.  A package that uses both `GLM` and `StatsModels` encounters a name conflict.

```julia
julia> whos(GLM)
                      @formula      0 bytes  StatsModels.#@formula
             AbstractContrasts     92 bytes  DataType
                     Bernoulli     40 bytes  UnionAll
                      Binomial     40 bytes  UnionAll
                   CauchitLink     92 bytes  DataType
                   CloglogLink     92 bytes  DataType
               ContrastsCoding    176 bytes  DataType
                     DensePred     92 bytes  DataType
                 DensePredChol     96 bytes  UnionAll
                   DensePredQR     56 bytes  UnionAll
                   DummyCoding    124 bytes  DataType
                 EffectsCoding    124 bytes  DataType
                       Formula    188 bytes  DataType
                           GLM    128 KB     Module
                         Gamma     40 bytes  UnionAll
        GeneralizedLinearModel    280 bytes  UnionAll
                       GlmResp    200 bytes  UnionAll
                 HelmertCoding    124 bytes  DataType
                  IdentityLink     92 bytes  DataType
               InverseGaussian     40 bytes  UnionAll
                   InverseLink     92 bytes  DataType
             InverseSquareLink     92 bytes  DataType
                       LinPred     92 bytes  DataType
                  LinPredModel     92 bytes  DataType
                   LinearModel    160 bytes  UnionAll
                          Link     92 bytes  DataType
                        LmResp     80 bytes  UnionAll
                       LogLink     92 bytes  DataType
                     LogitLink     92 bytes  DataType
                    ModelFrame      0 bytes  GLM.#ModelFrame
                   ModelMatrix      0 bytes  GLM.#ModelMatrix
                        Normal     40 bytes  UnionAll
                       Poisson     40 bytes  UnionAll
                    ProbitLink     92 bytes  DataType
                      SqrtLink     92 bytes  DataType
                   StatsModels    115 KB     Module
                         adjr2      0 bytes  StatsBase.#adjr2
                         adjr²      0 bytes  StatsBase.#adjr2
                 canonicallink      0 bytes  GLM.#canonicallink
                          coef      0 bytes  StatsBase.#coef
                     coefnames      0 bytes  StatsModels.#coefnames
                     coeftable      0 bytes  StatsBase.#coeftable
                       confint      0 bytes  StatsBase.#confint
                      delbeta!      0 bytes  GLM.#delbeta!
                      deviance      0 bytes  StatsBase.#deviance
                      devresid      0 bytes  GLM.#devresid
                           dof      0 bytes  StatsBase.#dof
                  dof_residual      0 bytes  StatsBase.#dof_residual
                      dropterm      0 bytes  StatsModels.#dropterm
                           fit      0 bytes  StatsBase.#fit
                          fit!      0 bytes  StatsBase.#fit!
                       formula      0 bytes  GLM.#formula
                         ftest      0 bytes  GLM.#ftest
                           glm      0 bytes  GLM.#glm
                        glmvar      0 bytes  GLM.#glmvar
                   inverselink      0 bytes  GLM.#inverselink
                       linkfun      0 bytes  GLM.#linkfun
                       linkinv      0 bytes  GLM.#linkinv
                       linpred      0 bytes  GLM.#linpred
                      linpred!      0 bytes  GLM.#linpred!
                            lm      0 bytes  GLM.#lm
                      logistic      0 bytes  StatsFuns.#logistic
                         logit      0 bytes  StatsFuns.#logit
                 loglikelihood      0 bytes  StatsBase.#loglikelihood
                model_response      0 bytes  StatsBase.#model_response
                         mueta      0 bytes  GLM.#mueta
                       mustart      0 bytes  GLM.#mustart
                          nobs      0 bytes  StatsBase.#nobs
                  nulldeviance      0 bytes  StatsBase.#nulldeviance
             nullloglikelihood      0 bytes  StatsBase.#nullloglikelihood
                       predict      0 bytes  StatsBase.#predict
                            r2      0 bytes  StatsBase.#r2
                     residuals      0 bytes  StatsBase.#residuals
                            r²      0 bytes  StatsBase.#r2
                 setcontrasts!      0 bytes  StatsModels.#setcontrasts!
                        stderr      0 bytes  StatsBase.#stderr
                      updateμ!      0 bytes  GLM.#updateμ!
                          vcov      0 bytes  StatsBase.#vcov
                       wrkresp      0 bytes  GLM.#wrkresp
```

nalimilan:
Maybe `@reexport` is confused by the fact that `ModelMatrix` and `ModelFrame` are both types and functions (via constructors)? The `GLM.#ModelMatrix` indicates a function type (the type of a type is `DataType`). Probably worth filing an issue.

dmbates:
Thanks @nalimilan.  I just discovered that problem myself.  For the time being I will remove those constructors.  I think it is another case of my getting trying to get too cute with the code.

dmbates:
Travis failure was transient.

nalimilan:
Do you want to tag a release?

dmbates:
I'll do so now.

dmbates:
I found that I needed to make one more change.  I am waiting on Travis.

PR #219, opened 2018-03-27
Fix test fail by stderr not defined

yukota:
Compat and StatsBase are both having stderr.
So, qualified with StatsBase.


nalimilan:
Hmm, thanks, but it looks like there's a deeper problem here. We should find a way to avoid this naming conflict on Julia 0.7.

nalimilan:
See https://github.com/JuliaLang/julia/pull/26636.

PR #222, opened 2018-04-25
use macro to construct formula in test for StatsModels v0.2.4

kleinschmidt:
Fix deprecated 2-argument `Formula` constructor in StatsModels.jl v0.2.4


nalimilan:
Failures are due to the `stderr` issue. Can you confirm tests pass locally even with StatsModels 0.2.3?

kleinschmidt:
Tests fail locally for the same reason on v0.2.3.

PR #227, opened 2018-05-03
Change StatsBase.stderr to StatsBase.stderror and bump StatsBase requirement

andreasnoack:
Tighten type for offset vectors in tests to avoid Missing types

Upper bound StatsModels to 0.2.4


BenjaminBorn:
Would it be possible to tag a new release? It's a bit unfortunate that the documentation already lists `stderror` but that doesn't work in the latest release (0.10.1).

andreasnoack:
It will probably happen soon. Indeed, it is unfortunate that the documentation is not in sync with the code. Ideally, we should doc documenter running since it can handle this situation.

PR #231, opened 2018-05-08
Enable Documenter hosting

andreasnoack:


andreasnoack:
This PR doesn't change anything but enabling the Travis build of the documentation so I'll merge to see if things work correctly. If it does, I'll open a separate PR with some adjustments to the Readme

nalimilan:
AFAIK mkdocs.yml isn't used when `format = :html`. CategoricalArrays doesn't have it for example.

andreasnoack:
I saw your previous comment about that after merging this so I've removed it again.

PR #232, opened 2018-05-09
Move documentation from Readme to docs/index.md

andreasnoack:
Fix doc tests

Use `lm` instead of `glm` for least squares example


docs/src/index.md
```diff
@@ -1,12 +1,377 @@
...
+ 0.288675
+
+julia> predict(ols)
+3-element Array{Float64,1}:
+ 1.83333
+ 4.33333
+ 6.83333
+
+```
+<!-- Andreas Noack: As of 9 May 2018 this example doesn't work so I've temporarily commented it out
```
ararslan: Nice, I always forget that markdown supports HTML-style comments.

src/glmtools.jl
```diff
@@ -17,7 +17,7 @@ abstract type Link01 <: Link end
     CauchitLink
 
 A [`Link01`](@ref) corresponding to the standard Cauchy distribution,
-[`Distributions.Cauchy`](@ref).
+[`Distributions.Cauchy`](https://juliastats.github.io/Distributions.jl/latest/univariate.html#Distributions.Cauchy).
```
ararslan: Should we link to stable instead of latest?

docs/src/index.md
kleinschmidt: the original and still best

kleinschmidt:
While we're updating the docs, I think it would be good to point out that there are two interfaces: the "low-level" (`X` matrix and `y` vector) and "high level" (`formula` and `DataFrame`), and that the high-level is provided by statsmodels

andreasnoack:
I think it is a good idea to add that to the documentation but I think it should go in a separate PR. This one just moves things from Readme to the docs and fixes the examples.

src/glmtools.jl
ararslan: I do not. That's odd...

andreasnoack:
I hadn't noticed that the docstests failed because `RDatasets` was missing since it wasn't missing locally. I've added some to install it during doctests if needed.

PR #234, opened 2018-05-10
Don't export as much

andreasnoack:
Currently, we export quite a few functions and types. I think most of them aren't relevant to most users so I'm suggesting that they are no longer exported. In this version, I haven't unexported `canonicallink`, `devresid`, and `linpred` but I think we could consider unexporting them as well. Thoughts?


ararslan:
I think the set of stuff you've chosen to unexport makes sense. I think we should leave `devresid`, since it's useful for model diagnostics. I don't really have an opinion about `linpred`. I don't think it really hurts to have `canonicallink` exported, but I don't care too much about this one as well.

nalimilan:
Merge?

Issue #209, opened 2018-01-02
Support for Negative Binomial Regression


DoktorMike:
It would be great if GLM also supported Negative Binomial Regression Models. This could be implemented in a number of ways. An over/under dispersed Poisson is one way. In R this corresponds to selecting the "quasipoisson" family in the call to glm.

Any plans to include it? 

Happy to elaborate if necessary.

nalimilan:
AFAIK quasi-poisson is completely different from negative binomial, even though these are often used in similar contexts. Implementation-wise, it seems that this package could support quasi-poisson, but fitting a negative binomial requires slightly different algorithms (in R, they are implemented in separate packages).

@dmbates must have thought about this already.

dmbates:
For the most part I prefer to stay away from quasi-Poisson and negative binomial regression models but that may be because I don't know enough about them or the general problem of over-dispersion.

Of the two approaches I prefer negative binomial but only because the quasi-Poisson approach of "let's model the probability distribution of the response as a function which isn't a probability distribution" offends my sensibilities.  (I'm a recovering mathematician.)

Someone could probably read section 7.4 of "Modern Applied Statistics with S, 4th ed." and figure out what should happen.  Apparently one sets a parameter, called "theta" in the 'negative.binomial' family in the MASS package for R, to a fixed value, fits a Poisson-like GLM conditional on that value, evaluates a likelihood then optimizes with respect to theta.  I can't remember offhand what 'theta' represents in terms of the probability of success and number of successes in a Bernoulli process.  Perhaps @bbolker can help.  He knows more about this than I do.

bbolker:
I agree with Doug in preferring explicit statistical models to quasi-likelihood, but I don't consider quasi-likelihood evil.

(1) "quasi-Poisson" doesn't really involve any different computation from the regular Poisson: the procedure is

- fit the Poisson model
- estimate the overdispersion, typically by dividing the deviance by the residual degrees of freedom (there's a bunch in McCullagh and Nelder about various ways to estimate overdispersion, e.g. putting residual deviance vs. Pearson chi-square statistics vs ... in the numerator)
- scale the estimated variance-covariance matrix of the parameters by the overdispersion
- compute standard errors, p-values, etc. as usual

[...]

PR #238, opened 2018-06-07
Negative Binomial regression support

hung-q-ngo:
There were a couple of requests for negative binomial regression support ( [issue 209](https://github.com/JuliaStats/GLM.jl/issues/209) for instance). From comments by @dmbates and @bbolker , and from reading section 7.4 of "_Modern Applied Statistics with S, 4th ed._", I understand that training the model involves fixing "theta", use GLM machinery to estimate "beta", then use that to estimate "theta" and so forth.

This PR allows `GLM.jl` to support calls of the type
```
nbmodel = glm(@formula(Days ~ Eth+Sex+Age+Lrn), quine, NegativeBinomial(2.0), LogLink())
```
in exactly the same way we can write that in `R`, without using the `glm.nb` package:
```
model = glm(Days ~ Eth + Sex + Age + Lrn, family = negative.binomial(2), data=quine)
```
In the above calls, the parameter `theta` is fixed/given. To also compute an MLE estimate for theta, we can call

```
model = negbin(@formula(Days ~ Eth+Sex+Age+Lrn), quine, LogLink())           
```

Comments/feedback are greatly appreciated.

Fixes https://github.com/JuliaStats/GLM.jl/issues/209.


hung-q-ngo:
From the log, it looks like `using RDatasets` is the culprit (not found on the current path). Is it not kosher to use `RDatasets` in the tests?

ararslan:
It should be okay, you just need to add RDatasets to test/REQUIRE.

hung-q-ngo:
FYI: I'm now also done with estimating `theta` using MLE, exactly in the same way described in the book ""Modern Applied Statistics with S, 4th ed.", and as implemented in `glm.nb` from `R`.

I updated the PR's description to this effect.

src/glmfit.jl
```diff
@@ -118,6 +119,20 @@ function updateμ!(r::GlmResp{V,D,L}) where {V<:FPVector,D<:Union{Bernoulli,Bino
...
+function updateμ!(r::GlmResp{V,D,L}) where {V<:FPVector,D<:NegativeBinomial,L<:NegativeBinomialLink}
+    y, η, μ, wrkres, wrkwt, dres = r.y, r.eta, r.mu, r.wrkresid, r.wrkwt, r.devresid
+
+    @inbounds for i in eachindex(y, η, μ, wrkres, wrkwt, dres)
+        θ = r.d.r # the shape parameter of the negative binomial distribution
+        μi, dμdη, μomμ = inverselink(L(θ), η[i])
+        μ[i] = μi
+        yi = y[i]
+        wrkres[i] = (yi - μi) / dμdη
+        wrkwt[i] = cancancel(r) ? dμdη : abs2(dμdη) / μomμ
```
dmbates: I think you can skip the `cancel(r)` here because the method specializes both `D` and `L`

src/negbinfit.jl
```diff
@@ -0,0 +1,81 @@
+function mleForθ(y::AbstractVector, μ::AbstractVector, wts::AbstractVector;
+                 maxIter = 30, convTol = 1.e-6)
+    function firstDeriv(θ::Real)
+        tmp = (y+θ)./(μ+θ) + log.(μ+θ) - 1 - log.(θ) - digamma.(θ+y) + digamma(θ)
```
nalimilan: It would be nice to avoid allocating intermediary copies here. You could turn this into an anonymous function to pass to `sum`, or use a generator. Same below and in the main function.

src/negbinfit.jl
```diff
@@ -0,0 +1,81 @@
...
+    end
+    function secondDeriv(θ::Real)
+        tmp = -(y+θ)./(μ+θ).^2 + 2 ./ (μ+θ) - 1/θ - trigamma.(θ+y) + trigamma(θ)
+        unitWeights ? sum(tmp) : sum(wts .* tmp)
+    end
+
+    unitWeights = length(wts) == 0
+    n = unitWeights ? length(y) : sum(wts)
+    θ = n / (unitWeights ? sum((y ./ μ - 1).^2) : sum(wts .* (y ./ μ - 1).^2))
+    δ, converged = 1, false
```
nalimilan: For type stability, better define `δ = one(θ)` or something like that.

src/negbinfit.jl
```diff
@@ -0,0 +1,81 @@
...
+    converged || throw(ConvergenceException(maxIter))
+    θ
+end
+
+function negbin(F, D, args...; kwargs...)
+    # if initialθ was specified, then use it to train the initial model, remove the argument from fitargs;
+    # otherwise, the initial model is trained with Poisson regression
+    new_kwargs = Dict(kwargs)
+    initialθset = haskey(new_kwargs, :initialθ)
+    maxIter = haskey(new_kwargs, :maxIter) ? new_kwargs[:maxIter] : 30
```
nalimilan: Why not use the standard keyword argument syntax? It doesn't prevent you from passing these arguments (manually) do `glm`.

src/negbinfit.jl
```diff
@@ -0,0 +1,81 @@
...
+    y = regModel.model.rr.y
+    wts = regModel.model.rr.wts
+    lw, ly = length(wts), length(y)
+    if (lw != ly && lw != 0)
+        throw(ArgumentError("length of wts must be either $ly or 0 but was $lw"))
+    end
+
+    θ = mleForθ(y, μ, wts)
+    d = sqrt(2 * max(1, deviance(regModel))); 
+    δ = 1
```
nalimilan: Same remark for type stability.

src/negbinfit.jl
```diff
@@ -0,0 +1,81 @@
...
+    μ = regModel.model.rr.mu
+    y = regModel.model.rr.y
+    wts = regModel.model.rr.wts
+    lw, ly = length(wts), length(y)
+    if (lw != ly && lw != 0)
+        throw(ArgumentError("length of wts must be either $ly or 0 but was $lw"))
+    end
+
+    θ = mleForθ(y, μ, wts)
+    d = sqrt(2 * max(1, deviance(regModel)));
```
nalimilan: No need for a semicolon.

src/negbinfit.jl
```diff
@@ -0,0 +1,81 @@
...
+    initialθ > 0 || throw(ArgumentError("initialθ must be positive"))
+
+    regModel = initialθset ? glm(F, D, NegativeBinomial(initialθ), args...; new_kwargs...) :
+                             glm(F, D, Poisson(), args...; new_kwargs...)
+
+    μ = regModel.model.rr.mu
+    y = regModel.model.rr.y
+    wts = regModel.model.rr.wts
+    lw, ly = length(wts), length(y)
+    if (lw != ly && lw != 0)
```
nalimilan: Can you remove unneeded parentheses with `if` statements here and elsewhere? Just a matter of style.

src/negbinfit.jl
```diff
@@ -0,0 +1,81 @@
...
+    # if initialθ was specified, then use it to train the initial model, remove the argument from fitargs;
+    # otherwise, the initial model is trained with Poisson regression
+    new_kwargs = Dict(kwargs)
+    initialθset = haskey(new_kwargs, :initialθ)
+    maxIter = haskey(new_kwargs, :maxIter) ? new_kwargs[:maxIter] : 30
+    convTol = haskey(new_kwargs, :convTol) ? new_kwargs[:convTol] : 1.e-6
+    verbose = haskey(new_kwargs, :verbose) ? new_kwargs[:verbose] : false
+
+    initialθ = 1.0
+    if initialθset
```
nalimilan: You can achieve this more simply by listing it as a keyword argument and not passing it to `glm`.

src/negbinfit.jl
```diff
@@ -0,0 +1,81 @@
...
+    wts = regModel.model.rr.wts
+    lw, ly = length(wts), length(y)
+    if (lw != ly && lw != 0)
+        throw(ArgumentError("length of wts must be either $ly or 0 but was $lw"))
+    end
+
+    θ = mleForθ(y, μ, wts)
+    d = sqrt(2 * max(1, deviance(regModel))); 
+    δ = 1
+    lm = loglikelihood(regModel)
```
nalimilan: `lm` generally stands for "linear model", why not `ll` instead?

test/runtests.jl
```diff
@@ -65,6 +66,7 @@ dobson = DataFrame(Counts = [18.,17,15,20,10,20,25,13,12],
 
 @testset "Poisson GLM" begin
     gm1 = fit(GeneralizedLinearModel, @formula(Counts ~ 1 + Outcome + Treatment),
+
```
nalimilan: Unintended change? Same above.

src/glmtools.jl
```diff
@@ -66,6 +66,16 @@ The canonical [`Link`](@ref) for [`Distributions.Poisson`](https://juliastats.gi
 """
 mutable struct LogLink <: Link end
 
+"""
+    NegativeBinomialLink
+
+The canonical [`Link`](@ref) for [`Distributions.NegativeBinomial`](@ref) distribution, defined as `η = log(μ/(μ+θ))`.
+The shape parameter θ has to be fixed for the distribution to belong to the exponential family
```
nalimilan: Missing ending dot.

src/glmtools.jl
```diff
@@ -391,3 +420,12 @@ loglik_obs(::Gamma, y, μ, wt, ϕ) = wt*logpdf(Gamma(inv(ϕ), μ*ϕ), y)
 loglik_obs(::InverseGaussian, y, μ, wt, ϕ) = wt*logpdf(InverseGaussian(μ, inv(ϕ)), y)
 loglik_obs(::Normal, y, μ, wt, ϕ) = wt*logpdf(Normal(μ, sqrt(ϕ)), y)
 loglik_obs(::Poisson, y, μ, wt, ϕ) = wt*logpdf(Poisson(μ), y)
+# in principle, μ = θp/(1-p) ==> p = μ/(μ+θ), and thus we should have been able
+# to do following
+# loglik_obs(d::NegativeBinomial, y, μ, wt, ϕ) = wt*logpdf(NegativeBinomial(d.r, μ/(μ+d.r)), y)
+# however, that does not give the correct answer compared to that of R's `glm.nb`
```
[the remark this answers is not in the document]
hung-q-ngo: @nalimilan : I appreciate the encouragement.

W.r.t. the above `loglik_obs` issue, if we use the straightforward 
```
loglik_obs(d::NegativeBinomial, y, μ, wt, ϕ) = wt*logpdf(NegativeBinomial(d.r, μ/(μ+d.r)), y)
```
then the `loglikelihood` numbers are not correct anymore. In `runtest.jl`, the `gm18` model would return
```
julia> loglikelihood(gm18)
-5208.360845881062
```
instead of the expected ` -553.259` (the same LL returned by `glm.nb` from `MASS`). I haven't been able to figure out why.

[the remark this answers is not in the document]
hung-q-ngo: The only other `R` package I found is `vglm`. Its output is similar to that of `glm.nb`:

```
> nb = glm.nb(Days ~ Eth + Sex + Age + Lrn, data=quine)                          
> nb2 = vglm(Days ~ Eth + Sex + Age + Lrn, family=negbinomial(), data=quine)                          
> logLik(nb)
'log Lik.' -546.5755 (df=8)
> logLik(nb2)
[1] -546.5755
```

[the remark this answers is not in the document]
bbolker: If you want more R-implementation comparisons, there are versions in (at least) the  `glmmTMB` and `mgcv` and  `bbmle` packages ...

src/negbinfit.jl
```diff
@@ -26,55 +26,43 @@ function mleForθ(y::AbstractVector, μ::AbstractVector, wts::AbstractVector;
...
+function negbin(F, D, args...; initialθ::Real=Inf, maxIter::Integer=30, convTol::Real=1.e-6, verbose::Bool=false, kwargs...)
     maxIter >= 1 || throw(ArgumentError("maxIter must be positive"))
     convTol > 0  || throw(ArgumentError("convTol must be positive"))
     initialθ > 0 || throw(ArgumentError("initialθ must be positive"))
 
-    regModel = initialθset ? glm(F, D, NegativeBinomial(initialθ), args...; new_kwargs...) :
-                             glm(F, D, Poisson(), args...; new_kwargs...)
+    # fit a Poisson regression model if the user does not specify an initial θ
+    regModel = isinf(initialθ) ? glm(F, D, Poisson(), args...; maxIter=maxIter, convTol=convTol, verbose=verbose, kwargs...) :
+                                 glm(F, D, NegativeBinomial(initialθ), args...; maxIter=maxIter, convTol=convTol, verbose=verbose, kwargs...)
```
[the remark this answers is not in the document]
hung-q-ngo: certainly. I come from the background of 80 chars limit, but I was not able to find an official width limit from the Julia style guide. 92 is reasonable.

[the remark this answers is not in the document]
ararslan: The style guide in the Julia manual tends to be pretty hands-off when it comes to aesthetics like indentation and line lengths. In the JuliaStats packages, we typically use the style laid out in Julia's own [CONTRIBUTING.md](https://github.com/JuliaLang/julia/blob/master/CONTRIBUTING.md).

src/glmtools.jl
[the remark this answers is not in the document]
hung-q-ngo: Many thanks @bbolker . will try them out

src/negbinfit.jl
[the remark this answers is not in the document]
hung-q-ngo: got it. Now I know where 92 comes from.

src/glmtools.jl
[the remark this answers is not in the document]
hung-q-ngo: Report from the `glmmTMP` package: logLik agrees with `glm.nb` and `vglm`
```
> nb_tmb = glmmTMB(Days ~ Eth + Sex + Age + Lrn, family=nbinom2(), data=quine)                          
> summary(nb_tmb)
 Family: nbinom2  ( log )
Formula:          Days ~ Eth + Sex + Age + Lrn
Data: quine

     AIC      BIC   logLik deviance df.resid 
  1109.2   1133.0   -546.6   1093.2      138 


Overdispersion parameter for nbinom2 family (): 1.27 

Conditional model:
            Estimate Std. Error z value Pr(>|z|)    
(Intercept)  2.89459    0.22793  12.700  < 2e-16 ***
EthN        -0.56938    0.15761  -3.613 0.000303 ***
SexM         0.08231    0.16468   0.500 0.617211    
AgeF1       -0.44843    0.23760  -1.887 0.059119 .  
AgeF2        0.08808    0.24155   0.365 0.715370    
AgeF3        0.35691    0.24662   1.447 0.147840    
LrnSL        0.29211    0.18294   1.597 0.110314    
---
```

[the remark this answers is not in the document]
nalimilan: Mmm... It's interesting that R packages don't use the built-in `dnbinom` function. Could it be due to using different parameterizations? To be clear, which [formulation](https://en.wikipedia.org/wiki/Negative_binomial_distribution#Alternative_formulations) do you use in the hand-written formula? Looking at glmmTB, it uses `dbinom_robust` from the [TMB](https://github.com/kaskr/adcomp) package (but you can't read the code yourself since it's GPL2 and GLM.jl is MIT), which is formulated in terms of `mu`. OTC, the parameterization used by Distributions.jl comes from R's `dbinom`, which is formulated in terms of `prob` by default. There's a `dnbinom_mu` variant formulated in terms of `mu` which can be used via an argument in R (see [details in the docs](https://stat.ethz.ch/R-manual/R-devel/library/stats/html/NegBinomial.html)). It can be access from Julia via something like (arguments are `x`, `size`, `mu` and `give_log`, and you'll need to adapt the path to the file):
```
ccall((:dnbinom, ".julia/Rmath/deps/usr/lib/libRmath-julia.so"), Float64, (Float64, Float64, Float64, Int), y, length(y), μ, true)
```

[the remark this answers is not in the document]
hung-q-ngo: The hand-written formula uses the formulation in Section 7.4 of [Modern Applied Statistics with S](http://www.bagualu.net/wordpress/wp-content/uploads/2015/10/Modern_Applied_Statistics_With_S.pdf). I'm not sure if the pdf is legal, just grabbed it off of a Google search.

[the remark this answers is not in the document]
nalimilan: OK. It turns out that at the end of [this section](https://en.wikipedia.org/wiki/Negative_binomial_distribution#Alternative_formulations) of the Wikipedia article, there's an explanation of the relationship between this formula and the standard expression for the negative binomial PDF (but with `m = μ`, `k = y` and `r = θ`). So indeed `p = μ/(μ+θ)` sounds correct.

I've been able to replicate the values of your `loglik_obs` definition in both R and Julia:
```R
> dnbinom(300, mu=100, size=1, log=T)
[1] -7.60022

> dnbinom(300, prob=1/(100+1), size=1, log=T)
[1] -7.60022
```

```julia
julia> function loglik_obs(d::NegativeBinomial, y, μ, wt, ϕ)
           θ = d.r
           wt * (lgamma(θ + y) - lgamma(θ) - lgamma(y + 1) + θ * log(θ) + y * log(μ + (y == 0)) - (θ + y) * log(θ + μ))
       end
loglik_obs (generic function with 2 methods)

julia> loglik_obs(NegativeBinomial(1, 1/(100+1)), 300, 100, 1, 0)
-7.600219772791661

julia> logpdf(NegativeBinomial(1, 1/(100+1)), 300)
-7.600219772791684
```

I guess this should allow replacing the custom method with the standard one?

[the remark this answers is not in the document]
hung-q-ngo: Oh, I know why; it was a stupid mistake of mine. The reparameterization should have been p = θ/(μ+θ), *not* p = μ/(μ+θ). Now we can use `logpdf` and everything looks fine. Apologies for wasting your time on this; thanks much for digging into the details.

src/glmtools.jl
```diff
@@ -391,3 +417,9 @@ loglik_obs(::Gamma, y, μ, wt, ϕ) = wt*logpdf(Gamma(inv(ϕ), μ*ϕ), y)
 loglik_obs(::InverseGaussian, y, μ, wt, ϕ) = wt*logpdf(InverseGaussian(μ, inv(ϕ)), y)
 loglik_obs(::Normal, y, μ, wt, ϕ) = wt*logpdf(Normal(μ, sqrt(ϕ)), y)
 loglik_obs(::Poisson, y, μ, wt, ϕ) = wt*logpdf(Poisson(μ), y)
+# We use the following parameterization for the Negative Binomial distribution:
+#    (Γ(θ+y) / (Γ(θ) * y!)) * μ^y * θ^θ / (μ+θ)^{θ+y}
+# The parameterization of NegativeBinomial(r=θ, p) in Distribution.jl is
```
nalimilan: "Distributions.jl"

Something still escapes me, since looking at the Wikipedia page it seems to me that `r/(m+r)` should be equal to `1-p`, but I trust the fact that it works more than my limited understanding of these parameterization subtleties.

src/glmtools.jl
```diff
@@ -352,6 +374,10 @@ function devresid(::Binomial, y, μ)
 end
 devresid(::Gamma, y, μ) = -2 * (log(y / μ) - (y - μ) / μ)
 devresid(::InverseGaussian, y, μ) = abs2(y - μ) / (y * abs2(μ))
+function devresid(d::NegativeBinomial, y, μ)
+    θ = d.r
+    return μ == 0.0 ? NaN : 2 * (xlogy(y, y / μ) + xlogy(y + θ, (μ + θ)/(y + θ)))
```
nalimilan: Should use `oftype`. I guess the most reliable solution is to compute the second operand anyway and store it in a variable, and use `oftype(v, NaN)`.

BTW, `μ == 0` is enough.

src/negbinfit.jl
```diff
@@ -0,0 +1,85 @@
...
+        if abs(ll0 - ll)/d + abs(δ) <= convTol
+            converged = true
+            break
+        end
+        verbose && println("[ Alternating iteration ", i, ", θ = ", θ, "]")
+        regmodel = glm(F, D, NegativeBinomial(θ), args...;
+                       maxIter=maxIter, convTol=convTol, verbose=verbose, kwargs...)
+        μ = regmodel.model.rr.mu
+        prevθ = θ
+        θ = mle_for_θ(y, μ, wts; maxIter = maxIter, convTol = convTol)
```
nalimilan: Check that no spaces are put around `=` for arguments here and elsewhere, for consistency.

nalimilan:
Thanks. I've already approved, so I'll leave somebody else merge this.

nalimilan:
Do not hesitate if you want to make other PRs @hung-q-ngo! ;-)

hung-q-ngo:
it's getting addictive, thanks @nalimilan & everyone for the warm welcome

Nosferican:
When running the tests for the negative binomials, it seems like the intercept differs from the hardcoded values.

hung-q-ngo:
@Nosferican : can you be more specific? which test and which intercept? Were you using `RDataSets` or the separate `csv` file?

Nosferican:
Separate text file. It was also documented [here](https://github.com/JuliaStats/GLM.jl/pull/242#issue-207436741). You can see which tests [here](https://github.com/Nosferican/GLM.jl/commit/4860b80adf2d14c0b48ce19638ceeadff2d65d76) and just look for `@test_broken`. All parameter estimates except the intercept match.

hung-q-ngo:
Ah, did you read the comment thread there? I was under the assumption that we agreed it was due to the different category orders when importing from RDatasets vs from csv. And the resolution was to wait for [this pull](https://github.com/johnmyleswhite/RDatasets.jl/pull/54) to merge and tests should pass as expected.

Nosferican:
I didn't I was just bumped when I found the PR after I had done updating the code independently. Hehe. That should be fixed by specifying the `contrasts` in the StatsModels such that it is consistent.

hung-q-ngo:
hehe. sorry about that :-) . Do you mind doing the `contrasts` fix? I don't know how to do it. I can find out, but it would take me a while.

Nosferican:
I will update it to in my script and people can take a look at it. No worries.

hung-q-ngo:
I also owe you a careful read on the multinomial case; just got *really* busy at work lately.

Nosferican:
I just realized that GLM still doesn't allow to pass contrast specifications. That should probably be a different PR.

kleinschmidt:
Applying contrasts really only makes sense when the input data is non-numeric (or rather is categorical), in which case the `DataFrameRegressionModel` wrappers and the `fit(::GeneralizedLinearModel, ::Formula, ::DataFrame)` methods should be used...

PR #242, opened 2018-08-09
Update for 0.7/1.0

jacobadenbaum:
I hope this isn't duplicate work of what someone else has locally, but since GLM isn't yet working on 0.7 I thought I'd take a stab at cleaning up the deprecated syntax that hadn't been sorted through by @dmbates yet, etc...  There's still an unresolved issue with the Negative Binomial test cases, so this isn't ready to be merged just yet.  

# Syntax Upgrades
I removed the RDatasets testing dependency, since it was throwing an error, and I replaced it with a CSV copy of the relevant test dataset.  The other updates to 0.7 are fairly straightforward upgrades of syntax, and I have all of the tests passing except for a few that are related to Negative Binomial link functions.  

# Negative Binomial Link Tests
Fixing this is a bit beyond what I'm sure of how to do, since I think it might be a numerical instability issue in the estimation.  I've set the tests to @test_broken so they're easy to spot in the code/diff.  @hung-q-ngo or someone else who understands the internals a little better, would you mind taking a look at what's going wrong?  

From my first pass at this, it looks to me like the regression problem that's being posed may not be identified.  It's passing all the tests for the log-likelihood, aic, other fit statistics, and most of the coefficients other than the intercept seem to be correct up to their sign.  For instance, in the first regression, we have

```julia
julia> gm18 = fit(GeneralizedLinearModel, @formula(Days ~ Eth+Sex+Age+Lrn), quine, NegativeBinomial(2.0), LogLink())
StatsModels.DataFrameRegressionModel{GeneralizedLinearModel{GlmResp{Array{Float64,1},NegativeBinomial{Float64},LogLink},DensePredChol{Float64,Cholesky{Float64,Array{Float64,2}}}},Array{Float64,2}}

Formula: Days ~ 1 + Eth + Sex + Age + Lrn

Coefficients:
               Estimate Std.Error   z value Pr(>|z|)
(Intercept)     3.27029  0.244688   13.3652   <1e-40
Eth: N        -0.567515  0.152449  -3.72265   0.0002
Sex: F       -0.0870771  0.159025 -0.547568   0.5840
Age: F1       -0.445076  0.239087  -1.86157   0.0627
Age: F2       0.0927999  0.234502  0.395731   0.6923
Age: F3        0.359485  0.246586   1.45785   0.1449
Lrn: AL       -0.296768  0.185934  -1.59609   0.1105


julia> @test !GLM.cancancel(gm18.model.rr)
Test Passed

julia> test_show(gm18)

julia> @test dof(gm18) == 8
Test Passed

julia> @test isapprox(deviance(gm18), 239.11105911824325, rtol = 1e-7)
Test Passed

julia> @test isapprox(loglikelihood(gm18), -553.2596040803376, rtol = 1e-7)
Test Passed

julia> @test isapprox(aic(gm18), 1122.5192081606751)
Test Passed

julia> @test isapprox(aicc(gm18), 1123.570303051186)
Test Passed

julia> @test isapprox(bic(gm18), 1146.3880611343418)
Test Passed

# Here's the problem
julia> hcat(coef(gm18), true_coef)
7×2 Array{Float64,2}:
  3.27029     2.88645
 -0.567515   -0.567515
 -0.0870771   0.0870771
 -0.445076   -0.445076
  0.0927999   0.0927999
  0.359485    0.359485
 -0.296768    0.296768
```

Looks to me like this might just be a bad test dataset/regression for this problem?  Either that, or there's something else wrong that I'm just not seeing.


andreasnoack:
I'll take a closer look tomorrow but I'm pretty sure we don't want to remove the test dependency on RDatasets here. We just want to fix the dependency. It will probably happen fairly soon.

jacobadenbaum:
I don't care very much one way or the other about whether to remove the test dependency (although I'd like to see GLM upgraded sooner rather than later since it's dependency for one of my packages and causing some of my 0.7 tests to fail).  But especially since all of the other test datasets are also stored as CSVs and this was the only one that got pulled in dynamically, it's not obvious why you would want it there.

dmbates:
Thank you for taking this on, @jacobadenbaum 

I'm good at starting projects but not so good at completing them. :smile:

jacobadenbaum:
Awesome.  Lemme know when it's done and I'll update the PR to put the dependency back in if that's what you guys want.  I'm still not sure what to do about the Negative Binomial tests failing on 0.7/1.0 though, since really nothing changed that I can tell that should affect them...

hung-q-ngo:
@jacobadenbaum : my best guess is that the data frame imported via RDatasets and via the csv file are somehow sorted slightly differently. With any given categorical variable, say `age = {F1, F2, F3}`, the model has coefficients for all but one of the categories. So if they were sorted differently, the set of coefficients are also slightly different. I'm all tied up at work; will try to verify my hunch as soon as I can.

jacobadenbaum:
Hmm.  The dataset didn't change (I just literally just pulled it from RDatasets on 0.6 and wrote it to a CSV file, plus I just double checked).  Is it possible that something changed in one of the dependencies to the way that categoricals are coded?

dmbates:
Are there categorical covariates in the model?  If you take a categorical variable, write it to a .csv file and read it back in, you can end up with a different order of the levels in the `CategoricalVector`, which will cause different contrasts to be used in the model.

jacobadenbaum:
Ahh yes.  Then that would be it.  I guess it is my fault after all. 😄 Once this [pull](https://github.com/johnmyleswhite/RDatasets.jl/pull/54) to RDatasets is merged, it should work then with the RDatasets dependency.

Nosferican:
I am a little late to the party since I didn't see the PR last week or an issue to track it. I went and did it today [here](https://github.com/Nosferican/GLM.jl/commit/4860b80adf2d14c0b48ce19638ceeadff2d65d76). So feel free to take whatever y'all might need/want or check for accuracy.

jacobadenbaum:
@andreasnoack do you know what's going on with the RDatasets update?  Do you still think it's worth waiting for this PR, or should I set those tests back to @test_broken pending the RDatasets update since we know what the problem is.  I could try to fix the negative binomial tests by manually setting the contrasts...

I'd like to get a working GLM version on 0.7/1.0 if possible

dmbates:
I think it has reached the point where it is not worth waiting on an RDatasets update.  I would recommend setting the tests back to `@test_broken` just so we can get a version working in 0.7/1.0

andreasnoack:
I think `@test_broken` is fine here. Hopefully, Julia 1.0.1 will be released fairly soon. The thing I'd like to avoid is that we check new datasets into the repo.

dmbates:
@jacobadenbaum Could you make the change on the quine example so that I can merge the pull request?

REQUIRE
```diff
@@ -1,8 +1,8 @@
-julia 0.6
-Compat 0.33.0
-Distributions 0.10.0
+julia 0.7-beta2
```
Nosferican: This should be `julia 0.7 2` for it to be parsed as `[0.7, 1]`

REQUIRE
```diff
@@ -1,8 +1,8 @@
...
+Compat 0.60.0
```
Nosferican: I don't think this is using `Compat` anymore.

src/GLM.jl
```diff
@@ -1,22 +1,22 @@
 __precompile__()
```
Nosferican: This should be dropped.

Nosferican:
This works just fine in Julia 1.0 (with RDataset v0.5.0)
```
using RDatasets
quine = dataset("MASS", "quine")
```
Can someone get `attobot` permissions for `RDatasets` or make a PR to METADATA?

.travis.yml
```diff
@@ -1,25 +1,23 @@
+## Documentation: http://docs.travis-ci.com/user/languages/julia/
 language: julia
-
 os:
   - linux
   - osx
-
 julia:
-  - 0.6
+  - 0.7
```
andreasnoack: Please add 1.0 as well

test/runtests.jl
```diff
@@ -331,7 +340,8 @@ end
 end
 
 # "quine" dataset discussed in Section 7.4 of "Modern Applied Statistics with S"
-quine = dataset("MASS", "quine")
+# quine = dataset("MASS", "quine")
```
andreasnoack: This should work once https://github.com/JuliaLang/METADATA.jl/pull/17366 has been merged and that will most likely happen later today so please uncomment this and delete the csv.

.travis.yml
jacobadenbaum: I'll change the allow failures to nightly as well?

REQUIRE
jacobadenbaum: Okay.  That came from one of @dmbates's commits, so I wasn't going to get rid of it.  But I think you're right.

test/runtests.jl
jacobadenbaum: Okay.  I have a commit with the updates locally.  I'll push once the metadata pr is merged so it'll retrigger the tests.

Nosferican:
Few details that could be added to the PR if you have time.
- `src/glmtools.jl` 287-291 in the `jldoctest` the examples are missing a `Ref(::Distribution)`
- `src/glmfit.jl` 286-287 should use `Nothing` instead of `Void`
Could you check two issues I saw from my branch, but ain't sure
- `src/lm.jl` line 30:
```
LmResp(y::V) where {V<:FPVector} = LmResp{V}(fill!(similar(y), 0), similar(y, 0), similar(y, 0), y)
```
I think it should be moved from constructor to `convert`
```
convert(::Type{LmResp{V}}, y::V) where {V<:FPVector} =
    LmResp{V}(fill!(similar(y), zero(eltype(y))), similar(y, 0), similar(y, 0), y)
```
and
- `src/linpred.jl` 161
```
cf .= mul!(p.scratchm2, adjoint(LinearAlgebra.lmul!(p.scratchm1, Diagonal(wt), p.X)), p.X)[piv, piv]
```
when I worked on it I had it as `rmul!` (this part doesn't seem to be tested).
```
cf .= mul!(p.scratchm2, adjoint(rmul!(p.scratchm1, Diagonal(wt), p.X)), p.X)[piv, piv]
```
Maybe take a quick look at whether it should be `lmul!` or `rmul!`.

.travis.yml
Nosferican: Hopefully there will be a release before julia `v2`. I think we can save the planet a bit and skip nightly for now.

src/linpred.jl
```diff
@@ -112,59 +117,50 @@ end
...
     p
 end
 
 function delbeta!(p::DensePredChol{T,<:CholeskyPivoted}, r::Vector{T}, wt::Vector{T}) where T<:BlasReal
     cf = cholfactors(p.chol)
     piv = p.chol.piv
-    cf .= Ac_mul_B!(p.scratchm2, scale!(p.scratchm1, wt, p.X), p.X)[piv, piv]
-    cholfact!(Hermitian(cf, Symbol(p.chol.uplo)))
-    A_ldiv_B!(p.chol, At_mul_B!(p.delbeta, p.scratchm1, r))
+    cf .= mul!(p.scratchm2, adjoint(LinearAlgebra.lmul!(p.scratchm1, Diagonal(wt), p.X)), p.X)[piv, piv]
```
andreasnoack: `lmul!`->`mul!`

andreasnoack:
> I think it should be moved from constructor to `convert`

Why do think that is the case?

Regarding `l/rmul!` then see my comment https://github.com/JuliaStats/GLM.jl/pull/242/files#r213058572.

Nosferican:
I thought it was affected by some deprecation that indicated using `convert` rather than a constructor (`convert` not calling constructors by default anymore maybe?). It seems fine to have a constructor defined there so I leave whether to include the constructor / `convert` methods there. It was probably one of the structs that was calling `convert` implicitly that gave me the warning.

test/runtests.jl
andreasnoack: Okay. The new version of RDatasets has been merged so this should work now.

jacobadenbaum:
Okay.  Now that RDatasets is released, all the tests are passing.  I think this is ready, if no one else has other changes they want to make.

Nosferican:
Should probably get this tagged and released ASAP.
StatsModels API will probably change soon so better to get a release for MixedModels before it.

Leo-C:
I higlight also a [problem ](https://github.com/JuliaLang/julia/issues/28688) with `linreg()` and deprecation in Base module

andreasnoack:
@Leo-C `linreg` is unrelated to this package

PR #252, opened 2018-09-10
Fix ftest printing

nalimilan:
And uncomment tests (commented out by https://github.com/JuliaStats/GLM.jl/pull/192).

Also avoid using mod as an identifier since it can conflict with Base.mod


iwelch:
There is still the String error.  Can you confirm one complete example as working in Julia 1.0.0, please?

nalimilan:
The example modified in the PR works AFAICT. But you need git master for that.

iwelch:
oh, I see.  any idea when master will become the stable publicly released version?

nalimilan:
https://github.com/JuliaLang/METADATA.jl/pull/17940

PR #253, opened 2018-09-11
Fix predict with confidence intervals

nalimilan:
Use keyword arguments, which are passed through by the DataFrameRegressionModel method.

While we're at it, I wonder whether we should change the API a bit. `interval=:confint` and `interval=:predint` could be changed to `interval=:confidence` and `interval=:prediction`, which are the names used by R (since "int" is redundant with `interval`).

Cc: @mkborregaard

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -81,16 +81,17 @@ julia> predict(ols)
  4.33333
  6.83333
 
-```
-<!-- Andreas Noack: As of 9 May 2018 this example doesn't work so I've temporarily commented it out
 julia> newX = DataFrame(X=[2,3,4]);
 
-julia> predict(ols, newX, :confint)
+julia> predict(ols, newX, interval=:confint)
  3×3 Array{Float64,2}:
   4.33333  1.33845   7.32821
   6.83333  2.09801  11.5687
   9.33333  1.40962  17.257
-The columns of the matrix are prediction, 95% lower and upper confidence bounds -->
+```
+
+The columns of the matrix are prediction, 95% lower and upper confidence bounds
+.
 
 ### Probit Regression:
 ```jldoctest
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -83,7 +83,7 @@ julia> predict(ols)
 
 julia> newX = DataFrame(X=[2,3,4]);
 
-julia> predict(ols, newX, interval=:confint)
+julia> predict(ols, newX, interval=:confidence)
  3×3 Array{Float64,2}:
   4.33333  1.33845   7.32821
   6.83333  2.09801  11.5687
--- a/src/GLM.jl
+++ b/src/GLM.jl
@@ -99,5 +99,6 @@ module GLM
     include("glmfit.jl")
     include("ftest.jl")
     include("negbinfit.jl")
+    include("deprecated.jl")
 
 end # module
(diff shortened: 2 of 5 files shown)


mkborregaard:
Wonderful. I also agree wholeheartedly with the name change.

mkborregaard:
Does this mean this will work with DataFrames now? There was a PR on DataFrames I never got merged...

nalimilan:
> Does this mean this will work with DataFrames now? There was a PR on DataFrames I never got merged...

Do you have a link? It must have worked at some point, since it's in the examples, but I wonder since when it's been broken.

mkborregaard:
https://github.com/JuliaData/DataFrames.jl/pull/1160

I don't think it's ever worked from DataFrames...

mkborregaard:
I'll rebase and write the test if that's still doable, but I'm guessing the interface has changed a lot?

nalimilan:
Thanks. Actually I now realize I hadn't tested the example based on DataFrames correctly. The test in runtests.jl only covers matrices. So this PR shouldn't be merged without fixing `DataFrameRegressionModel` first.

The code now lives in StatsModels.jl. If we make interval-related arguments keyword arguments rather than positional, I think we can avoid defining a separate function as in https://github.com/JuliaData/DataFrames.jl/pull/1160. Instead, you can just check whether `yp` is a vector or a matrix, and slightly adapt the code to suit each possibility.

mkborregaard:
Here? https://github.com/JuliaStats/StatsModels.jl/blob/33f7b9cceac7e6624bf9e4802df3c4c8b2f5b83d/src/statsmodel.jl#L100

mkborregaard:
Or here in GLM?

nalimilan:
Yes, in StatsModels. GLM should be ready with this PR.

mkborregaard:
By the way @nalimilan I just rememembered that the way I implemented it like it is, and not with a keyword as in this PR, is because the keyword version breaks type stability of `predict`. With confidence intervals it returns a three-column `Matrix`, without it's a `Vector`. Making the interval an optional positional argument allows the compiler to figure out the return type from the method signature.

nalimilan:
AFAIK type-stability isn't a problem anymore with keyword arguments:
```julia
julia> f(; x=nothing) = x === nothing ? 1 : 1.0
f (generic function with 1 method)

julia> @code_warntype f()
Body::Int64
[...]

julia> @code_warntype f(x=nothing)
Body::Int64
[...]

julia> @code_warntype f(x=1)
Body::Float64
[...]
```

mkborregaard:
sweet :tada:

nalimilan:
I've pushed a commit to change the names. Is this good to go? IIRC a separate change is needed in StatsModels to support DataFrames, but that's already broken currently.

mkborregaard:
Yes, I'll have a look a StatsModels done time during the week

mkborregaard:
@nalimilan I can't see that the name has been changed? It still says `:confint` but I agree that `:confidence` makes more sense.

Here's the StatsModels PR:
https://github.com/JuliaStats/StatsModels.jl/pull/77

I have a question - we currently return a three-row Matrix as `[lower prediction upper]`. Would it make more sense / be more useful to return a Tuple, e.g. `(prediction, (lower, upper))` or maybe a NamedTuple? Allows things like `pred, int = predict(mymodel, newdata, interval = :confidence); plot(pred, ribbon = int)` etc.?

mkborregaard:
I've added prediction intervals in a PR against this branch

PR #264, opened 2018-11-06
Fixes bug in dof_residual with allowrankdeficient

Nosferican:
See #263 
It also fixes errors due to #201

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -178,10 +178,10 @@ anorexia = CSV.read(joinpath(glm_datadir, "anorexia.csv"))
     @test isapprox(aicc(gm6), 490.8823884513153)
     @test isapprox(bic(gm6), 501.35662813730465)
     @test isapprox(coef(gm6),
-        [45.674043486911685, -0.5655388496390964, 4.097065528072901, 8.660128180991693])
+        [49.7711090, -0.5655388, -4.0970655, 4.5630627])
     @test isapprox(GLM.dispersion(gm6.model, true), 48.6950385282296)
     @test isapprox(stderror(gm6),
-        [13.21670540145523, 0.16118236185182783, 1.893492606966926, 2.1931494116430517])
+        [13.3909581, 0.1611824, 1.8934926, 2.1333359])
 end
 
 @testset "Normal LogLink offset" begin
@@ -191,10 +191,10 @@ end
     test_show(gm7)
     @test isapprox(deviance(gm7), 3265.207242977156)
     @test isapprox(coef(gm7),
-        [3.9416285291318798, -0.9944526931311773, 0.050698258703983666, 0.1021922886616272])
+        [3.99232679, -0.99445269, -0.05069826, 0.05149403])
     @test isapprox(GLM.dispersion(gm7.model, true), 48.017753573192266)
     @test isapprox(stderror(gm7),
-        [0.1554026019351794, 0.0018862835443589627, 0.022584040191142126, 0.025187228659634627],
+        [0.157167944, 0.001886286, 0.022584069, 0.023882826],
         atol=1e-6)
 end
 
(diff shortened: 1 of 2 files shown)


andreasnoack:
Please use more informative issue titles.

tir. 6. nov. 2018 kl. 23.09 skrev Codecov <notifications@github.com>:

nalimilan:
Also there's no point in filing an issue at the same time as a PR to fix it, one is enough. (Even better if you don't file two other issues with empty descriptions.)

nalimilan:
In more practical terms, can you add a more detailed description/commit message so that we understand what the fix does?

Nosferican:
I filed the issues as trackers for myself since this PR only does one. It helps me organize my TODO, but also allows others to assign it to them if they so wish to do so. This PR is pretty straightforward. The general behavior is for `dof(obj) = nobs(obj) - length(coef(obj))`. When there are `allowrankdeficient`, the coefficient on the dropped terms are set to zero. The standard errors are computed such that the value for those are `NaN`. However, `dof` acts as if those were parameters for the model (which were not, since those were dropped due to multicollinearity). As a consequence, the model uses the wrong `dof` resulting for instance in computing the incorrect `vcov`. The PR corrects the issue by counting the actual number of parameters estimated from the rank of the factorization.

nalimilan:
OK, thanks. Maybe we should fix the `dof` fallback to use `count(!isnan, coef(obj))` instead?

Nosferican:
Sure. You cannot use `coef` since you cannot distinguish whether an estimate is `0.0` as the estimates value or was it set to `0.0` since it was dropped. You could use the `isnan`, but for the `stderror`. This has the disadvantages of not being a no-op namely `rmul!(invchol(x.pp), dispersion(x, true))`. The current set-up relies on `rank`, a pre-computed stable value, being available which is true for all models that opted-in the rank deficient option.

nalimilan:
Mmm. Maybe we should set coefficients for dropped variables to `NaN`? FWIW, R uses `NA`.

Nosferican:
I think that decision to set them to zero (à la Stata) was to avoid `NaN` propagating to functions that rely on
```
modelmatrix(obj) * coef(model)
```
Regardless, I think this bug is pretty substantial so I would favor merging and releasing a patch for now. The decision on whether coef should `NaN` or zero out would be a breaking change requiring at least next major release.

nalimilan:
OK. But we can tag a major release as quickly as a patch release if there's a consensus.

nalimilan:
BTW, shouldn't we also define `dof` to return the rank? And is the code already correct for models not fitted using Cholesky decomposition?

andreasnoack:
I think rank is the right way of doing it here. We deliberately chose to return zeros such that you could compute the predicted values.

pdeffebach:
quick thought on this:

in stata i am often in a situation where we have a vector of covariates `covariates` and then we want to see heterogeneity in the treatment effect by, say, gender and baseline wealt, which are only included as covariates. So I write a function that does 
```
reg `interactions` `covariates`
```

I don't want to have to think about doing some sort of `setdiff` between the two to make things work. So having `0` and a `NaN` SE is easiest.

Nosferican:
The solution I use in my package is a `rank` `BitVector` which is given by the factorization.
```
F = qr(X, Val(true))
Rank = (abs.(diag(F.R)) .≥ sqrt(eps()))[invperm(F.jpvt)]
if any(~, Rank)
    X = X[:,Rank]
    F = qr(X, Val(true))
end
```
The `F.jpvt` can give you the valid predictors. In my case I just drop them, but keep the information for `coefnames` and such.

test/runtests.jl
```diff
@@ -178,10 +178,10 @@ anorexia = CSV.read(joinpath(glm_datadir, "anorexia.csv"))
     @test isapprox(aicc(gm6), 490.8823884513153)
     @test isapprox(bic(gm6), 501.35662813730465)
     @test isapprox(coef(gm6),
-        [45.674043486911685, -0.5655388496390964, 4.097065528072901, 8.660128180991693])
+        [49.7711090, -0.5655388, -4.0970655, 4.5630627])
```
nalimilan: For reference, these changes are needed because CSV now sorts levels in lexicographic order. So the coefficients correspond to different contrasts than before.

PR #265, opened 2018-11-09
Using dof approach to proper allowrankdeficient values

Nosferican:
This PR fixes the `dof` for `allowrankdeficient` models and also `dof_residual` through the use of `dof`.


Nosferican:
This is ready to be merged.

PR #266, opened 2018-11-12
`:prediction` intervals

mkborregaard:
This implements the promised :prediction interval. Note that this PR is made against @nalimilan 's `nl/predict` branch.


mkborregaard:
I'm guessing I need a test...

src/lm.jl
```diff
@@ -206,17 +206,18 @@ function predict(mm::LinearModel, newx::AbstractMatrix;
...
 
     R = cholesky!(mm.pp).U #get the R matrix from the QR factorization
     residvar = (ones(size(newx,2),1) * deviance(mm)/dof_residual(mm))
     retvariance = (newx/R).^2 * residvar
+    interval == :prediction && (retvariance .+= deviance(mm)/dof_residual(mm))
+    retinterval = quantile(TDist(dof_residual(mm)), (1. - level)/2) * sqrt.(retvariance)
 
-    interval = quantile(TDist(dof_residual(mm)), (1 - level)/2) * sqrt.(retvariance)
-    hcat(retmean, retmean .+ interval, retmean .- interval)
+    hcat(retmean, retmean .+ retinterval, retmean .- retinterval)
```
mkborregaard: I changed `interval` to `retinterval` because we were changing the type of that variable in the function

mkborregaard:
Alright, I added a test. 
One thing - I'm not sure how this works with weighted regression but then I'm not completely sure how that is implemented in GLM.
My main todo (not for this PR) is to finalize confidence intervals for `glm` not just `lm` but that is a litte more involved.

src/lm.jl
```diff
@@ -206,17 +206,18 @@ function predict(mm::LinearModel, newx::AbstractMatrix;
...
-        error("only interval=:confidence is currently implemented") #:predint will be implemented
+    elseif !(interval ∈ (:confidence, :prediction))
+        error("only :confidence and :prediction intervals are defined")
     end
     length(mm.rr.wts) == 0 || error("prediction with confidence intervals not yet implemented for weighted regression")
 
     R = cholesky!(mm.pp).U #get the R matrix from the QR factorization
     residvar = (ones(size(newx,2),1) * deviance(mm)/dof_residual(mm))
     retvariance = (newx/R).^2 * residvar
+    interval == :prediction && (retvariance .+= deviance(mm)/dof_residual(mm))
```
nalimilan: I'd find this more readable with an `if` block. You could also copy the full `(newx/R).^2 * residvar` part so that it's fused with the squaring step.

test/runtests.jl
```diff
@@ -463,6 +463,15 @@ end
                             0.02118806218116165 0.6477623101170038
                             0.8543142404183606  2.564532433982956]
 
+    pred3 = predict(mm, newX, interval=:prediction)
+    @test pred1 == pred3[:, 1] ≈
+        [1.6488076594462182, 0.4706674451801356, 2.5010808086024423,
```
nalimilan: I assume you checked this against another implementation?

mkborregaard: Yes, against R

src/lm.jl
mkborregaard: I did try a few times to include it in the `(newx/R)` block, but found them less readable. I also had a go with the ternary operator
```
retvariance = interval == :prediction ? (newx/R).^2 * residvar .+ deviance(mm)/dof_residual(mm)) : (newx/R).^2 * residvar
```
Do you prefer that?

mkborregaard: The fusion doesn't matter as `deviance(mm)/dof_residual(mm)` is scalar?

nalimilan: It would avoid a second pass over the data (if you also use `.*`, that is). To make it more readable, use an `if` block.

mkborregaard: no - that's a vector-matrix multiply, it shouldn't be broadcast.
I've pushed a commit that rewrites it to use an if block.

mkborregaard:
@nalimilan there's a final consideration to take here - given that we're now returning a Matrix and thus staying fairly close to the R API, should we change the order of columns in the matrix? R uses `[fit lower upper]` whereas we return `[fit upper lower]`. Given that this hasn't really been working it should be safe enough to change it, though I don't see an easy way to change the behaviour to not break anything downstream if we want to be 100% sure.

nalimilan:
Yeah, makes sense. It seems weird to have the upper bound first.

mkborregaard:
Oh sorry that was me being stupid, I misread the code. `retinterval` is negative.
This is good to go :shipit: - should I change this to point to master now that you've merged?

nalimilan:
Yes, if you know how to do it.

mkborregaard:
OK, I rebased and changed the target branch.

mkborregaard:
I'd think this is ready for merge :shipit:

PR #271, opened 2018-12-14
Docs

davidzentlermunro:
It would be helpful to have a bit more guidance around options, for example what format weights should be in as per https://discourse.julialang.org/t/glm-with-weights/18603/4


andreasnoack:
It looks like many of the changes here are unrelated to the docs.

nalimilan:
Is this still relevant? I don't see any docs-related commits.

kleinschmidt:
Closing since there are no doc related changes in these commits.

PR #273, opened 2019-01-07
Implement support for rank deficient Generalized Linear Models

DilumAluthge:
cc: @Nosferican


test/runtests.jl
```diff
@@ -68,6 +68,120 @@ end
                    8.848886704358202, 2.8697881579099085, 11.15107375630744, 11.8392578374927])
 end
 
+@testset "rankdeficient-glm" begin
+    Random.seed!(9999)
+    df = DataFrames.DataFrame()
+    num_rows = 100
+    df[:feature_1] = randn(num_rows)
+    df[:feature_2] = 2 .* df[:feature_1]
```
Nosferican: No need for the broadcast, `2 * df[:feature_1]` does the same.

test/runtests.jl
```diff
@@ -68,6 +68,120 @@ end
...
+    df[:feature_3] = randn(num_rows)
+    df[:label_1] = round.(Int, rand(num_rows))
```
Nosferican: This should probably just be `rand(0:1, num_rows)`.

test/runtests.jl
```diff
@@ -68,6 +68,120 @@ end
...
+    Test.@test(GLM.coeftable(LR_1.model).cols[1][2] < -4e12)
+    Test.@test(GLM.coeftable(LR_1.model).cols[1][3] > 2e12)
+    Test.@test(0.01 < GLM.coeftable(LR_1.model).cols[1][4] < 0.03)
+    Test.@test(0.6 < GLM.coeftable(LR_1.model).cols[4][1].v < 0.8)
+    Test.@test(isnan(GLM.coeftable(LR_1.model).cols[4][2].v))
+    Test.@test(GLM.coeftable(LR_1.model).cols[4][3].v < 1e-90)
+    Test.@test(0.9 < GLM.coeftable(LR_1.model).cols[4][4].v < 1.0)
+
+    Random.seed!(123)
+    df[:label_2] = Vector{Int}(undef, num_rows)
```
Nosferican: Suggestion for clarity,
```
df.label_2 = ifelse.(df.feature_3 .> 0,
                     sample(0:1, weights([0.2, 0.8])),
                     sample(0:1, weights([0.8, 0.2])))
```
or
```
df.label_2 = rand.(Bernoulli.(ifelse.(df.feature_3 .> 0, 0.2, 0.8)))
```

test/runtests.jl
```diff
@@ -68,6 +68,120 @@ end
...
+    Test.@test(-0.003 > GLM.coeftable(OLS_2.model).cols[1][3] > -0.005)
+    Test.@test(0.2 < GLM.coeftable(OLS_2.model).cols[1][4] < 0.3)
+    Test.@test(GLM.coeftable(OLS_2.model).cols[4][1].v < 1e-10)
+    Test.@test(isnan(GLM.coeftable(OLS_2.model).cols[4][2].v))
+    Test.@test(0.8 < GLM.coeftable(OLS_2.model).cols[4][3].v < 1.0)
+    Test.@test(GLM.coeftable(OLS_2.model).cols[4][4].v < 1e-5)
+
+    Random.seed!(123)
+    Test.@test_throws(
+        LinearAlgebra.PosDefException,
```
Nosferican: Same comments as above and make the test sensible to default options (not specifying it to catch any errors if changes to these)
```
@test_throws(PosDefException,
             glm(@formula(label_2 ~ 1 + feature_1 + feature_2 + feature_3),
                 df,
                 Binomial()))
```

src/glmfit.jl
```diff
@@ -198,7 +198,7 @@ end
 dof(x::GeneralizedLinearModel) = dispersion_parameter(x.rr.d) ? length(coef(x)) + 1 : length(coef(x))
 
 function _fit!(m::AbstractGLM, verbose::Bool, maxIter::Integer, minStepFac::Real,
-              convTol::Real, start)
+              convTol::Real, start; allowrankdeficient::Bool = false,)
```
nalimilan:
```suggestion
              convTol::Real, start; allowrankdeficient::Bool = false)
```

src/linpred.jl
```diff
@@ -7,6 +7,14 @@ The effective coefficient vector, `p.scratchbeta`, is evaluated as `p.beta0 .+ f
 and `out` is updated to `p.X * p.scratchbeta`
 """
 function linpred!(out, p::LinPred, f::Real=1.)
+    for i = 1:length(p.delbeta)
+        if isnan(p.delbeta[i])
+            p.delbeta[i] = 0
+        end
+        if isinf(p.delbeta[i])
```
nalimilan:
```suggestion
        elseif isinf(p.delbeta[i])
```
You could also do `x = p.delbeta[i]` to avoid repeating this, and add `@inbounds` to the loop.

src/linpred.jl
```diff
@@ -148,18 +156,23 @@ function delbeta!(p::DensePredChol{T,<:CholeskyPivoted}, r::Vector{T}) where T<:
...
     p
 end
 
-function delbeta!(p::DensePredChol{T,<:CholeskyPivoted}, r::Vector{T}, wt::Vector{T}) where T<:BlasReal
+function delbeta!(p::DensePredChol{T,<:CholeskyPivoted}, r::Vector{T}, wt::Vector{T}; allowrankdeficient::Bool = false) where T<:BlasReal
     cf = cholfactors(p.chol)
     piv = p.chol.piv
     cf .= mul!(p.scratchm2, adjoint(LinearAlgebra.mul!(p.scratchm1, Diagonal(wt), p.X)), p.X)[piv, piv]
-    cholesky!(Hermitian(cf, Symbol(p.chol.uplo)))
+    # cholesky!(Hermitian(cf, Symbol(p.chol.uplo)))
```
nalimilan: Remove this.

test/runtests.jl
```diff
@@ -496,24 +610,24 @@ end
 
     ft = ftest(mod, nullmod)
     @test isapprox(ft.pval[1].v,  2.481215056713184e-8)
-    @test sprint(show, ftest(mod, nullmod)) ==
-        """
+    @test all(split(strip(sprint(show, ftest(mod, nullmod)))) .==
```
nalimilan: Why would this be needed?

src/linpred.jl
```diff
@@ -126,7 +134,7 @@ function cholesky(p::DensePredChol{T}) where T<:FP
 end
 cholesky!(p::DensePredQR{T}) where {T<:FP} = Cholesky{T,typeof(p.X)}(p.qr.R, 'U', 0)
 
-function delbeta!(p::DensePredChol{T,<:Cholesky}, r::Vector{T}) where T<:BlasReal
+function delbeta!(p::DensePredChol{T,<:Cholesky}, r::Vector{T}; allowrankdeficient::Bool = false) where T<:BlasReal
```
nalimilan: No spaces around `=` for consistency with other uses in the file AFAICT.

Nosferican:
@andreasnoack, @dmbates would there be any downside to always use `CholeskyPivoted` instead of having the pivoted and non/pivoted depending on the `allowrankdeficient` argument? The extra memory allocation seems to be quite trivial and it would make interaction with the models easier (e.g., ANOVA).

src/linpred.jl
DilumAluthge: Done.

src/linpred.jl
DilumAluthge: Done.

test/runtests.jl
DilumAluthge: Done.

test/runtests.jl
DilumAluthge: Done.

test/runtests.jl
```diff
@@ -68,6 +68,120 @@ end
...
+    df[:label_1] = round.(Int, rand(num_rows))
+
+    Random.seed!(9999)
+    OLS_1 = GLM.lm(
+        StatsModels.@formula(
+            label_1 ~ 1 + feature_1 + feature_2 + feature_3
+            ),
+        df,
+        true,
+        )
```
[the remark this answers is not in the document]
DilumAluthge: Done. I used the second one.

test/runtests.jl
```diff
@@ -68,6 +68,120 @@ end
...
+    Test.@test(0.4 < GLM.coeftable(OLS_1.model).cols[1][1] < 0.6)
+    Test.@test(isapprox(GLM.coeftable(OLS_1.model).cols[1][2], 0.0))
+    Test.@test(0.003 < GLM.coeftable(OLS_1.model).cols[1][3] < 0.004)
+    Test.@test(0.004 < GLM.coeftable(OLS_1.model).cols[1][4] < 0.005)
+    Test.@test(GLM.coeftable(OLS_1.model).cols[4][1].v < 1e-10)
+    Test.@test(isnan(GLM.coeftable(OLS_1.model).cols[4][2].v))
+    Test.@test(0.8 < GLM.coeftable(OLS_1.model).cols[4][3].v < 1.0)
+    Test.@test(0.9 < GLM.coeftable(OLS_1.model).cols[4][4].v < 1.0)
+
+    Random.seed!(9999)
```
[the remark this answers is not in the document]
DilumAluthge: Done.

test/runtests.jl
```diff
@@ -68,6 +68,120 @@ end
...
+    Test.@test_throws(
+        LinearAlgebra.PosDefException,
+        GLM.glm(
+            StatsModels.@formula(
+                label_1 ~ 1 + feature_1 + feature_2 + feature_3
+                ),
+            df,
+            GLM.Binomial();
```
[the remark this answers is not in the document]
DilumAluthge: Done.

test/runtests.jl
DilumAluthge: When I ran the tests locally, this particular test failed, but only because there was a difference in whitespace. The actual content was the same, but the whitespace was slightly different, so the test failed. I added this code to remove the whitespace and only compare the contents.

Nosferican: This caught me off-guard some time ago. Atom/Juno by default uses uses the Whitespace package that adds a trailing line and trims whitespace in every line after each save. I think is fine to have the modification, but maybe add a comment on the reasoning.

test/runtests.jl
```diff
@@ -68,6 +68,120 @@ end
...
+
+    Random.seed!(9999)
+    LR_1 = GLM.glm(
+        StatsModels.@formula(
+            label_1 ~ 1 + feature_1 + feature_2 + feature_3
+            ),
+        df, GLM.Binomial();
+        allowrankdeficient = true,
+        )
+    Test.@test(-0.1 < GLM.coeftable(LR_1.model).cols[1][1] < 0.0)
```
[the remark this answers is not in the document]
DilumAluthge: Done.

test/runtests.jl
DilumAluthge: I'm going to read the test data from file, instead of generating it randomly. That way, the tests will be deterministic and should always pass.

Nosferican:
It didn't use to, but it does by default now. The code would be simpler and only require using,
```
cholesky!(Hermitian(cf, Symbol(p.chol.uplo)), Val(true), tol = -one(T), check = ~allowrankdeficient)
```
with a default `allowrankdeficient = false` it would check that it is full rank and get the same behavior as we have currently. Allowing for the rank deficient option, would just skip the check, but everything would work the same way for either method instead of having the internals dependent on an argument. If not mistaken, pivoting can help numerically stability even if it is full rank. Sort of a compromised between using QR and unpivoted Cholesky.

DilumAluthge:
> It didn't use to, but it does by default now. The code would be simpler and only require using,
> 
> ```
> cholesky!(Hermitian(cf, Symbol(p.chol.uplo)), Val(true), tol = -one(T), check = ~allowrankdeficient)
> ```
> 
> with a default `allowrankdeficient = false` it would check that it is full rank and get the same behavior as we have currently. Allowing for the rank deficient option, would just skip the check, but everything would work the same way for either method instead of having the internals dependent on an argument. If not mistaken, pivoting can help numerically stability even if it is full rank. Sort of a compromised between using QR and unpivoted Cholesky.

@Nosferican So you are suggesting that we would set a default `allowrankdeficient = true`. I suppose that would be fine, as long as users have the ability to disable pivoting by setting `allowrankdeficient = false`.

DilumAluthge:
@Nosferican Is the pivoting process non-deterministic? I am getting different values for the coefficients when running on my local machine versus Travis. (And I am reading the test data from a `.csv` file, so I thought that the results should be the same.)

Nosferican:
I personally prefer having it as the default similar to most if not all statistical software/packages (Stata, SAS, SPSS, R, etc.), but that would probably be best to discuss it a bit more. However, the proposal at hand is less intrusive at all. The idea is to keep things as they are, but the internals consistent. So the value on the `cholesky` to whether check if it is full rank is set be implied by the `allowrankdeficient` option. In other words, same exact behavior as we currently have. The change is that internally everything uses `PivotedCholesky` rather than sometimes using `PivotedCholesky` and others `Cholesky.`

test/runtests.jl
DilumAluthge: I have added a comment explaining the reasoning.

Nosferican:
The process is deterministic AFAIK. It re-orders the columns based on the having the absolute value of the diagonal in the R matrix from largest to smallest.

DilumAluthge:
> I personally prefer having it as the default similar to most if not all statistical software/packages (Stata, SAS, SPSS, R, etc.), but that would probably be best to discuss it a bit more. However, the proposal at hand is less intrusive at all. The idea is to keep things as they are, but the internals consistent. So the value on the `cholesky` to whether check if it is full rank is set be implied by the `allowrankdeficient` option. In other words, same exact behavior as we currently have. The change is that internally everything uses `PivotedCholesky` rather than sometimes using `PivotedCholesky` and others `Cholesky.`

I see what you mean, you are proposing changing the internals so that the internal type is always `PivotedCholesky`. That sounds good to me, but I'm not sure if I understand exactly how to implement it. I've given you push access to my fork - could you make those changes and push them to the `da/rankdeficient` branch of my fork?

DilumAluthge:
> The process is deterministic AFAIK. It re-orders the columns based on the having the absolute value of the diagonal in the R matrix from largest to smallest.

Hmmm then I need to figure out why the tests pass on my local machine but fail on Travis.

Nosferican:
I would have to look into what drives the differences between 0.7/1.0... I would have expected not to have breaking changes between these for a standard library. I will try to have an in depth look in the next couple days.

test/runtests.jl
```diff
@@ -68,6 +68,170 @@ end
...
 
+@testset "rankdeficient-glm" begin
+    df = CSV.read("rankdeficient-glm.csv")
+
+    OLS_1 = fit(
+        LinearModel,
+        @formula(label_1 ~ 1 + feature_1 + feature_2 + feature_3),
+        df,
+        true)
+    Test.@test( coef(OLS_1)[1] ≈ 0.4814822002921986 )
```
Nosferican: No need for parenthesis.

Nosferican:
I think Github has the squash and merge so it might not be necessary to squash it on your part if it enabled in the repo.

@nalimilan I am good merging this for now after you OK it and working on the internal change to pivoted in a different PR.

test/runtests.jl
DilumAluthge: Done.

DilumAluthge:
I'll squash and rebase locally anyway.

DilumAluthge:
Done. Ready to merge.

DilumAluthge:
Any thoughts on merging this PR?

@nalimilan @Nosferican

src/linpred.jl
```diff
@@ -126,7 +131,7 @@ function cholesky(p::DensePredChol{T}) where T<:FP
 end
 cholesky!(p::DensePredQR{T}) where {T<:FP} = Cholesky{T,typeof(p.X)}(p.qr.R, 'U', 0)
 
-function delbeta!(p::DensePredChol{T,<:Cholesky}, r::Vector{T}) where T<:BlasReal
+function delbeta!(p::DensePredChol{T,<:Cholesky}, r::Vector{T}; allowrankdeficient::Bool=false) where T<:BlasReal
```
Nosferican: Is the argument used in the function at all?

src/linpred.jl
```diff
@@ -148,18 +153,22 @@ function delbeta!(p::DensePredChol{T,<:CholeskyPivoted}, r::Vector{T}) where T<:
...
-function delbeta!(p::DensePredChol{T,<:Cholesky}, r::Vector{T}, wt::Vector{T}) where T<:BlasReal
+function delbeta!(p::DensePredChol{T,<:Cholesky}, r::Vector{T}, wt::Vector{T}; allowrankdeficient::Bool = false) where T<:BlasReal
     scr = mul!(p.scratchm1, Diagonal(wt), p.X)
     cholesky!(Hermitian(mul!(cholfactors(p.chol), transpose(scr), p.X), :U))
     ldiv!(p.chol, mul!(p.delbeta, transpose(scr), r))
     p
 end
 
-function delbeta!(p::DensePredChol{T,<:CholeskyPivoted}, r::Vector{T}, wt::Vector{T}) where T<:BlasReal
+function delbeta!(p::DensePredChol{T,<:CholeskyPivoted}, r::Vector{T}, wt::Vector{T}; allowrankdeficient::Bool = false) where T<:BlasReal
```
Nosferican: I see that it only applies when `DensePredChol[T,<:CholeskyPivoted}`... should be fine for now since it would be the one used when they all use `CholeskyPivoted`... Maybe have `DensePredChol` with a field for `allowrankdeficient` rather than having to pass it?

test/runtests.jl
```diff
@@ -68,6 +68,173 @@ end
...
+    LR_2c = fit(
+        GeneralizedLinearModel,
+        @formula(label_2 ~ 1 + feature_1 + feature_2 + feature_3),
+        df,
+        GLM.Binomial(),
+        GLM.LogitLink(),
+        allowrankdeficient=true)
+
+    # Julia 0.7 and Julia 1.0 give different values for the coefficients
+    # and standard errors.
```
Nosferican: Comment on why if possible.

src/glmfit.jl
```diff
@@ -274,8 +274,8 @@ function _fit!(m::AbstractGLM, verbose::Bool, maxIter::Integer, minStepFac::Real
 end
 
 StatsBase.fit!(m::AbstractGLM; verbose::Bool=false, maxIter::Integer=30,
-              minStepFac::Real=0.001, convTol::Real=1.e-6, start=nothing) =
-    _fit!(m, verbose, maxIter, minStepFac, convTol, start)
+              minStepFac::Real=0.001, convTol::Real=1.e-6, start=nothing, allowrankdeficient::Bool = false,) =
+    _fit!(m, verbose, maxIter, minStepFac, convTol, start; allowrankdeficient = allowrankdeficient,)
```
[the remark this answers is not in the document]
DilumAluthge: I have no idea, that's how it was previously.

src/glmfit.jl
```diff
@@ -312,8 +313,8 @@ function fit(::Type{M},
     end
 
     rr = GlmResp(y, d, l, offset, wts)
-    res = M(rr, cholpred(X), false)
-    return dofit ? fit!(res; fitargs...) : res
+    res = M(rr, cholpred(X, allowrankdeficient,), false)
+    return dofit ? fit!(res; allowrankdeficient = allowrankdeficient, fitargs...) : res
```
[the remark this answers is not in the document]
DilumAluthge: It's not... but I can make `allowrankdeficient` a field in the model, if you want?

The reason that I made it an argument here is because that is how it currently works when setting `allowrankdeficient=true` for linear models (it's an argument, not a field in the model), and I wanted to be consistent.

src/linpred.jl
DilumAluthge: It is not. But if I don't include it in the signature, then you get an error when using `allowrankdeficient=false`, and the tests will fail.

src/glmfit.jl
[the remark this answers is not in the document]
Nosferican: @nalimilan, care to weight in on this? GLM will probably get some needed love after Terms 2.0. We could keep it like how it is now and open an issue for cleaning it in a future PR.

test/runtests.jl
DilumAluthge: Oh god I wish I knew why. I tried experimenting a little bit to see if I could figure out why this happens, but I think that I don't have the requisite knowledge to understanding what's going on.

src/linpred.jl
Nosferican: It probably just needs to dispatch on the correct `Type{p}`... it should be fine for now, but this seems like another pro-having it as a field to simplify the code.

src/glmfit.jl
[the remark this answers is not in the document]
DilumAluthge: I'm happy to make `allowrankdeficient` a field in this PR, or to do it in a future PR. Let me know what you prefer @Nosferican @nalimilan

test/runtests.jl
nalimilan: I don't like this FWIW, if editors are broken they should be fixed instead of making our code more complex. Doctests like this are common in Julia, we really don't want to add comments explaining this every time they occur.

test/runtests.jl
```diff
@@ -68,6 +68,173 @@ end
                    8.848886704358202, 2.8697881579099085, 11.15107375630744, 11.8392578374927])
 end
 
+@testset "rankdeficient-glm" begin
+    df = CSV.read("rankdeficient-glm.csv")
+
+    OLS_1 = fit(
```
nalimilan: I think the style in this package is to have the first argument on the same line as the function, and other arguments aligned with it.

src/glmfit.jl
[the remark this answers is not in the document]
nalimilan: I don't have an opinion on this. Do you have a use case where it would be useful to have this information stored in the model?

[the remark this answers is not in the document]
Nosferican: It would be good for packages such as ANOVA that build on it.

test/runtests.jl
```diff
@@ -68,6 +68,173 @@ end
...
+        LinearModel,
+        @formula(label_1 ~ 1 + feature_1 + feature_2 + feature_3),
+        df,
+        true)
+    Test.@test coef(OLS_1)[1] ≈ 0.4814822002921986
```
nalimilan: No need for `Test.`.

test/runtests.jl
```diff
@@ -68,6 +68,173 @@ end
...
+    Test.@test coef(OLS_1)[2] ≈ 0.0
+    Test.@test coef(OLS_1)[3] ≈ 7.100773448428529e-6
+    Test.@test coef(OLS_1)[4] ≈ 4.306733126416892e-6
+    Test.@test stderr(OLS_1)[1] ≈ 0.013269599196573958
+    Test.@test isnan(stderr(OLS_1)[2])
+    Test.@test stderr(OLS_1)[3] ≈ 8.60301517985726e-6
+    Test.@test stderr(OLS_1)[4] ≈ 3.452441071366933e-6
+    Test.@test coeftable(OLS_1).cols[4][1].v ≈ 6.294030157059516e-271
```
nalimilan: Does this really test something different from the line above? `coeftable` is just a convenience function, it shouldn't be needed to test it here. Same below.

test/runtests.jl
```diff
@@ -68,6 +68,173 @@ end
...
+    Test.@test coeftable(OLS_2).cols[4][1].v ≈ 2.399616832216088e-106
+    Test.@test isnan(coeftable(OLS_2).cols[4][2].v)
+    Test.@test coeftable(OLS_2).cols[4][3].v ≈ 0.19814609164909394
+    Test.@test coeftable(OLS_2).cols[4][4].v ≈ 0.0
+
+    LR_2a = fit(
+        GeneralizedLinearModel,
+        @formula(label_2 ~ 1 + feature_1  + feature_3),
+        df,
+        GLM.Binomial(),
```
nalimilan: No need for `GLM.`.

test/runtests.jl
```diff
@@ -68,6 +68,173 @@ end
...
+        GLM.LogitLink())
+    Test.@test coef(LR_2a)[1] ≈ -5.794539719571302
```
nalimilan: Better test for array equality in a single line.

test/runtests.jl
nalimilan: Maybe something changed in the underlying libraries? Anyway, better do `if VERSION >= 1.0` and use two distinct blocks. But are these two solutions correct?

BTW, why not check the deviance for each model?

DilumAluthge:
I’ll make those changes this weekend!

Dilum Aluthge
M.D. Candidate, Class of 2020
The Warren Alpert Medical School and
The Center for Biomedical Informatics
Brown University
Email: dilum_aluthge@brown.edu

nalimilan:
Any news?

DilumAluthge:
@nalimilan I don't have time to work on this right now, but I'll have some time in 2 weeks, so I'll finish this up then.

DilumAluthge:
I’ve been working a little more on this. On some cases of logistic regression that are rank deficient, this patch fixes the rank deficiency issue (i.e. you no longer get the `PosDefException`), but I then run into the `AssertionError: isfinite(crit)` error discussed in #286. So I think that this solution won’t work until #286 is fixed.

I’ll close this PR. Once #286 is fixed, I’ll make a new PR off of master.

PR #276, opened 2019-01-17
return a NamedTuple, not a Matrix from predict

mkborregaard:
As discussed on Slack.
cc @kleinschmidt @nalimilan @andreasnoack

See also https://github.com/JuliaStats/StatsModels.jl/pull/81.


src/lm.jl
```diff
@@ -220,7 +220,7 @@ function predict(mm::LinearModel, newx::AbstractMatrix;
         error("only :confidence and :prediction intervals are defined")
     end
     retinterval = quantile(TDist(dof_residual(mm)), (1. - level)/2) * sqrt.(retvariance)
-    hcat(retmean, retmean .+ retinterval, retmean .- retinterval)
+    (newy = retmean, lower = retmean .+ retinterval, upper = retmean .- retinterval)
```
nalimilan: Just `y` would look better IMHO.

mkborregaard: fixed

nalimilan: FWIW, R uses `fit`, `lwr` and `upr`. The two latter are clearly bad, but `fit` could also work. Opinions?

kleinschmidt: I don't like "fit" actually, since this is a prediction from new data, not for fitted values.

mkborregaard: mean?

mkborregaard: value is not bad either

mkborregaard:
I've updated the tests too now. Travis test error is unrelated.

src/lm.jl
mschauer: Sorry, what kind of interval does this return when the caller provides `:prediction`?

mkborregaard: The prediction interval

mschauer: Can you elaborate? Does this have to do with taking or not taking uncertainty in the estimated parameters into account?

mkborregaard: It's the 95% confidence interval on unobserved data points, whereas the :confidence interval is the confidence on the fit. So, it incorporates the variance in the data itself

mkborregaard: In Bayesian stats you'd get it by simulating new values from the posterior and taking the (0.025, 0.975)  percentiles

mschauer: Thanks, rephrasing in my own words `:confidence:` gives the confidence interval for the location parameter of the conditional distribution of the response and `:prediction` gives an interval which contains the unobserved response variable with a given probability

mkborregaard: Yes exactly except that those words are Bayesian

nalimilan: Let's go with `value` then?

andreasnoack: I think I prefer `prediction`.

kleinschmidt: I also prefer prediction even if it's a little verbose.  in my own code I tend to go for `pred` but I understand that that is a convention that works for me but maybe not everyone.

nalimilan: Fine with me.

mkborregaard: Great! The only possibly problem with that (and argument for pred or value) is that we now have `:confidence` and `:prediction` intervals, both of which return a variable named `prediction`

mkborregaard: sorry to be so wavering :-/

nalimilan: Fine with me.

mkborregaard: OK it's `prediction` now and tests should pass so feel free to merge when you're happy.

test/runtests.jl
```diff
@@ -498,7 +498,7 @@ end
     @test sprint(show, ftest(mod, nullmod)) ==
         """
                 Res. DOF DOF ΔDOF    SSR    ΔSSR      R²    ΔR²       F* p(>F)
-        Model 1       10   3      0.1283          0.9603
+        Model 1       10   3      0.1283          0.9603
```
mkborregaard: oh wow, Atom is really aggressive on whitespace

nalimilan:
So this should break with the current StatsModels, meaning we should first tag a new release, and then require it from this PR, right?

mkborregaard:
Yes the StatsModels PR is harmless but this one requires it

Issue #283, opened 2019-02-14
Dispersion docstring


tlienart:
The docstring currently specifies

```
Return the estimated dispersion (or scale) parameter for a model's distribution,
generally written σ² for linear models and ϕ for generalized linear models.
It is, by definition, equal to 1 for the Bernoulli, Binomial, and Poisson families.
```

I may have confused myself but shouldn't it be just `σ` for linear models?

nalimilan:
Good catch. AFAICT σ is returned by default since `sqr=false`. See https://github.com/JuliaStats/GLM.jl/pull/284.

PR #284, opened 2019-02-15
Fix docstring for dispersion

nalimilan:
σ² is only returned when `sqr=true`.

Fixes https://github.com/JuliaStats/GLM.jl/issues/283.


dmbates:
Looks good to me in that there is a `sqr::Bool=false` argument

PR #285, opened 2019-02-25
Add basic documentation for `fit` and `glm`

galenlynch:
I was unable to find any documentation on how to fit a GLM to data, beyond a few
examples in the docs. I have added a small amount of documentation for how to
use `fit`, and added them to the Documenter markdown. I have also re-iterated
that `glm` is an alias for `fit`, which wasn't clear to me when I first started
using this package.

### DIFF
--- a/src/glmfit.jl
+++ b/src/glmfit.jl
@@ -304,7 +304,7 @@ Fit a generalized linear model to data.
 # Keyword Arguments
 - `verbose::Bool=false`: Display convergence information for each iteration
 - `maxIter::Integer=30`: Maximum number of iterations allowed to achieve convergence
-- `convTol::Real=1.e-6`: Convergence is achieved when the relative change in
+- `convTol::Real=1e-6`: Convergence is achieved when the relative change in
 deviance is less than this
 - `minStepFac::Real=0.001`: Minimum line step fraction. Must be between 0 and 1.
 - `start=nothing`: Starting values for beta
--- a/src/glmfit.jl
+++ b/src/glmfit.jl
@@ -299,9 +299,9 @@ end
 """
     fit(GeneralizedLinearModel, X, y, d, [l = canonicallink(d)]; <keyword arguments>)
 
-Fit a generalized linear model to data. `X` and `y` can either be a formula and
-a dataframe, respectively, or a matrix and a vector. `l` must be a
-[`Link`](@ref), if supplied.
+Fit a generalized linear model to data. `X` and `y` can either be a matrix and a
+vector, respectively, or a formula and a data frame. `d` must be a
+`UnivariateDistribution`, and `l` must be a [`Link`](@ref), if supplied.
 
 # Keyword Arguments
 - `verbose::Bool=false`: Display convergence information for each iteration
--- a/src/glmfit.jl
+++ b/src/glmfit.jl
@@ -339,7 +339,7 @@ l::Link=canonicallink(d); kwargs...) where {M<:AbstractGLM} =
 """
     glm(F, D, args...; kwargs...)
 
-Fit a generalized linear model to data. Alias for fit(GeneralizedLinearModel, ...).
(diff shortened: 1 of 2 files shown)


src/glmfit.jl
```diff
@@ -296,6 +296,19 @@ function StatsBase.fit!(m::AbstractGLM, y; wts=nothing, offset=nothing, dofit::B
     end
 end
 
+"""
+    fit(AbstractGLM, X, y, d, [l = canonicallink(d)]; <keyword arguments>)
```
nalimilan: I don't think `AbstractGLM` is right: it should be `::Type{<:AbstractGLM}`. But better show the actual concrete type that users can pass, i.e. `GeneralizedLinearModel`.

Can you also give the type and meaning of positional arguments (see `?lm`)?

src/glmfit.jl
```diff
@@ -296,6 +296,19 @@ function StatsBase.fit!(m::AbstractGLM, y; wts=nothing, offset=nothing, dofit::B
...
+
+Fit a generalized linear model to data.
+
+# Keyword Arguments
+- `verbose::Bool=false`: Display convergence information for each iteration
+- `maxIter::Integer=30`: Maximum number of iterations allowed to achieve convergence
+- `convTol::Real=1.e-6`: Convergence is achieved when the relative change in
+deviance is less than this
+- `minStepFac::Real=0.001`: Minimum line step fraction. Must be between 0 and 1.
+- `start=nothing`: Starting values for beta
```
nalimilan: Can you give the type of `start` and its required dimension?

src/glmfit.jl
```diff
@@ -296,6 +296,19 @@ function StatsBase.fit!(m::AbstractGLM, y; wts=nothing, offset=nothing, dofit::B
...
 
+"""
+    fit(AbstractGLM, X, y, d, [l = canonicallink(d)]; <keyword arguments>)
+
+Fit a generalized linear model to data.
+
+# Keyword Arguments
+- `verbose::Bool=false`: Display convergence information for each iteration
+- `maxIter::Integer=30`: Maximum number of iterations allowed to achieve convergence
+- `convTol::Real=1.e-6`: Convergence is achieved when the relative change in
```
nalimilan:
```suggestion
- `convTol::Real=1e-6`: Convergence is achieved when the relative change in
```

src/glmfit.jl
```diff
@@ -323,6 +336,12 @@ d::UnivariateDistribution,
 l::Link=canonicallink(d); kwargs...) where {M<:AbstractGLM} =
     fit(M, float(X), float(y), d, l; kwargs...)
 
+"""
+    glm(F, D, args...; kwargs...)
+
+Fit a generalized linear model to data. Alias for fit(GeneralizedLinearModel, ...).
```
nalimilan:
```suggestion
Fit a generalized linear model to data. Alias for `fit(GeneralizedLinearModel, ...)`.
```

src/glmfit.jl
galenlynch: Done

src/glmfit.jl
galenlynch: Done

src/glmfit.jl
```diff
@@ -296,6 +296,22 @@ function StatsBase.fit!(m::AbstractGLM, y; wts=nothing, offset=nothing, dofit::B
     end
 end
 
+"""
+    fit(GeneralizedLinearModel, X, y, d, [l = canonicallink(d)]; <keyword arguments>)
+
+Fit a generalized linear model to data. `X` and `y` can either be a formula and
+a dataframe, respectively, or a matrix and a vector. `l` must be a
```
nalimilan:
```suggestion
a data frame, respectively, or a matrix and a vector. `l` must be a
```
And `d`?

Issue #286, opened 2019-02-26
Logit link saturating


galenlynch:
I've been running into `ERROR: AssertionError: isfinite(crit)` errors from https://github.com/JuliaStats/GLM.jl/blob/49853b1283da59183a537783c5296aa0d27b194e/src/glmfit.jl#L268 when trying to fit a GLM to Bernoulli response values and continuous predictors (~10,000 samples, each ~90 dimensional) with a logit link. The two distributions are quite well separated. 

The problem seems to originate here: https://github.com/JuliaStats/GLM.jl/blob/49853b1283da59183a537783c5296aa0d27b194e/src/glmfit.jl#L113
Which returns `dμdη = 0` for `η[i] = 753.8917511344454` and `L = LogitLink`, which then causes a `nan` by dividing by zero.

galenlynch:
If I add `eps()` to the denominator here https://github.com/JuliaStats/GLM.jl/blob/49853b1283da59183a537783c5296aa0d27b194e/src/glmfit.jl#L116 then I no longer get `nan`s, but then another error crops up: the cholesky factorization in `delbeta!` fails. By using `verbose = true` for the fitting process, it seems like the issue is one of deciding when the model has converged.

```
1: 1052.1198412358851, 1.3730004602918433
2: 506.7047157958967, 1.0763963871606923
3: 266.04627438164914, 0.9045736196591793
4: 142.88243036403583, 0.8619943243113691
5: 75.36922839123089, 0.8957661291469452
6: 38.35440441500381, 0.9650736216815635
7: 18.3199039714074, 1.0935920010751723
8: 8.002112845754784, 1.2893833571875115
9: 3.2471408364396583, 1.464356567462203
10: 1.2629020585665245, 1.5711739199517762
11: 0.4804890855923021, 1.6283678369296084
12: 0.1806130562166514, 1.6603230998756775
13: 0.06741461316608544, 1.6791380642010896
14: 0.02505538645426235, 1.6906235626877373
15: 0.009286627103756644, 1.6980071638848184
16: 0.0034356012216298507, 1.7030573412565806
17: 0.001269294251962807, 1.706701945838892
18: 0.00046847299122024255, 1.709428880108214
19: 0.00017277232118587915, 1.7115048753453412
20: 6.368105934025369e-5, 1.7130880512326427
21: 2.3461453826452924e-5, 1.7142844519060847
22: 8.64086312206108e-6, 1.715174803145907
[...]

PR #287, opened 2019-02-26
WIP: Consider model converged when deviance is less than an amount

galenlynch:
As described in #286 I've been running into convergence problems with logistic regression, where the model achieves 100% accuracy, the deviance is very small, but the fold-change in deviance is not. This is a crude fix for the problem, which adds another keyword argument to `fit(GeneralizedLinearModel, ...)` to set an absolute lower bound on deviance.


galenlynch:
I think this is superseded by #290

PR #288, opened 2019-03-01
Rename keyword arguments to follow lower case convention

nalimilan:
We could take the occasion to rename arguments. Maybe `tolerance` would be better than `convtol`? FWIW [FixedEffectsModels](https://github.com/matthieugomez/FixedEffectModels.jl/blob/c64231417ab5b874317e4e00f1497b09332e0a60/src/reg.jl#L11) and [StatsModels.org](http://www.statsmodels.org/stable/generated/statsmodels.genmod.generalized_linear_model.GLM.fit.html#statsmodels.genmod.generalized_linear_model.GLM.fit) use `tol`, so we could use that too.

BTW, `allowrankdeficient` is still a positional argument, should we make it a keyword? And if so under what name?

### DIFF
--- a/src/glmfit.jl
+++ b/src/glmfit.jl
@@ -197,15 +197,15 @@ end
 
 dof(x::GeneralizedLinearModel) = dispersion_parameter(x.rr.d) ? length(coef(x)) + 1 : length(coef(x))
 
-function _fit!(m::AbstractGLM, verbose::Bool, maxIter::Integer, minStepFac::Real,
-              convTol::Real, start)
+function _fit!(m::AbstractGLM, verbose::Bool, maxiter::Integer, minstepfac::Real,
+               convtol::Real, start)
 
     # Return early if model has the fit flag set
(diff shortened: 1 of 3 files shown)


ararslan:
> And if so under what name?

`allowrankdeficient` seems as good a name as any

nalimilan:
It's a bit verbose but why not. FWIW R calls this `singular.ok`. Anyway I'll make a separate PR.

What about `minstepfac`? I couldn't find an equivalent in R and StatsModels.org so I guess we should just take the name we find the clearest.

ararslan:
> FWIW R calls this `singular.ok`.

Sounds like we should be consistent with R while also taking advantage of Julia's emoji tab-completion, which would make it `singular❓👌😎`.

ararslan:
> What about `minstepfac`?

I don't know what that does and the name provides no clarification...

nalimilan:
OK to merge then?

PR #290, opened 2019-03-07
Perfectly separable test

### LINKED ISSUE #286: Logit link saturating

IanButterworth:
A WIP candidate for the solution discussed in https://github.com/JuliaStats/GLM.jl/issues/286

TODOs:
- Needs to automatically disabled for all but Bernoulli and Binomial models, or at least able to be overridden
- Extraction of `Y` and `Y_pred` inside `perfectlyseparable()` needs to be handled properly
- Review of method for efficiency
 
Closes #286.

### DIFF
--- a/src/glmfit.jl
+++ b/src/glmfit.jl
@@ -178,7 +178,7 @@ function perfectlyseparable(m::AbstractGLM)
     y_pred = m.rr.y
     y = m.pp.X[:,1]
     order = sortperm(y_pred)
-    count(diff(y).<0) == 0  #perfectly separable? (bool)
+    count(diff(y).!=0) == 1  #perfectly separable? (bool)
 end
 
 function loglikelihood(m::AbstractGLM)
--- a/src/glmfit.jl
+++ b/src/glmfit.jl
@@ -271,6 +271,7 @@ function _fit!(m::AbstractGLM, verbose::Bool, maxIter::Integer, minStepFac::Real
         
         # Test for separable
         perfsep = perfectlyseparable(m)
+        perfsep && @warn "Iteration halted early: The estimated coefficients perfectly separate results from each class, meaning that the theoretical best estimates are not finite."
         
         if crit < convTol || dev == 0 || perfsep
             cvg = true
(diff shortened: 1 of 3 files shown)


src/glmfit.jl
```diff
@@ -174,6 +174,13 @@ confint(obj::AbstractGLM) = confint(obj, 0.95)
 
 deviance(m::AbstractGLM) = deviance(m.rr)
 
+function perfectlyseparable(m::AbstractGLM)
+    y_pred = m.rr.y
+    y = m.pp.X[:,1]
+    order = sortperm(y_pred)
+    count(diff(y[order]).!=0) == 1  #perfectly separable? (Given only 1 class interface in score-based sort) (bool)
```
nalimilan: Turn this into an explicit loop to avoid an allocation. That will also allow returning early as soon as you find two consecutive elements that aren't equal.

src/glmfit.jl
```diff
@@ -174,6 +174,13 @@ confint(obj::AbstractGLM) = confint(obj, 0.95)
 
 deviance(m::AbstractGLM) = deviance(m.rr)
 
+function perfectlyseparable(m::AbstractGLM)
+    y_pred = m.rr.y
```
nalimilan: No need to create that variable since it's used only once.

src/glmfit.jl
ianshmean: Ok. Updated

ianshmean:
I've added a test for this issue. There may be a better test for proving success with the perfectly separable example dataset.

ianshmean:
Ok, those sound like good points @andreasnoack. I'm going to bow out and let someone with better understanding of the package implement the suggested method.

IanButterworth:
Stale

PR #293, opened 2019-03-21
Add more documentation for keyword arguments used to fit GLMs

galenlynch:
Some important keyword arguments were not specified for
`fit(GeneralizedLienarModel, ...)`, such as `offset`. This commit copies the
documentation from `GlmResp` to the corresponding keyword arguments of `fit`.

### DIFF
--- a/src/glmfit.jl
+++ b/src/glmfit.jl
@@ -306,7 +306,8 @@ vector, respectively, or a formula and a data frame. `d` must be a
 # Keyword Arguments
 - `dofit::Bool=true`: Determines whether model will be fit
 - `wts::Vector=similar(y,0)`: prior case weights. Can be length 0.
-- `offset::V=similar(y,0)`: offset added to `Xβ` to form `eta`.  Can be of length 0
+- `offset::Vector=similar(y,0)`: offset added to `Xβ` to form `eta`.  Can be of
+length 0
 - `verbose::Bool=false`: Display convergence information for each iteration
 - `maxIter::Integer=30`: Maximum number of iterations allowed to achieve convergence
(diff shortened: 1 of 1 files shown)


src/glmfit.jl
```diff
@@ -158,6 +158,37 @@ mutable struct GeneralizedLinearModel{G<:GlmResp,L<:LinPred} <: AbstractGLM
     fit::Bool
 end
 
+"""
```
nalimilan: This is already documented in StatsBase, so I'm not sure it's useful to repeat it here. Also to access p-values we should rather implement `pvalue` than recommend accessing fields manually.

galenlynch: It's very confusing to have to look in a bunch of different places to figure out how to use GLM.jl. There is no link to StatsBase in the existing (stable) documentation, so even figuring out how to access the coefficients of a fit with `coef` was quite cryptic to me. I was only able to figure out how to use this package by going through the examples, which is where `coef` is mentioned. In contrast, `coeftable` is nowhere to be found. If the documentation for this package will be largely in StatsBase rather than GLM, that should be made much more explicit in some way.

I guess I can understand not wanting to document access to pvalues through a means you don't intend to support, but as it stands this is the only way to access them.

I'll revert this commit, but I was only able to figure out how to access z-values and p-values by grepping through the source code. Even if I had known to look at StatsBase for documentation on GLM.jl, I wouldn't have guessed that `coeftable` would be how I could gain access to these values.

nalimilan: Yes it's a general issue when you implement generic functions defined elsewhere. There might be a setting in Documenter.jl to include a docstring from another package we depend on in the online manual.

The docstring could also be made more explicit in StatsBase.

PR #298, opened 2019-03-28
Update manual.md to include `r2` for R^2 computation

xiaodaigh:


nalimilan:
This also has to apply to GLMs. I've made edits.

BTW, there are many more functions which could be documented. Maybe we should just point to the StatsBase manual for the full list?

KronosTheLate:
Still can not see documentation for `r2` in GLM.jl. That should be added, right?

nalimilan:
Yes, that was my last comment. We could use an approach similar to https://github.com/JuliaStats/StatsBase.jl/pull/744/ to include docstrings from StatsAPI in the GLM manual.

PR #299, opened 2019-03-30
Add confidence intervals to coeftable

nalimilan:
Confidence intervals are usually more informative than p-values, which are increasingly criticized. For example the recent American Statistical Association statement at https://www.amstat.org/asa/files/pdfs/P-ValueStatement.pdf) advises to report CIs (or other kinds of intervals).

Also make level a keyword argument for consistency with new StatsBase API (https://github.com/JuliaStats/StatsBase.jl/pull/406).


nalimilan:
@pdeffebach Can you suggest a better name for the CI columns? I know you didn't like the ones I chose. Cc @matthieugomez since I took the names from FixedEffectsModels.

pdeffebach:
I would just say "Lower 95%\nCI bound" if feasible.

nalimilan:
That's not feasible unless we change `CoefTable` printing again to support that. But honestly I don't think that's a good idea: "CI bound" would appear twice (for upper and for lower), and that would take a full line just for these two words.

pdeffebach:
Sounds good. I think the user will know what it means. Apologies for the bikeshedding.

src/deprecated.jl
```diff
@@ -1 +1,5 @@
-@Base.deprecate predict(mm::LinearModel, newx::AbstractMatrix, interval::Symbol, level::Real = 0.95) predict(mm, newx; interval=interval, level=level)
\ No newline at end of file
+using Base: deprecate
```
ararslan: Doesn't seem to be used.

src/glmfit.jl
```diff
@@ -158,19 +158,21 @@ mutable struct GeneralizedLinearModel{G<:GlmResp,L<:LinPred} <: AbstractGLM
...
 end
 
-function coeftable(mm::AbstractGLM)
+function coeftable(mm::AbstractGLM; level::Real=0.95)
     cc = coef(mm)
     se = stderror(mm)
     zz = cc ./ se
-    CoefTable(hcat(cc,se,zz,2.0 * ccdf.(Ref(Normal()), abs.(zz))),
-              ["Estimate","Std.Error","z value", "Pr(>|z|)"],
+    p = 2.0 * ccdf.(Ref(Normal()), abs.(zz))
```
ararslan:
```suggestion
    p = 2 * ccdf.(Ref(Normal()), abs.(zz))
```
Avoids unnecessary promotion to `Float64`.

src/deprecated.jl
nalimilan: Apparently that's equivalent to `@deprecate`, but let's change that.

ararslan: Is it? I think this worked before because `@deprecate` is exported from Base.

nalimilan: Woops, you're right. :-D

nalimilan:
OK to merge now?

PR #301, opened 2019-04-02
Add basic documentation to negbin

galenlynch:


src/negbinfit.jl
```diff
@@ -34,6 +34,20 @@ function mle_for_θ(y::AbstractVector, μ::AbstractVector, wts::AbstractVector;
...
+           initialθ::Real=Inf, maxIter::Integer=30, convTol::Real=1.e-6,
+           verbose::Bool=false, kwargs...)
+
+Fit a negative binomial generalized linear model to data, while simultaneously
+estimating the shape parameter θ. `maxIter`, `convTol`, and `verbose` correspond
+to the keyword arguments of the same name used by [`glm`](@ref). Extra arguments
+and keyword arguments will be passed to [`glm`](@ref).
+
+# Keyword Arguments
+- `initialθ::Real=Inf`: Starting values for shape parameter θ. If it is `Inf`
```
nalimilan: Singular "value".

Also, better document other arguments here rather than above, that's more logical.

nalimilan:
Thanks. Note that with https://github.com/JuliaStats/GLM.jl/pull/288, the argument names will switch to lower case.

src/negbinfit.jl
galenlynch: done

galenlynch:
Thanks for doing that!

nalimilan:
Sorry, this was completely forgotten. Would you mind rebasing against master?

galenlynch:
I think it's done, but it's been long enough that I've forgotten everything I knew about this function, and have only glanced at the new keyword arguments.

PR #302, opened 2019-04-03
Pass stopping criteria to mle_for_θ

galenlynch:
`negbin` relies on helper function `mle_for_θ` to estimate the parameter θ, but
the stopping criteria `maxIter` and `convTol` were not being passed from
`negbin` to `mle_for_θ`.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -385,6 +385,19 @@ end
     @test isapprox(coef(gm20)[1:7],
         [2.894527697811509, -0.5693411448715979, 0.08238813087070128, -0.4484636623590206,
          0.08805060372902418, 0.3569553124412582, 0.2921383118842893])
+
+    # Issue #302
+    let df = DataFrame(
+        ;y = [1, 1, 0, 2, 3, 0, 0, 1, 1, 0, 2, 1, 3, 1, 1, 1, 4, 1, 2, 4]
+    )
+        for maxIter in [30, 50]
+            try
+                negbin(@formula(y ~ 1), df, maxIter = maxIter)
+            catch err
+                @test err.msg == "failure to converge after $(maxIter) iterations."
+            end
+        end
+    end
 end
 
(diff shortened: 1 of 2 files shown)


nalimilan:
Thanks. Can you also add a test?

galenlynch:
A test to ensure the keyword arguments are being passed?

nalimilan:
Yes, any test that ensures this won't regress in the future.

galenlynch:
I don't know a good way to test that the keyword arguments are passed. Right now I added a test that will fail to converge, and look at the iteration number that it failed it. However, I don't know if these tests work because for whatever reason I can't run tests for GLM.jl on my computer (#304).

test/runtests.jl
```diff
@@ -385,6 +385,20 @@ end
...
+
+    # Issue #302
+    let df = DataFrame(
+        ;y = [1, 1, 0, 2, 3, 0, 0, 1, 1, 0, 2, 1, 3, 1, 1, 1, 4, 1, 2, 4]
+    )
+        for maxIter in [30, 50]
+            try
+                negbin(@formula(y ~ 1), df, maxIter = maxIter)
+            catch err
+                errmsg = sprint(showerror, err)
```
nalimilan: `err.iters` should give you the value directly.

test/runtests.jl
```diff
@@ -385,6 +385,20 @@ end
     @test isapprox(coef(gm20)[1:7],
         [2.894527697811509, -0.5693411448715979, 0.08238813087070128, -0.4484636623590206,
          0.08805060372902418, 0.3569553124412582, 0.2921383118842893])
+
+    # Issue #302
+    let df = DataFrame(
```
nalimilan: Better use `@testset`, and define `df` inside the block. Also `DataFrame(y=...)` is enough (keep it on a single line).

nalimilan:
Woops, I should have run tests again as they now fail due to changes on master since then. See https://github.com/JuliaStats/GLM.jl/pull/355.

PR #303, opened 2019-04-03
Emit warning message from mle_for_θ upon convergence failure

galenlynch:
This commit provides some hint to the user as to what went wrong when estimating θ fails. I'm not sure that convergence only fails when the dispersion parameter grows very large, but in my hands this is the common failure mode. Googling similar problems for R suggests that having the dispersion parameter grow without bound is a common failure mode when fitting a negative binomial to data.


src/negbinfit.jl
```diff
@@ -30,7 +30,13 @@ function mle_for_θ(y::AbstractVector, μ::AbstractVector, wts::AbstractVector;
         end
         θ = θ - δ
     end
-    converged || throw(ConvergenceException(maxIter))
+    if !converged
+        @warn """
```
nalimilan: Triple quotes include line breaks. AFAIK you need to use normal quotes, possibly splitting message over several lines using `*`.

But I think we'd better pass the message to `ConvergenceException` than print a warning. Clearly that's not a warning, that's an error. And if the caller catches it, the warning shouldn't be printed to the output. Would you make a PR against StatsBase to add a `msg::String` field to `ConvergenceException`?

galenlynch:
I've changed the PR to use the `msg` field of `ConvergenceException` added by JuliaStats/StatsBase.jl#484. I guess this will have to wait until that change makes it into the next tagged version. After that, we can bump the minimum required version of `StatsBase` for GLM.jl and use the new field.

nalimilan:
StatsBase has been tagged. Can you bump the dependency?

galenlynch:
It seems like StatsBase requires a Distributions version in the range of `0.1.0-0.1.2`, while GLM requires a Distributions version to be in the range of `0.16.0-*`.

nalimilan:
Will be fixed by https://github.com/JuliaStats/GLM.jl/pull/307.

nalimilan:
Can you rebase? The StatsBase dependency should be updated in Project.toml now rather than REQUIRE.

nalimilan:
Ah, we also need https://github.com/JuliaStats/GLM.jl/pull/299 to fix CI.

Project.toml
```diff
@@ -33,6 +33,6 @@ Distributions = "0.16, 0.17, 0.18, 0.19"
 RDatasets = "0.5, 0.6"
 Reexport = "0.1, 0.2"
 SpecialFunctions = "0.6, 0.7"
-StatsBase = ">= 0.30"
+StatsBase = "0.30"
```
galenlynch: So with each new version of StatsBase you'll have to expand this?

nalimilan: Yes, since this is considered as a major release. See https://github.com/JuliaRegistries/General/pull/227 and https://github.com/JuliaRegistries/General/pull/256.

nalimilan:
Thanks!

PR #307, opened 2019-04-23
Use Project.toml instead of REQUIRE

nalimilan:


Project.toml
```diff
@@ -0,0 +1,38 @@
...
+RDatasets = "ce6b1742-4840-55fa-b093-852dadbb1d8b"
+Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
+
+[targets]
+test = ["CategoricalArrays", "CSV", "DataFrames", "RDatasets", "Test"]
+
+[compat]
+CategoricalArrays = "0.3, 0.4, 0.5"
+CSV = "0.2, 0.3, 0.4"
+DataFrames = "0.11, 0.12, 0.13, 0.14, 0.15, 0.16, 0.17"
```
ararslan: Is there really not a more concise way of specifying this?

nalimilan: @fredrikekre can tell.

fredrikekre: No.

PR #309, opened 2019-04-28
More small improvements to CoefTable

nalimilan:
Use the new argument in StatsBase 0.30 to print t/z values with fewer digits. Rename the "t/z" value column to just "t/z" to avoid wasting space which isn't needed to print the digits (removing an inconsistency with other columns which use upper case). Also rename the "Estimate" column to "Coef.", which is more consistent with
the name of the accessor function `coef`, clearer and more correct since all columns are actually estimates.

Update doctests output in manual.

(Based on https://github.com/JuliaStats/GLM.jl/pull/308 to avoid test failures.)


nalimilan:
I'll merge in a few days if nobody objects.

palday:
Why is Travis failing for pr but not push? 

If you're shortening names, then why not `SE` instead of `Std. Error`? 

(I have no strong feelinga either way, but perhaps others do.)

nalimilan:
> Why is Travis failing for pr but not push?

I guess master wasn't in a good state at the time. I've restarted CI and it passed.
 
> If you're shortening names, then why not `SE` instead of `Std. Error`?
> 
> (I have no strong feelinga either way, but perhaps others do.)

None of the other columns use acronyms for their names so I'd rather keep things consistent. I just wanted to shorten "t value" in particular since the name was really longer than the useful number of digits for the column.

OK to merge before I forget it for one more year? ;-) BTW, do the names sound OK to you for MixedModels to adopt the same terminology? Better keep things consistent.

palday:
I would give @dmbates two days to express his opinion, but I suspect it will be "shrug".

dmbates:
Seems fine to me.  There are a couple of instances in the examples where the t/z statistic is printed with two significant digits then the p value is printed with four significant digits but that is probably not worth making a fuss about.  Far too much time is spent obsessing about p-values as it is.

nalimilan:
OK, merging then.

PR #312, opened 2019-06-14
RFC: Compute working weights and residuals more carefully to reduce underflow

### LINKED ISSUE #286: Logit link saturating

andreasnoack:
This is another attempt at fixing #286. A little explanation of the changes here is probably required. The models are fitted with the iteratively reweighted least squares (IRLS) algorithm which is a version of Newton's method where the Fisher information is used as the Hessian. Each update can be efficiently computed as a weighted least squares problem, i.e. the update becomes
image

so the two quantities to compute are the weights
image

and the weighted residuals
image

where `μᵢ=μ(ηᵢ)` with `ηᵢ=xᵢ'β` being the linear predictor and `μ` the inverse link function. The current approach in GLM.jl has been to compute `μ`, `dμdη`, and `Var(yᵢ)` and then compute the ratios defining the weights and residuals.

This approach fairly frequently runs into problems in the Bernoulli case (`yᵢ` is either zero or one and `Var(yᵢ) = μᵢ(1-μᵢ)`). The reason is that while `wᵢ` is bounded and `rʷᵢ` moves modestly, the numerator and denominator become small for extreme values of `ηᵢ` causing `NaN`s which is essentially the problem in #286. However, the limits are actually well known for each of the link functions so by using dispatch when computing the weights and residuals it is possible to compute the right limits instead of getting `NaN`. The simplest example here might be for the logit model where it can easily be verified that the residual is either -1, 0, or 1 in the limit.

@dmbates It would be great if you could sanity check what I'm doing here. I'm thinking that it might generally be a good idea to focus the specialized methods on the weights and residuals calculations directly instead of the derivative of the inverse link and the variance function since the two former seem to behave much better. Notice, that I've currently changed what `inverselink` returns for `Link01` links. I'm not sure what the implications are for `MixedModels`. I've also removed the clamping used in `CloglogLink` because they were no longer needed here once I took care of the limits computations for the weights and residuals.

Another part of the issue in #286 is that when there is a perfect fit then a relative convergence criterion isn't working well. Therefore, I've also introduced changed `tol` to `rtol` and introduced `atol`. In total, I believe this should fix #286 and I've added tests for all the `Link01`s to verify that it's the case.

### DIFF
--- a/src/negbinfit.jl
+++ b/src/negbinfit.jl
@@ -52,15 +52,15 @@ function negbin(F,
         minstepfac = kwargs[:minStepFac]
     end
     if haskey(kwargs, :convTol)
-        Base.depwarn("'convTol' argument is deprecated, use 'tol' instead", :negbin)
-        tol = kwargs[:convTol]
+        Base.depwarn("`convTol` argument is deprecated, use `atol` and `rtol` instead", :negbin)
+        rtol = kwargs[:convTol]
     end
     if !issubset(keys(kwargs), (:maxIter, :minStepFac, :convTol))
         throw(ArgumentError("unsupported keyword argument"))
     end
     if haskey(kwargs, :tol)
         Base.depwarn("`tol` argument is deprecated, use `atol` and `rtol` instead", :negbin)
-        tol = kwargs[:tol]
+        rtol = kwargs[:tol]
     end
 
     maxiter >= 1 || throw(ArgumentError("maxiter must be positive"))
--- a/src/negbinfit.jl
+++ b/src/negbinfit.jl
@@ -13,7 +13,7 @@ function mle_for_θ(y::AbstractVector, μ::AbstractVector, wts::AbstractVector;
 
     unit_weights = length(wts) == 0
     if unit_weights
-        n = length(y) 
+        n = length(y)
         θ = n / sum((yi/μi - 1)^2 for (yi, μi) in zip(y, μ))
     else
         n = sum(wts)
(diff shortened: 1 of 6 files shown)


dmbates:
The changes look reasonable to me, although I would still advocate for defining methods for `updateμ!` that dispatch on both `D` and `L` when `L` is the canonical link for `D`.  Just the two cases
```
updateμ!(r::GlmResp{V,<:Union{Binomial,Bernoulli},LogitLink}, η::V) where V <: AbstractVector{<:AbstractFloat}
```
and `Poisson, LogLink` would cover the majority of the usage and could be defined compactly.  It is fine to have the flexibility of definition through `linkfun`, `linkinv`, ... for the general case but why not make these widely used cases where terms cancel self contained.

src/negbinfit.jl
```diff
@@ -34,36 +34,47 @@ function mle_for_θ(y::AbstractVector, μ::AbstractVector, wts::AbstractVector;
...
-        Base.depwarn("'convTol' argument is deprecated, use 'tol' instead", :fit!)
+        Base.depwarn("'convTol' argument is deprecated, use 'tol' instead", :negbin)
         tol = kwargs[:convTol]
     end
     if !issubset(keys(kwargs), (:maxIter, :minStepFac, :convTol))
         throw(ArgumentError("unsupported keyword argument"))
     end
+    if haskey(kwargs, :tol)
+        Base.depwarn("`tol` argument is deprecated, use `atol` and `rtol` instead", :negbin)
+        tol = kwargs[:tol]
```
nalimilan: I guess this should be `rtol`? Same above if we want the `convTo` deprecation to work.

test/runtests.jl
```diff
@@ -620,3 +615,12 @@ end
     @test dof_residual(model1) == dof_residual(model2)
     @test dof_residual(model3) == dof_residual(model4)
 end
+
+@testset "Issue #286 (separable data)" begin
+    x  = rand(1000)
+    df = DataFrame(y = x .> 0.5, x₁ = x, x₂ = rand(1000))
+    @testset "Binomial with $l" for l in (LogitLink(), ProbitLink(), CauchitLink(), CloglogLink())
+        @test glm(@formula(y ~ x₁ + x₂), df, Binomial(), l, maxiter=40) isa StatsModels.DataFrameRegressionModel
```
nalimilan: How about using a stricter test of the results?

andreasnoack:
> The changes look reasonable to me, although I would still advocate for defining methods for `updateμ!`...It is fine to have the flexibility of definition through `linkfun`, `linkinv`, ... for the general case but why not make these widely used cases where terms cancel self contained.

The reason was mainly that much of `updateμ!` would be repeated in each of the methods so I thought it would be good to share that part. Just to make sure that the proposal is clear: there is a fallback method so the old approach still works.

PR #321, opened 2019-07-19
typo fix

tlienart:
Wiebull -> Weibull


andreasnoack:
Thanks. The Anglo-Saxons always struggle with Germanic "ei". The test errors are unrelated so I'll merge but we should figure out why things broke.

Issue #322, opened 2019-07-20
Tests currently failing


andreasnoack:
There seems to be an issue somehow related to CSV. See https://travis-ci.org/JuliaStats/GLM.jl/jobs/561284900. This can probably be fixed by allowing more recent version in compat but it still shouldn't fail.

cc @quinnj

quinnj:
The 0.4 release is pretty old at this point; pre-Project.toml files even. So for that release, CSV.jl didn't have Parsers.jl in it's explicit dependencies (hence the warning), but it also is trying to use an older API of Parsers.jl that doesn't exist, since 0.3.6 gets installed (which is the current, latest version). So yeah, the best solution is to just bump the CSV.jl requirement; I can make a quick PR.

andreasnoack:
Fixed by #323.

PR #323, opened 2019-07-21
Bump required CSV.jl dependency

quinnj:


quinnj:
Should fix https://github.com/JuliaStats/GLM.jl/issues/322

PR #328, opened 2019-08-22
Allow for more recent versions of dependencies

nalimilan:
Adjusting the upper bound of all dependencies every time they tag a new minor release
is a nightmare, and since we don't do we hold back other packages. It's much simpler to
fix the bounds in the relatively rare event when things break.


gragusa:
Can this be tagged?

andreasnoack:
Should be obsolete with https://github.com/JuliaStats/GLM.jl/pull/329 and version 1.3.1. Are the current bounds causing problems?

gragusa:
The problem is `SpecialFunctions.jl`. GLM requires <=0.7.

PR #329, opened 2019-08-22
Update version bounds on dependencies

andreasnoack:
Alternative to #328


nalimilan:
Should we tag a new version, or simply change bounds for the existing one in the registry?

andreasnoack:
Let's tag a new version. Since we just updating the bounds without removing any version, I'd say that bugfix is fine.

nalimilan:
https://github.com/JuliaStats/GLM.jl/pull/330

PR #335, opened 2019-10-05
Add test for predict with DataFrame

nalimilan:
Previously not covered. Also return a vector rather than a column matrix for confidence bounds.

Will fail until https://github.com/JuliaStats/StatsModels.jl/pull/160 is merged and released, and https://github.com/JuliaStats/GLM.jl/pull/336 merged.


nalimilan:
This is fixed on master already.

PR #336, opened 2019-10-06
Allow CategoricalArrays 0.7

nalimilan:


nalimilan:
CI failure is due to CSV.jl not supporting CategoricalArrays 0.7, which sounds too complex for the dependency resolver.

PR #337, opened 2019-10-11
ftest improvements

### LINKED ISSUE #275: Ftest printing
Inspired by the discussion in #274, we could consider changing the F-test output from
```julia
        Res. DOF DOF ΔDOF    SSR    ΔSSR     R²    ΔR²       F* p(>F)
Model 1       10   3      0.1283         0.9603
Model 2       11   2   -1 3.2292 -3.1008 0.0000 0.9603 241.6234 <1e-7
```
to
```julia
Model Res. DOF DOF ΔDOF    SSR    ΔSSR     R²    ΔR²       F* p(>F)
    1       10   3      0.1283         0.9603
    2       11   2   -1 3.2292 -3.1008 0.0000 0.9603 241.6234 <1e-7
```
I'm not just suggesting this to because it would work around the Documenter parser issue but because of the redundancy in repeating `Model` for each line.

nalimilan:
Allow passing models from simpler to more complex, giving equivalent tests but with
reversed signs for ΔDOF, ΔSSR and ΔR².
Stop printing degrees of freedom, as residual DOF and difference in DOF seems enough.
Print two spaces instead of one between columns to make table more readable, and add
horizontal lines to match the `CoefTable` printing.
Make struct type stable by filling tuples with an initial NaN entry when needed.
Make tests cover more situations by having a model with ΔDOF of 2.
Small code improvements.

Fixes #275 (Cc: @andreasnoack)


nalimilan:
Cc: @LewisHein

src/ftest.jl
```diff
@@ -57,75 +55,95 @@ this is an ANOVA, our null hypothesis is that `Result ~ 1` fits the data as well
...
+julia> bigmodel = lm(@formula(Result ~ 1 + Treatment + Other), dat);
 
-julia> ftest(bigmodel.model, model.model, nullmodel.model)
-        Res. DOF DOF ΔDOF    SSR    ΔSSR     R²    ΔR²       F*  p(>F)
-Model 1        9   4      0.1038         0.9678
-Model 2       10   3   -1 0.1283 -0.0245 0.9603 0.0076   2.1236 0.1790
-Model 3       11   2   -1 3.2292 -3.1008 0.0000 0.9603 241.6234  <1e-7
+julia> ftest(nullmodel.model, model.model)
+──────────────────────────────────────────────────────────────────────
+     Res. DOF  ΔDOF     SSR     ΔSSR       R²     ΔR²        F*  p(>F)
```
nalimilan: Something to discuss is whether to print the residual degrees of freedom of the consumed degrees of freedom. Other implementations are inconsistent in that regard.

It could make more sense to print the consumed ones, as they give an indication about the complexity of the model, while the residual degrees of freedom reflect the sample size (which is common to all models). It also takes fewer characters in the heading (just "DOF"). Opinions?

nalimilan:
Another question is whether we should continue allowing to pass a single model. That doesn't sound terribly useful, and we could eventually do a test against the null model in that case instead (which is what `ftest` in the lmtest R package does).

But technically speaking, throwing an error would be breaking. Maybe just print a deprecation warning for now?

kleinschmidt:
My feeling is that passing a single model doesn't provide a useful test and if you're just doing it to get a convenient printout of the DOF, RSS, etc. statistics that's too punny.  So we should deprecate that.

src/ftest.jl
```diff
@@ -28,15 +28,14 @@ _diffn(t::NTuple{N, T}) where {N, T} = ntuple(i->t[i]-t[i+1], N-1)
...
+    ftest(mod::LinearModel...; atol::Real=0.0)
 
-For each sequential pair of linear predictors in `mod`, perform an F-test to determine if
-the first one fits significantly better than the next.
+For each sequential pair of linear models in `mod...`, perform an F-test to determine if
+the one model fits significantly better than the other. Models must have been fitted
+on the same data, and be nested either in forward or backward direction.
 
-A table is returned containing residual degrees of freedom (DOF), degrees of freedom,
+A table is returned containing residual degrees of freedom (DOF),
```
kleinschmidt: It's consumed now, not residual right?

nalimilan: Good catch, fixed.

compleathorseplayer:
Hi.   Pardon me for being thick, but how does one easily get the fstat which is routinely printed out on all R regressions?
[i.e., for elementary students who won't initially understand the theory of  '2 model comparison']

PR #339, opened 2019-10-18
Get rid of TableRegressionModel by storing formula directly in model

nalimilan:
This is an exploration of what is needed to get rid of `TableRegressionModel` (in GLM to help designing the correct API in StatsModels (https://github.com/JuliaStats/StatsModels.jl/issues/32). Overall changes are incredibly limited.

The approach taken here is to have packages define `fit(::Type{M}, f::FormulaTerm, data, ...)` for their model type `M` and call `modelmatrix(f::FormulaTerm, data, contrasts)` to get the model matrix from that. Model objects then have to store the formula, which can be retrieved using the `formula` function which is added to the `StatisticalModel` API. `fit` can still allow passing matrices directly without any formula, in which case it is `nothing`.

Thanks to the `formula` generic function, fallback definitions can be provided in StatsModels for many functions, like `coefnames` or `predict`. The latter is the most tricky case as missing values have to be handled: the StatsModels fallback method allocates vectors and passes them to `predict!` (which has to be defined by the package), and returns them.

For simplicity, the PR currently includes a few functions which should defined in StatsModels once we have agreed on the API. `modelmatrix` already exists there and just needs to handle missing values.

------
For reference, here are notes @kleinschmidt wrote after we discussed this issue:

1. Remove the `TableStatisticalModel` wrapper
2. Repurpose `ModelFrame` as a data transformer that keeps track of, e.g., which rows in the model matrix have missing values (in order to map the missings-free array rows onto the underlying table rows).
3. By including the formula in the model frame, this allows for removing rows from the arrays that are introduced by terms like lead and lag, as long as the filtering of missing values happens after `modelcols` is called
4. Models are responsible for holding onto the model frame.  When the input data is not in the form of a table when the model is constructed, a fallback `IdentityTransformer` will be a placeholder.  (Or maybe `Nothing`).
5. Part of the api will be an accessor function `transformer(::MyModel)` which will return the transformer (`ModelFrame`, `IdentityTransformer`, etc.).  This then allows StatsModels.jl to define generic methods like `coeftable` and `coefnames` that are sensitive to the presence of a model frame and the formula it contains.

And an updated summary:
1. maybe we want to think of terms as "lazy views" into the underlying data.  with this philosophy, we could think of having an additional step of `apply_data`, which annotates each term with a pointer to the underlying data, which are then coalesced into array-likes with `modelcols` or `Tables.matrix` or something
2. then we can think of missing-dropping as another lazy view that only views rows that have no missings.  we can apply this to the original table OR to the "table" that's generated on the output end.  this would be implemented as a wrapper around the table
3. treating the missing-dropping as a view has the advantage of allowing us to naturally write predicted values into the corresponding rows in the original table (or at least in a way that aligns with those rows).
4. this also has the advantage of making the formula (and even individual terms) completely self-contained data transformers which eliminates the need for a `ModelFrame` `struct`


nalimilan:
Yes, actually it doesn't appear that `ModelFrame` and `ModelMatrix` are needed now. So indeed packages are free to avoid `modelmatrix` and call individual steps instead. Though it would be good to make `modelmatrix` as general as possible so that most packages can use it: that makes it easier to change the pipeline if needed without breaking too many packages.

Regarding the missing mask, that indeed raises an issue with the current state of the PR: it doesn't store the mask (nor the data). That means functions like `fitted` (not sure if there are others) return the fitted values without the missing values from the original data.

There's also a related issue with the `predict`/`predict!` system I described above: in that approach, `predict` includes missing values, but `predict!` skips them. That sounds inconsistent/arbitrary. We should probably find another mechanism, like `predict_nonmissing!` or something using dispatch. The same approach could be used to define a fallback for `fitted` if we stored the missing mask in the formula.

But is the formula the best place to store the missing mask? That sounds like a slight abuse, since that's not an information that can be known statically.

kleinschmidt:
My philosophy is that terms should be self-contained table-to-array transformers, including `FormulaTerm`.  Whether or not things like the missing mask should be part of that transformation I'm not so clear on.  But the missing mask probably at least should be **aware** of the formula so that things like `lead`/`lag` can be supported.  Then again, you could also leave it up to the modeling package what to do with missing values in the arrays and make it part of the API that `modelcols` can return arrays with missing values.  But that seems a little undesirable to me.

nalimilan:
Right, but does that say anything about where the missingness mask should live? Terms that can generate missing values (like `lead` and `lag`) just need to mark corresponding observations as missing in the mask, and either no matrix row would be generated for them (if we can sync with other terms), or they would be dropped after the fact. In both cases, that operation can happen without storing the mask in the formula, right?

For example, `modelmatrix` could return the missingness mask. But that wouldn't be convenient for model implementations since they would have to pass it to all functions, and defining `formula` wouldn't be enough to get the nice fallback definitions to work. So I see two solutions:
- store the mask in the formula
- store it in a structure which behaves like a formula but also stores the missingness mask

I agree modeling packages shouldn't have to deal with missing values by default, especially since it can get tricky with `lead` and `lag` and because we could implement optimizations in common cases.

kleinschmidt:
And as far as making `modelmatrix` as general as possible, I think returning a tuple of arrays is the right way to go.  That's all MixedModels needs I think.  I'm not sure about the econometrics models though (@Nosferican or @matthieugomez could weigh in...)

Nosferican:
In Econometrics.jl and similar packages there are multiple intermediate models (e.g., instrumental variable models with two stages, or transformations such as features absorption). A formula also has multiple parts since there are multiple model components (response, exogenous features, endogenous features, instruments, high-dimensional features to absorb, etc.). At least in my package, I don't rely on `ModelFrame` or `ModelMatrix` and just have separate calls to `modelcols` for each parsed/split `Formula`. The models store the `Formula`s as part of the model. I would love to see tools in the panel/temporal context to make the model building less involved in each package. It is in my to-do list to work out a nice `predict(model, newdata)` which I believe it should be doable, but haven't implement it just yet... Some term names change a bit, especially for multi-response models like mlogit and I have to keep track of dropped terms (i.e., linearly deficient). As for `modelmatrix(model::RegressionModel)` I think it should return a `Matrix{<:Real}`.
It might be good to have in the `modelcols` or related,
- `eterms` I believe there were some time ago to keep track of the features involved (important for doing partial effects such as with logistic regression)
- `mask` are nice especially after `lead`/`lag`/`diff` and similar operators.
These are fine to obtain I think, `mask` by checking any missing in rows in each model term. `eterm` I 🤔 `Formula` had it, but not sure it that was only `ModelFrame`... that one would be good to keep.

nalimilan:
I've rebased and cleaned the PR. For now I propose we just merge this to get rid of `TableRegressionModel` while keeping feature parity with the current release. AFAICT, we don't have to find a solution to the issue of how to store the missingness mask for that. Later we can try to improve GLM (e.g. by allowing `predict` to insert missing values where appropriate) and StatsModels (by moving some convenience methods like `predict` and a simpler `modelmatrix` from this PR there).

src/linpred.jl
```diff
@@ -260,7 +271,40 @@ end
...
+    prediction = Tables.allocatecolumn(Union{_coltype(f.lhs), Missing}, length(nonmissings))
+    fill!(prediction, missing)
+    if interval === nothing
+        predict!(view(prediction, nonmissings), mm, newx; kwargs...)
+        return prediction
+    else
+        # Finding integer indices once is faster
+        nonmissinginds = findall(nonmissings)
+        lower = Vector{Union{Float64, Missing}}(missing, nr)
+        upper = Vector{Union{Float64, Missing}}(missing, nr)
```
kleinschmidt, line 303: what's `nr` here?

kleinschmidt:
```suggestion
        lower = Vector{Union{Float64, Missing}}(missing, length(nonmissings))
        upper = Vector{Union{Float64, Missing}}(missing, length(nonmissings))
```

?

src/lm.jl
```diff
@@ -119,11 +122,18 @@ const FIT_LM_DOC = """
     is less-than-full rank. If `true` (the default), only the first of each set of
     linearly-dependent columns is used. The coefficient for redundant linearly dependent columns is
     `0.0` and all associated statistics are set to `NaN`.
+
+    `contrasts` may optionally be supplied in the form of a `Dict` mapping term names
+    (as `Symbol`s) to term or contrast types. If a contrast is not provided
```
kleinschmidt, line 127: migth be good to give a few examples; IIRC you can't pass a contrast _type_ (like `HelmertCoding`) but need an instantiated contrasts struct (like `HelmertCoding()`).
```suggestion
    (as `Symbol`s) to term types (e.g. `ContinuousTerm`) or contrasts 
    (e.g., `HelmertCoding()`, `SeqDiffCoding(; levels=["a", "b", "c"])`, 
    etc.). If contrasts are not provided
```

src/lm.jl
```diff
@@ -233,9 +254,10 @@ function coeftable(mm::LinearModel; level::Real=0.95)
         ci = [isnan(t) ? NaN : -Inf for t in tt]
     end
     levstr = isinteger(level*100) ? string(Integer(level*100)) : string(level*100)
+    cn = mm.f === nothing ? ["x$i" for i = 1:size(mm.pp.X, 2)] : coefnames(mm)
```
kleinschmidt, line 257: seems like this might be better handled by `coefnames(::LinearModel)` etc.

src/glmfit.jl
```diff
@@ -458,6 +460,11 @@ const FIT_GLM_DOC = """
     - `minstepfac::Real=0.001`: Minimum line step fraction. Must be between 0 and 1.
     - `start::AbstractVector=nothing`: Starting values for beta. Should have the
       same length as the number of columns in the model matrix.
+    - `contrasts::AbstractDict{Symbol}=Dict{Symbol,Any}()`: a `Dict` mapping term names
+      (as `Symbol`s) to term or contrast types. If a contrast is not provided
+      for a variable, the appropriate term type will be guessed based on the data type
+      from the data column: any numeric data is assumed to be continuous, and any
+      non-numeric data is assumed to be categorical.
```
kleinschmidt, line 467: same here as below.

kleinschmidt (changes requested):
this is looking great, thanks for resurrecting it.  I do want to see the coverage to make sure that all the new fiddly bits are actually covered by the tests.  I don't see it in the checks, but from manually looking up the PR in codecov (https://app.codecov.io/gh/JuliaStats/GLM.jl/compare/339/diff) it looks like one place where I think I found a bug isn't covered (predictions with non-`nothing` intervals) so that at least needs some tests.

src/glmfit.jl
```diff
@@ -229,6 +229,7 @@ abstract type AbstractGLM <: LinPredModel end
 mutable struct GeneralizedLinearModel{G<:GlmResp,L<:LinPred} <: AbstractGLM
     rr::G
     pp::L
+    f::Union{FormulaTerm,Nothing}
```
palday, line 232: can we call this `formula`? One of the hardest parts of understanding the GLM codebase is the huge number of overly succinct field names. Or maybe just add a comment here in the struct definition:

 ```suggestion
    f::Union{FormulaTerm,Nothing} # formula used to construct the model matrices
```

File: src/linpred.jl [hunk 6]
```diff
@@ -234,13 +234,24 @@ function show(io::IO, obj::LinPredModel)
...
+    f, modelcols(f, data)
+end
+
 modelmatrix(obj::LinPredModel) = obj.pp.X
 response(obj::LinPredModel) = obj.rr.y
 
 fitted(m::LinPredModel) = m.rr.mu
 predict(mm::LinPredModel) = fitted(mm)
-StatsModels.formula(obj::LinPredModel) = modelframe(obj).formula
+formula(obj::LinPredModel) = obj.f
```
palday, line 254: Should we error on `isnothing(obj.f)` ? I suspect downstream consumers of `formula()` expect for it to return a `FormulaTerm`.

nalimilan:
Thanks for the reviews! I've pushed changes which should address all your suggestions (notably test coverage).

I've noticed that some packages rely on `TableRegressionModel` to support GLM: https://github.com/jmboehm/RegressionTables.jl/issues/128, https://github.com/yufongpeng/AnovaBase.jl/issues/52 and https://github.com/yufongpeng/AnovaGLM.jl/issues/6. Even if they adapt to support the new approach, we'd better bump version to 2.0 to avoid any breakage. That can also be the occasion to drop some long-deprecated API. We should check whether we would like to make any other breaking changes. (A few other packages use `TableRegressionModel` for their own models, it would be good that they also stop using it but there's no hurry.)

src/deprecated.jl
```diff
@@ -2,3 +2,14 @@
...
+function Base.getproperty(mm::LinPredModel, f::Symbol)
+    if f === :model
+        Base.depwarn("accessing the `model` field of GLM.jl models is deprecated, " *
+                     "as they are no longer wrapped in a `TableRegressionModel` " *
+                     "and can be used directly now", :getproperty)
+        return mm
+    else
+        return getfield(mm, f)
+    end
+end
```
palday, line 15: missing final newline.

Are there other TableRegressionModel fields we should warn about?

src/glmfit.jl
```diff
@@ -540,6 +543,13 @@ const FIT_GLM_DOC = """
     - `minstepfac::Real=0.001`: Minimum line step fraction. Must be between 0 and 1.
     - `start::AbstractVector=nothing`: Starting values for beta. Should have the
       same length as the number of columns in the model matrix.
+    - `contrasts::AbstractDict{Symbol}=Dict{Symbol,Any}()`: a `Dict` mapping term names
+      (as `Symbol`s) to term types (e.g. `ContinuousTerm`) or contrasts
+      (e.g., `HelmertCoding()`, `SeqDiffCoding(; levels=["a", "b", "c"])`,
+      etc.). If contrasts are not provided for a variable, the appropriate
+      term type will be guessed based on the data type from the data column:
+      any numeric data is assumed to be continuous, and any non-numeric data
+      is assumed to be categorical.
```
palday, line 552: should we mention that the default contrast coding for categorical data is `DummyCoding`, per StatsModels?

src/glmfit.jl
```diff
@@ -579,6 +589,30 @@ fit(::Type{M},
...
+             f::FormulaTerm,
+             data,
+             d::UnivariateDistribution,
+             l::Link=canonicallink(d);
+             offset::Union{AbstractVector, Nothing} = nothing,
+             wts::Union{AbstractVector, Nothing} = nothing,
+             dofit::Bool = true,
+             contrasts::AbstractDict{Symbol}=Dict{Symbol,Any}(),
+             fitargs...) where {M<:AbstractGLM}
+    f, (y, X) = modelframe(f, data, contrasts)
```
palday, line 602: @kleinschmidt should we do explicit `apply_schema`, etc. here instead of `modelframe` convenience? I'm thinking ahead about e.g. a hypothetical future `apply_context` stage.

src/glmfit.jl
```diff
@@ -579,6 +589,30 @@ fit(::Type{M},
...
         fit(M, float(X), float(y), d, l; kwargs...)
 
+function fit(::Type{M},
+             f::FormulaTerm,
+             data,
+             d::UnivariateDistribution,
+             l::Link=canonicallink(d);
+             offset::Union{AbstractVector, Nothing} = nothing,
+             wts::Union{AbstractVector, Nothing} = nothing,
+             dofit::Bool = true,
```
palday, line 599: if they're calling `fit` instead of a constructor, I kinda feel like we shouldn't add the complexity of `dofit` kwarg.

src/glmfit.jl
```diff
@@ -635,11 +668,54 @@ response around the linear predictor. The `:delta` method intervals are symmetri
...
+
+Store in `res` the predicted response of model `mm` from covariate values `newX`
+and, optionally, an `offset`. `res` must be a vector with a length equal to the number
+of rows in `newX` if `interval=nothing` (the default), and otherwise a `NamedTuple`
+of vectors with names `prediction`, `lower` and `upper`.
+
+$PREDICT_COMMON
+"""
+function predict!(res::Union{AbstractVector,
+                             NamedTuple{(:prediction, :lower, :upper),
```
palday, line 712: should we allow `Table.ColumnTable`? Then we have an intermediate method that takes `res::Any` and just called `Table.columntable(res)` to invoke this method making it possible to pass e.g. DataFrames

src/glmfit.jl
palday: oh nevermind, `modelframe` is defined below to do exactly that

src/linpred.jl
```diff
@@ -260,7 +275,43 @@ end
...
+        lower = Vector{Union{Float64, Missing}}(missing, length(nonmissings))
+        upper = Vector{Union{Float64, Missing}}(missing, length(nonmissings))
+        tup = (prediction=view(prediction, nonmissinginds),
+               lower=view(lower, nonmissinginds),
+               upper=view(upper, nonmissinginds))
+        predict!(tup, mm, newx;
+                 interval=interval, kwargs...)
+        return (prediction=prediction, lower=lower, upper=upper)
+    end
+end
```
palday, line 317: missing newline

src/lm.jl
```diff
@@ -119,11 +122,20 @@ const FIT_LM_DOC = """
...
     linearly-dependent columns is used. The coefficient for redundant linearly dependent columns is
     `0.0` and all associated statistics are set to `NaN`.
+
+    `contrasts` may optionally be supplied in the form of a `Dict` mapping term names
+    (as `Symbol`s) to term types (e.g. `ContinuousTerm`) or contrasts
+    (e.g., `HelmertCoding()`, `SeqDiffCoding(; levels=["a", "b", "c"])`,
+    etc.). If contrasts are not provided for a variable, the appropriate
+    term type will be guessed based on the data type from the data column:
+    any numeric data is assumed to be continuous, and any non-numeric data
+    is assumed to be categorical.
```
palday, line 132: should this be factored out to `const CONTRASTS_DESC` to make it easy to e.g. address my comment above everywhere?

src/lm.jl
```diff
@@ -140,12 +152,23 @@ function fit(::Type{LinearModel}, X::AbstractMatrix{<:Real}, y::AbstractVector{<
...
+end
+
+function fit(::Type{LinearModel}, f::FormulaTerm, data,
+             allowrankdeficient_dep::Union{Bool,Nothing}=nothing;
+             wts::Union{AbstractVector{<:Real}, Nothing}=nothing,
+             dropcollinear::Bool=true,
+             contrasts::AbstractDict{Symbol}=Dict{Symbol,Any}())
+    f, (y, X) = modelframe(f, data, contrasts)
+    wts === nothing && (wts = similar(y, 0))
+    fit!(LinearModel(LmResp(y, wts), cholpred(X, dropcollinear), f))
```
palday, line 165: no `dofit` here, good :smile:

test/runtests.jl
```diff
@@ -1440,9 +1592,49 @@ end
...
+    @test_throws ArgumentError formula(m)
+
+    df = DataFrame(x=["a", "b", "c"], y=[1, 2, 3])
+    m = lm(@formula(y ~ x), df)
+    @test formula(m)::FormulaTerm === m.formula
+
+    df = DataFrame(x=["a", "b", "c"], y=[1, 2, 3])
+    m = glm(@formula(y ~ x), df, Normal(), IdentityLink())
+    @test formula(m)::FormulaTerm === m.formula
+end
```
palday, line 1640: missing newline

test/runtests.jl
```diff
@@ -1440,9 +1592,49 @@ end
...
-        @test GLM.dispersion(mdl1.model, true) ≈ GLM.dispersion(mdl2.model,true)
+        @test GLM.dispersion(mdl1, true) ≈ GLM.dispersion(mdl2,true)
         @test confint(mdl1) ≈ confint(mdl2)
         @test aic(mdl1) ≈ aic(mdl2)
         @test predict(mdl1) ≈ predict(mdl2)
     end
 end
+
+@testset "contrasts argument" begin
+    m = lm(@formula(y~x), (y=1:25, x=repeat(1:5, 5)),
```
palday, line 1603: you should try setting non default contrasts, e.g. `EffectsCoding()` to really test this.

palday (commented):
very close, but some additional nitpicking

src/glmfit.jl
nalimilan: Yeah that doesn't make a lot of sense, but it's supported currently so I'd say we'd better keep supporting it. I've added deprecations so that we can remove them at some point.

src/glmfit.jl
nalimilan: We could do that, but TBH I don't expect people to use this method a lot (if at all). I mainly added it because it makes the design more composable and because I needed it internally. So I'd be inclined to wait and see whether people request it before making the code more complex.

test/runtests.jl
```diff
@@ -1621,6 +1622,46 @@ end
...
+             (y=1:25, x=repeat(1:5, 5)), Normal(), IdentityLink(),
+             dofit=true)
+    gm2 = fit(GeneralizedLinearModel, @formula(y~x),
+             (y=1:25, x=repeat(1:5, 5)), Normal(), IdentityLink(),
+             dofit=false)
+    @test gm1.fit
+    @test !gm2.fit
+    fit!(gm2)
+    @test gm2.fit
+    @test coef(gm1) == coef(gm2) ≈ [10, 1]
```
palday, line 1664: did you mean to do the same thing twice?

palday (approved):

src/glmfit.jl
kleinschmidt: I came here to say the same thing before I looked further down 😂

test/runtests.jl
```diff
@@ -847,6 +847,28 @@ end
     @test ndims(gm11_pred3.upper) == 1
     @test ndims(gm11_pred3.lower) == 1
 
+    @test predict!(similar(Y, size(newX, 1)), gm11, newX) == predict(gm11, newX)
+    @test predict!((prediction=similar(Y, size(newX, 1)),
+                    lower=similar(Y, size(newX, 1)),
+                    upper=similar(Y, size(newX, 1))),
+                   gm11, newX, interval=:confidence, interval_method=:delta) ==
+        predict(gm11, newX, interval=:confidence, interval_method=:delta)
+   @test predict!((prediction=similar(Y, size(newX, 1)),
```
kleinschmidt, line 856:
```suggestion
    @test predict!((prediction=similar(Y, size(newX, 1)),
```

kleinschmidt (commented):
i think in general this is great, thanks for doing all this!  my only quibble is that there are a few places where inconsistencies with the old behavior might sneak in:
- printing does not print the formula anymore
- `apply_schema` previously woudl get teh ACTUAL type of model being fit; now it just get `LinPredModel` which may cause issues if someone has defined some special overloads for a particular subtype.  instead, I'd suggest passing the type parameer `M` to `modelframe` as a fourth argument.

kleinschmidt (approved):

test/runtests.jl
nalimilan: Yes, the first one checks `glm`, the second `fit`. :-)

nalimilan:
Good catch. While we're at it, I propose that we no longer print type parameters, which are not super useful and obscure the interesting part of the output. See last commit. If we support switching between QR and Cholesky in the future, we may print only the name of the decomposition without all the parameters.

I've also changed `model_frame` to take the model type as you suggest.

bkamins:
> I propose that we no longer print type parameters

For standard display I think we should print as little as possible. In my lectures I always need to tell my students "ignore this type info as it is not relevant to you".

src/deprecated.jl
nalimilan: I've added a branch to allow accessing the formula. Other fields are not very likely to be used, except maybe the data but we probably don't want to store it.

Issue #258, opened 2018-09-27
`lm` doesn't support wts


pdeffebach:
```
julia> using DataFrames, GLM

julia> df = DataFrame(y = randn(100), x = randn(100), w = rand(100) .+ 1);

julia> m = lm(@formula(y ~ x), df)
StatsModels.DataFrameRegressionModel{LinearModel{LmResp{Array{Float64,1}},DensePredChol{Float64,LinearAlgebra.Cholesky{Float64,Array{Float64,2}}}},Array{Float64,2}}

Formula: y ~ 1 + x

Coefficients:
               Estimate Std.Error   t value Pr(>|t|)
(Intercept)   -0.108285 0.0909041  -1.19119   0.2365
x            -0.0154871 0.0855186 -0.181096   0.8567


julia> m = lm(@formula(y ~ x), df, wts = df.w)
ERROR: function lm does not accept keyword arguments
Stacktrace:
 [1] kwfunc(::Any) at .\boot.jl:321
 [2] top-level scope at none:0
```

Is there any way to add weights to `lm`. Or is the recommendation to use `glm` for any more complex regression? 

I would have expected `lm(f, df; kwargs...)` to simply be a wrapper for `glm(f, df, Normal(), IdentityLink(); kwargs...)` but that isn't the case. Is there a reason for this?

nalimilan:
> I would have expected `lm(f, df; kwargs...)` to simply be a wrapper for `glm(f, df, Normal(), IdentityLink(); kwargs...)` but that isn't the case. Is there a reason for this?

Yes, linear regression can be fitted more efficiently than GLMs.

I guess nobody has added support for weights, but it should be relatively easy to do for frequency weights.

pdeffebach:
It looks like all of the model response objects allow for weights. But it's a bit beyond me at this point to make a PR. 

Bump for weights to be in the model object as well though.

nalimilan:
Indeed. Maybe all that's needed is to fix `fit` to accept weights an pass them to `LmResp` then.

PR #341, opened 2019-11-01
Add wts keyword to `lm`

pdeffebach:
Addresses #258. It definitely doesn't work right now, as I haven't been able to track down all the up-stream functions that need to use the weights of a `LmResp` object.


pdeffebach:
I spent more time on this tonight, trying to make `lm` seem more like `glm`. I realized that the `GlmResp` holds a lot of things that are necessary for the weights to work. I think we should just dispatch to `glm(..., Normal(), wts = wts)` when the user uses `wts` with `lm` rather than alter the `LmResp` object and algorithm to use weights. 

Does that sounds good?

nalimilan:
I'd rather keep `lm` always returning a `LinearModel`. Otherwise people who dispatch on it will get in trouble when using weights, `ftest` won't work, p-values won't be based on the t distribution, etc.

pdeffebach:
Thanks for the feedback. That's a valid concern. 

If you have time, could you verify that we would need to add more fields to the `LmResp` object to accommodate actually running OLS with weights? At the moment it's hard to imagine enabling `LinearModel` to work with weights without basically making it look exactly like `GeneralizedLinearModel`.

nalimilan:
Well `LmResp` already has a `wts` field, and other ones seem irrelevant (linked to the IRLS estimation algorithm). What would you need to add?

pdeffebach:
After a decent amount of time inspecting functions to figure out what to change, I think I made the necessary changes. 

```
julia> df = DataFrame(y = rand(10), x = rand(10), w = float.(rand(Bool, 10)))
julia> lm(@formula(y ~ x), df, wts = df.w)
StatsModels.TableRegressionModel{LinearModel{GLM.LmResp{Array{Float64,1}},GLM.DensePredChol{Float64,LinearAlgebra.Cholesky{Float64,Array{Float64,2}}}},Array{Float64,2}}

y ~ 1 + x

Coefficients:
────────────────────────────────────────────────────────────────────────────
              Estimate  Std. Error   t value  Pr(>|t|)  Lower 95%  Upper 95%
────────────────────────────────────────────────────────────────────────────
(Intercept)   0.870868    0.447738   1.94504    0.1912   -1.05559    2.79733
x            -0.493512    0.902051  -0.5471     0.6392   -4.37472    3.3877 
────────────────────────────────────────────────────────────────────────────
julia> glm(@formula(y ~ x), df, Normal(), wts = df.w)
StatsModels.TableRegressionModel{GeneralizedLinearModel{GLM.GlmResp{Array{Float64,1},Normal{Float64},IdentityLink},GLM.DensePredChol{Float64,LinearAlgebra.Cholesky{Float64,Array{Float64,2}}}},Array{Float64,2}}

y ~ 1 + x

Coefficients:
──────────────────────────────────────────────────────────────────────────────
              Estimate  Std. Error   z value  Pr(>|z|)    Lower 95%  Upper 95%
──────────────────────────────────────────────────────────────────────────────
(Intercept)   0.870868    0.447738   1.94504    0.0518  -0.00668183    1.74842
x            -0.493512    0.902051  -0.5471     0.5843  -2.2615        1.27448
──────────────────────────────────────────────────────────────────────────────
```

Actually it looks like the confidence intervals are all off, which will require some tweaks. @nalimilan can you let me know if I'm heading in the right direction in terms of the way I changed the `updatemu!` method?

Manifest.toml
```diff
@@ -0,0 +1,232 @@
+# This file is machine-generated - editing it directly is not advised
```
nalimilan: Remove this file.

src/lm.jl
```diff
@@ -27,11 +27,12 @@ mutable struct LmResp{V<:FPVector} <: ModResp  # response in a linear model
         new{V}(mu, off, wts, y)
     end
 end
-LmResp(y::V) where {V<:FPVector} = LmResp{V}(fill!(similar(y), 0), similar(y, 0), similar(y, 0), y)
 
-LmResp(y::AbstractVector{T}) where T<:Real = LmResp(float(y))
+LmResp(y::FPVector; wts::FPVector = similar(y, 0)) = LmResp{typeof(y)}(fill!(similar(y), 0), similar(y, 0), wts, y)
```
nalimilan: Make `wts` a positional argument for consistency with `GlmResp`.

Also, no spaces around `=` for arguments, and break lines at 92 chars.

src/lm.jl
```diff
@@ -27,11 +27,12 @@ mutable struct LmResp{V<:FPVector} <: ModResp  # response in a linear model
...
 
-function updateμ!(r::LmResp{V}, linPr::V) where V<:FPVector
+LmResp(y::AbstractVector{T}; wts::AbstractVector{T} = similar(y, 0)) where T<:Real = LmResp(float(y), float(wts))
```
nalimilan:
```suggestion
LmResp(y::AbstractVector{<:Real}; wts::AbstractVector{<:Real}=similar(y, 0)) = LmResp(float(y), float(wts))
```

src/lm.jl
```diff
@@ -126,24 +127,30 @@ end
 LinearAlgebra.cholesky(x::LinearModel) = cholesky(x.pp)
 
 function StatsBase.fit!(obj::LinearModel)
-    installbeta!(delbeta!(obj.pp, obj.rr.y))
+    if length(obj.rr.wts) == 0
+        installbeta!(delbeta!(obj.pp, obj.rr.y))
```
nalimilan: `installbeta!(obj.pp)` can be moved outside of the `if` AFAICT.

src/lm.jl
```diff
@@ -126,24 +127,30 @@ end
...
+    end          
     updateμ!(obj.rr, linpred(obj.pp, zero(eltype(obj.rr.y))))
     return obj
 end
 
-function fit(::Type{LinearModel}, X::AbstractMatrix, y::AbstractVector,
-             allowrankdeficient::Bool=false)
-    fit!(LinearModel(LmResp(y), cholpred(X, allowrankdeficient)))
+function fit(::Type{LinearModel}, X::AbstractMatrix, y::V,
+             allowrankdeficient::Bool=false; wts::V = similar(y, 0)) where {V<:FPVector}
```
nalimilan: Why make this signature more restrictive?

src/lm.jl
```diff
@@ -126,24 +127,30 @@ end
...
+    fit!(LinearModel(LmResp(y, wts = wts), cholpred(X, allowrankdeficient)))
 end
 
 """
-    lm(X, y, allowrankdeficient::Bool=false)
+    lm(X, y, allowrankdeficient::Bool=false; wts = nothing)
```
nalimilan:
```suggestion
    lm(X, y, allowrankdeficient::Bool=false; wts=similar(y, 0))
```

src/lm.jl
pdeffebach: I didn't want to make `wts` a positional argument because there are 4 fields in `LmResp` and `wts` is like the 4th argument. So it seemed weird to have the arguments taken be different from the fields of the object.

nalimilan: Maybe but then `GlmResp` should be changed too... Not sure it's worth it.

nalimilan:
Do you think this is ready? If so, could you add a test?

pdeffebach:
No it's not ready yet. I still have to figure out why the confidence intervals don't match, as well as stuff like `nobs`, which I think is in StatsBase.

nalimilan:
OK. FWIW, as I noted, I don't think confidence intervals should match with `glm`.

pdeffebach:
Oh I see your point now. `lm` is more robust than `glm` with `Normal()` in that sense. I'll double check with R and Stata.

nalimilan:
You don't even need other software: just fit the same model without weights but with replicated rows. R doesn't actually support frequency weights so I don't think you can reproduce the p-values. With Stata you can with `fweights`.

pdeffebach:
This PR is ready for merging, with a caveat. 

Currently issue #84 requires `lm` to support `y` and `X` having different types, while GLM doesn't, so it makes sense to allow the `wts` keyword argument to be `AbstractVector{<:Real}` rather than the same type as `y`. Also note that canonically frequency weights are integers, so there are statistical reasons for `wts` to be flexible. 

Currently the constructors for `GLMResp` in the `glm` function are much stricter, `wts` must be the same type as `y`. So in my tests I have weights be a vector of `Float`s, but the reality is that a vector of integers works for `wts` for `lm` but not for `glm`. 

I propse merging this and then in a separate PR adding a new constructor method for `GLMResp` which allows for more type flexibility.

src/lm.jl
```diff
@@ -27,11 +27,16 @@ mutable struct LmResp{V<:FPVector} <: ModResp  # response in a linear model
         new{V}(mu, off, wts, y)
     end
 end
-LmResp(y::V) where {V<:FPVector} = LmResp{V}(fill!(similar(y), 0), similar(y, 0), similar(y, 0), y)
 
-LmResp(y::AbstractVector{T}) where T<:Real = LmResp(float(y))
+function LmResp(y::FPVector, wts::FPVector=similar(y, 0))
```
nalimilan: Why not keep short function syntax? Same below.

src/lm.jl
```diff
@@ -27,11 +27,16 @@ mutable struct LmResp{V<:FPVector} <: ModResp  # response in a linear model
...
+function LmResp(y::FPVector, wts::FPVector=similar(y, 0)) 
+    LmResp{typeof(y)}(fill!(similar(y), 0), similar(y, 0), wts, y)
+end
+
+function LmResp(y::AbstractVector{<:Real}, wts::AbstractVector{<:Real}=similar(y, 0))
+    LmResp(float(y), float(wts))
+end
 
-function updateμ!(r::LmResp{V}, linPr::V) where V<:FPVector
+function updateμ!(r::LmResp{V}, linPr::V) where {V<:FPVector}
```
nalimilan: This isn't needed.

src/lm.jl
```diff
@@ -126,24 +131,31 @@ end
...
+    installbeta!(obj.pp)     
     updateμ!(obj.rr, linpred(obj.pp, zero(eltype(obj.rr.y))))
     return obj
 end
 
-function fit(::Type{LinearModel}, X::AbstractMatrix, y::AbstractVector,
-             allowrankdeficient::Bool=false)
-    fit!(LinearModel(LmResp(y), cholpred(X, allowrankdeficient)))
+function fit(::Type{LinearModel}, X::AbstractMatrix{<:Real}, y::AbstractVector{<:Real},
+             allowrankdeficient::Bool=false; wts::AbstractVector=similar(y, 0))
```
nalimilan:
```suggestion
             allowrankdeficient::Bool=false; wts::AbstractVector{<:Real}=similar(y, 0))
```

src/lm.jl
```diff
@@ -126,24 +131,31 @@ end
...
 
 """
-    lm(X, y, allowrankdeficient::Bool=false)
+    lm(X, y, allowrankdeficient::Bool=false; wts=similar(y, 0))
 
 An alias for `fit(LinearModel, X, y, allowrankdeficient)`
 
 The arguments `X` and `y` can be a `Matrix` and a `Vector` or a `Formula` and a `DataFrame`.
+
+The keyword argument `wts` can be a `Vector{Float64}`.
```
nalimilan:
```suggestion
The keyword argument `wts` can be a `Vector` specifying frequency weights for observations.
```

test/runtests.jl
```diff
@@ -46,6 +46,26 @@ linreg(x::AbstractVecOrMat, y::AbstractVector) = qr!(simplemm(x)) \ y
     @test isapprox(coef(lm1), coef(lm2) .* [1., 10.])
 end
 
+@testset "lm with weights" begin 
+    df = dataset("quantreg", "engel")
+    N = nrow(df)
+    df.weights = repeat(1.0:5.0, Int(N/5))
```
nalimilan: Probably not worth repeating this test every time: now that you know you got the correct values, just hardcode them in the test (just like we do above).

Also test other statistics like p-values, confidence intervals, R², deviance, log-likelihood, etc. (see examples in the file).

nalimilan:
> I propse merging this and then in a separate PR adding a new constructor method for `GLMResp` which allows for more type flexibility.

Sure.

By the way, we could remove `wts` to `weights`, which is a much more explicit name. It would be nice to document clearly that these are frequency weights (in docstrings and in the manual).

pdeffebach:
Thanks for requiring the R^2 be a test, there is a bug. We should be using the weighed mean to calculated the `nulldeviance` of the model, that is, instead of the sum of deviations from the mean, we should use the sum of deviations from the un-weighted mean. 

I don't have Stata open to check that this is what fweights does, but this is what `lfe` does in R. The following code shows that the R^2 in `lfe` with weights is equivelent to the R^2 of the "long" dataset. 
```
library("quantreg")
library("tidyverse")
library("lfe")
data(engel)
df = as_tibble(engel)
df$weights = rep(1:5, nrow(df) / 5)
felm(foodexp ~ income, data = df, weights = df$weights)
r2 = summary(felm_model)$r.squared # get r-squared
print(r2)
```

pdeffebach:
As an addendum to the above post: I realized that `lfe` doesn't document the way weights are used when calculating the variance-covariance matrix. So I broke out stata on my other computer and confirmed that Stata also uses the weighted mean when calculating the R^2. 

I will go through and just use stata to ensure everything is correct rather than mess with `felm`.

pdeffebach:
@nalimilan I would like you to please confirm that the calculation of the loglikelihood is correct for `lm` with weights. 

Stata assumes normality of errors when calculating the log-likelihood. This means that 

```
reg foodexp income [fweight = weight]
```

gives you an `e(ll) = -1445`, which matches the `loglikelihood` from the Julia model 

```
glm(@formula(foodexp ~ income), wts = df.weight)
```

However the `loglikelihood` of 

```
lm(@formula(foodexp ~ income), wts = df.weight)
```

is different, due to the implementation [here](https://github.com/JuliaStats/GLM.jl/blob/abadefaf18a2b5373cd858880358c362771b7693/src/lm.jl#L76). If you could expand on why it's written like that so we know it's right I'd appreciate it.

nalimilan:
Thanks for checking against Stata.

I don't remember the details of why I implemented `loglikelihood` that way for linear models, and that method doesn't seem to be covered by tests with weights, so maybe it's wrong. I'm not sure why the sum of weights appears in the formula separately from `n`: actually I would just replace `n` with the sum of weights, or just `n = nobs(m)` which does the right thing whether there are weights or not. (Maybe I incorrectly took the formula for analytic weights...)

If you can't find the solution, you could also use the same approach as for GLMs, which should be a bit less efficient, but we don't care that much.

pdeffebach:
I didn't end up figuring out why the log-likelihood was implemented the way it was. It didn't match `aweights`. I implemented the `glm` version for `lm` and made sure it matched stata. 

This is a weird PR because there exists a lot of code for `lm` with weights that has never been used since `lm` has never supported weights until now. For some reason the `residuals` function for `lm` multiplied each residual (i.e. `x_i' * beta - y`) by the weight, which is wrong and I have now fixed.

src/lm.jl
```diff
@@ -74,38 +81,50 @@ function nulldeviance(r::LmResp)
...
 
 function nullloglikelihood(r::LmResp)
-    n = length(r.y)
     wts = r.wts
-    sw = zero(log(one(eltype(wts))))
-    for w in wts
-        sw += log(w)
+    y   = r.y
+    if isempty(wts)
+        @show mu  = mean(y)
```
nalimilan:
```suggestion
        mu = mean(y)
```

src/lm.jl
```diff
@@ -74,38 +81,50 @@ function nulldeviance(r::LmResp)
...
+        ϕ = nulldeviance(r)/length(y)
+        @inbounds for i in eachindex(y)
+            ll += loglik_obs(Normal(), y[i], mu, 1, ϕ)
         end
-        resid
     end
+    ll
+end
+
+function residuals(r::LmResp)
```
nalimilan: Make this a one-liner?

src/lm.jl
```diff
@@ -126,24 +145,31 @@ end
 LinearAlgebra.cholesky(x::LinearModel) = cholesky(x.pp)
 
 function StatsBase.fit!(obj::LinearModel)
-    installbeta!(delbeta!(obj.pp, obj.rr.y))
+    if length(obj.rr.wts) == 0
```
nalimilan:
```suggestion
    if isempty(obj.rr.wts)
```

src/lm.jl
```diff
@@ -126,24 +145,31 @@ end
...
+    lm(X, y, allowrankdeficient::Bool=false; wts=similar(y, 0))
 
 An alias for `fit(LinearModel, X, y, allowrankdeficient)`
 
 The arguments `X` and `y` can be a `Matrix` and a `Vector` or a `Formula` and a `DataFrame`.
+
+The keyword argument `wts` can be a `Vector` specifying frequency weights for observations.
 """
-lm(X, y, allowrankdeficient::Bool=false) = fit(LinearModel, X, y, allowrankdeficient)
+lm(X, y, allowrankdeficient::Bool=false; kwargs...) = fit(LinearModel, X, y, allowrankdeficient; kwargs...)
```
nalimilan:
```suggestion
lm(X, y, allowrankdeficient::Bool=false; kwargs...) =
    fit(LinearModel, X, y, allowrankdeficient; kwargs...)
```

test/runtests.jl
```diff
@@ -46,6 +46,26 @@ linreg(x::AbstractVecOrMat, y::AbstractVector) = qr!(simplemm(x)) \ y
     @test isapprox(coef(lm1), coef(lm2) .* [1., 10.])
 end
 
+@testset "weights" begin
```
nalimilan:
```suggestion
@testset "linear model with weights" begin 
```

test/runtests.jl
```diff
@@ -46,6 +46,26 @@ linreg(x::AbstractVecOrMat, y::AbstractVector) = qr!(simplemm(x)) \ y
...
+    df = dataset("quantreg", "engel")
+    N = nrow(df)
+    df.weights = repeat(1:5, Int(N/5))
+    f = @formula(FoodExp ~ Income)
+    lm_model = lm(f, df, wts = df.weights)
+    glm_model = glm(f, df, Normal(), wts = df.weights)
+    @test isapprox(coef(lm_model), [154.35104595140706, 0.4836896390157505])
+    @test isapprox(coef(glm_model), [154.35104595140706, 0.4836896390157505])
+    @test isapprox(stderror(lm_model), [9.382302620120193, 0.00816741377772968])
+    @test isapprox(r2(lm_model), 0.8330258148644486)
```
nalimilan: Could also test pseudo-R² of LM and GLM.

src/lm.jl
```diff
@@ -74,38 +81,50 @@ function nulldeviance(r::LmResp)
 end
 
 function loglikelihood(r::LmResp)
```
nalimilan: Actually I've found the correct definition: it's just
```julia
function loglikelihood(m::LinearModel)
    n = nobs(m)
    -n/2 * (log(2π * deviance(m)/n) + 1)
end

function nullloglikelihood(m::LinearModel)
    n = nobs(m)
    -n/2 * (log(2π * nulldeviance(m)/n) + 1)
end
```

nalimilan:
Indeed, multiplying the residual for each observation by its weight doesn't make sense.

src/lm.jl
pdeffebach: thanks for finding this. It works (when `n` is the sum of weights).

pdeffebach:
I can't find any reference to Pseudo-R^2. Is that something that lives in GLM? Not sure how to test it.

nalimilan:
See https://juliastats.org/StatsBase.jl/latest/statmodels/#StatsBase.r2.

pdeffebach:
`r2` doesn't work for GLM, unfortunately. It says `mss not defined for Generalized...`. It should be a one-liner to implement though the way it is for `lm`.

nalimilan:
I mean `r²(m, :McFadden)`.

getzze:
The definition of loglikelihood (and nullloglikelihood) seems incorrect to me. As defined in StatsBase (https://juliastats.org/StatsBase.jl/latest/statmodels/#StatsBase.deviance), the deviance is `2*(l_S - l)`, where `l` is the loglikelihood of the model and `l_S` the loglikelihood of the saturated model, that is often 0. So I would expect that the loglikelihood is proportional to the deviance.

Also for GLMs, the definition of the loglikelihood is a bit confusing to me, I don't get why the deviance is used as an estimate of the second parameter of the distribution. Is the deviance supposed to be an estimate of the variance?

PR #352, opened 2020-01-06
null deviance and null loglikelihood for GLM

### LINKED ISSUE #256: nulldeviance function
It is not currently defined in GLM. Can this be added? It is required for pseudo R2 for GLM models.

pdeffebach:
This is a commit to fix #256, noting that GLM is without a null deviance or null log likelihood function. It's important to have this because likelihood ratio tests and R^2 use null deviance. 

I am not 100% if this is correct. So far my results match R's `glm` with `gaussian()` and `binomial()` families, but to be honest I am still taking econometrics courses so my knowledge of this is slim. 

I will add `nullloglikelihood` next, then tests for everything I can, then add `r2` functions and anything else that might use them.

### DIFF
--- a/src/glmfit.jl
+++ b/src/glmfit.jl
@@ -65,6 +65,26 @@ end
 
 deviance(r::GlmResp) = sum(r.devresid)
 
+function nulldeviance(r::GlmResp)
+    y = r.y
+    d = r.d
+    wts = r.wts
+    if isempty(r.wts)
+        mu = mean(y)
+        dv = 0.0 
+        for i in eachindex(y)
+            dv += devresid(d, y[i], mu)
+        end
+    else 
+        mu = mean(y, weights(wts))
+        dv = 0.0
+        for i in eachindex(y, wts)
+            dv += devresid(d, y[i], mu) * wts[i]
+        end
+    end
+    return dv
+end
+
 """
     cancancel(r::GlmResp{V,D,L})
 
@@ -243,6 +263,7 @@ function confint(obj::AbstractGLM; level::Real=0.95)
 end
 
 deviance(m::AbstractGLM) = deviance(m.rr)
+nulldeviance(m::AbstractGLM) = nulldeviance(m.rr)
 
 function loglikelihood(m::AbstractGLM)
     r   = m.rr
--- a/src/glmfit.jl
+++ b/src/glmfit.jl
@@ -71,13 +71,13 @@ function nulldeviance(r::GlmResp)
     wts = r.wts
     if isempty(r.wts)
         mu = mean(y)
-        dv = 0.0 
+        dv = zero(mu) 
         for i in eachindex(y)
             dv += devresid(d, y[i], mu)
         end
     else 
         mu = mean(y, weights(wts))
-        dv = 0.0
+        dv = zero(mu)
         for i in eachindex(y, wts)
             dv += devresid(d, y[i], mu) * wts[i]
         end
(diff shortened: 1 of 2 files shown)


pdeffebach:
I realized I forgot about this. I apologize for that. I will be picking this up slowly, starting with a rebase and then adding more checks against R as I continue tests.

pdeffebach:
Bumping this for @andreasnoack and @nalimilan. I forgot about this PR for a long time but am determined to finish it. 

It just adds a `nulldeviance` method, previously this was defined for `lm` but not for `glm`. 

I think I need some help getting this over the finish line. As you can see in the `test` file, I don't know how to add tests for models with `offset` and for negative binomial models. 

The original intention of this PR was to add `nullloglikelihood` as well. Though I don't quite know the way forward on that, hopefully it is easy. My `RCall` testing setup has been very useful, but also a lot of work to set up. I would rather not merge this PR, deleting all the RCall stuff and have to re-write it for `nullloglikelihood` tests.

nalimilan:
Thanks. Looks good. For offsets, I think you can just add `+ offset(x)` to the formula in R (there's an example in ?glm) to check whether results are consistent.

For the negative binomial, I think there are two different cases:
- If theta is fixed (as in two testsets), you can just fit the null model in R with the same value using `glm(formula, family = MASS::negative.binomial(2), ...)`. That should give the same results as in Julia.
- If theta is not fixed (as in the two testsets calling `negbin`), the null deviance will correspond to the null model but with the same theta value as the considered model, which is not equivalent to fitting the null model separately. You can check that the null log-likelihood is correct by fitting the model with theta fixed to the value that `negbin` estimated.

Regarding `nulloglikelihood`, I think you can just take the definition for `loglikelihood` and adjust it to use the mean as a prediction, as you did for `nulldeviance`.

kleinschmidt:
One possibility is extending the use of `hasintercept` which right now just applies to terms

https://github.com/JuliaStats/StatsModels.jl/blob/04a0ccfb80f6ef7fe4a618556e9d3d21d1c445d1/src/terms.jl#L562-L565

The default would check the formula stored with a model and packages can implement their own specializations as needed...

nalimilan:
We're in GLM here so we don't have to follow any public interface. :-)

kleinschmidt:
Well, at the moment in order to support GLM models as they are usual fit by folks you'd have to define a `nulldeviance(::TableRegressionModel)` method to unwrap the model, which is pretty extreme type piracy...although now that I'm thinking about it I don't see how my solution solves that problem...and this PR isn't changing anything there.

The advantage of using `hasintercept` is that it's potentially more efficient (once we require that models store the formula) than checking al the values in the model matrix.  But that's probably a pretty minor gain.

pdeffebach:
> But we allow fitting models from raw matrices, without going through a formula/TableRegressionModel, so it would be better to detect whether the model matrix has a constant. 

Following discussion on Slack, I will write a little function that checks if the matrix has all constant for some value. This should be fast since if it was constructed via a `formula` then the first column will be all ones. 

After this I will add a few tests of models without intercepts. 

I am going to abandon my effort to get `offsets` to work. If someone is using offsets and sees `error: nulldeviance with offsets not yet implemented` then hopefully they know more than I do and can implement them. 

I also still have to work on negative binomial models.

nalimilan:
I ended up reimplementing this at https://github.com/JuliaStats/GLM.jl/pull/479. The handling of offsets is indeed tricky. Closing in favor the the other PR.

PR #355, opened 2020-01-15
Fix passing tolerance to mle_for_θ

nalimilan:
Old PR #302 used `tol`, which no longer exists since #312.


nalimilan:
Merging to fix CI on master.

PR #364, opened 2020-03-19
Add confidence intervals for GLM predictions

### LINKED ISSUE #101: Interval functionality in predict() function?

nignatiadis:
As per the title. I tried to follow the interface of the `predict` function for linear models. Note that for GLMs `:prediction` intervals do not make sense in general; `:confidence` intervals however do.

R's `predict.glm` does not provide intervals, but it can calculate the standard error of the predictions through the `se.fit` option. The intervals in this PR are calculated based on a Normal approximation and the standard error of the predictions is calculated as in R (through the Delta method).

Fixes #101.

### DIFF
--- a/src/glmfit.jl
+++ b/src/glmfit.jl
@@ -533,7 +533,8 @@ function predict(mm::AbstractGLM, newX::AbstractMatrix;
         # Use Delta method to estimate variance in two steps
         # 1. Estimate variance for eta based on variance for coefficients
         #    through the diagonal of newX*vcov(mm)*newX'
-        vareta =  dot.(eachrow(newX), eachcol(vcov(mm)*newX'))
+        vcovXnewT = vcov(mm)*newX'
+        vareta = [dot(view(newX, i, :), view(vcovXnewT, :, i)) for i in axes(newX,1)]
         # 2. Now compute the variance for mu based on variance of eta
         varmu = vareta .* [abs2(mueta(Link(mm), x)) for x in eta]
     else
--- a/src/glmfit.jl
+++ b/src/glmfit.jl
@@ -534,32 +534,30 @@ function predict(mm::AbstractGLM, newX::AbstractMatrix;
     if interval === nothing
         return mu
     elseif interval == :confidence
-        normalquantile = quantile(Normal(), (1. + level)/2)
+        normalquantile = quantile(Normal(), (1 + level)/2)
         # Use Delta method to estimate variance in two steps
         # (2nd step varies as `symmetric` is `true` or `false`)
         # 1. Estimate variance for eta based on variance for coefficients
         #    through the diagonal of newX*vcov(mm)*newX'
         vcovXnewT = vcov(mm)*newX'
-        vareta = [dot(view(newX, i, :), view(vcovXnewT, :, i)) for i in axes(newX,1)]
+        stdeta = [sqrt(dot(view(newX, i, :), view(vcovXnewT, :, i))) for i in axes(newX,1)]
         
         if symmetric
             # 2. Now compute the variance for mu based on variance of eta and
             # construct intervals based on that
-            varmu = vareta .* [abs2(mueta(Link(mm), x)) for x in eta]
-            lower = mu .- normalquantile .* sqrt.(varmu)
-            upper = mu .+ normalquantile .* sqrt.(varmu)
+            stdmu = stdeta .* abs.(mueta.(Link(mm), eta))
+            lower = mu .- normalquantile .* stdmu
+            upper = mu .+ normalquantile .* stdmu
         elseif !symmetric 
             # 2. Construct intervals for eta, then apply inverse link
-            lower = linkinv.(Link(mm), eta .- normalquantile .* sqrt.(vareta))
-            upper = linkinv.(Link(mm), eta .+ normalquantile .* sqrt.(vareta))
+            lower = linkinv.(Link(mm), eta .- normalquantile .* stdeta)
+            upper = linkinv.(Link(mm), eta .+ normalquantile .* stdeta)
         end
     else
         error("only :confidence intervals are defined")
     end
     (prediction = mu, lower = lower, upper = upper)
 end
-
-
 
 # A helper function to choose default values for eta
 function initialeta!(eta::AbstractVector,
--- a/src/glmfit.jl
+++ b/src/glmfit.jl
@@ -507,13 +507,16 @@ function dispersion(m::AbstractGLM, sqr::Bool=false)
 end
 
 """
-    predict(mm::AbstractGLM, newX::AbstractMatrix; offset::FPVector=Vector{eltype(newX)}(0))
+    predict(mm::AbstractGLM, newX::AbstractMatrix; offset::FPVector=eltype(newX)[],
+            interval::Union{Symbol,Nothing}=nothing, level::Real = 0.95)
 
-Form the predicted response of model `mm` from covariate values `newX` and, optionally,
-an offset.
+If `interval` is `nothing` (the default), return the predicted response of model
+`mm` from covariate values `newX` and, optionally, an `offset`. If `interval` is
+`:confidence`, also return upper and lower bounds for a given coverage `level`.
 """
 function predict(mm::AbstractGLM, newX::AbstractMatrix;
-                 offset::FPVector=eltype(newX)[])
+                 offset::FPVector=eltype(newX)[],
+                 interval::Union{Symbol,Nothing}=nothing, level::Real=0.95)
     eta = newX * coef(mm)
     if !isempty(mm.rr.offset)
         length(offset) == size(newX, 1) ||
@@ -523,7 +526,24 @@ function predict(mm::AbstractGLM, newX::AbstractMatrix;
         length(offset) > 0 && throw(ArgumentError("fit without offset, so value of `offset` kw arg does not make sense"))
     end
     mu = [linkinv(Link(mm), x) for x in eta]
+
+    if interval === nothing
+        return mu
+    elseif interval == :confidence
+        # Use Delta method to estimate variance in two steps
+        # 1. Estimate variance for eta based on variance for coefficients
+        #    through the diagonal of newX*vcov(mm)*newX'
+        vareta =  dot.(eachrow(newX), eachcol(vcov(mm)*newX'))
+        # 2. Now compute the variance for mu based on variance of eta
+        varmu = vareta .* [abs2(mueta(Link(mm), x)) for x in eta]
+    else
+        error("only :confidence intervals are defined")
+    end
+    ciwidth = quantile(Normal(), (1. - level)/2) .* sqrt.(varmu)
+    (prediction = mu, lower = mu .+ ciwidth, upper = mu .- ciwidth)
 end
+
+
 
 # A helper function to choose default values for eta
 function initialeta!(eta::AbstractVector,
--- a/src/glmfit.jl
+++ b/src/glmfit.jl
@@ -508,15 +508,19 @@ end
 
 """
     predict(mm::AbstractGLM, newX::AbstractMatrix; offset::FPVector=eltype(newX)[],
-            interval::Union{Symbol,Nothing}=nothing, level::Real = 0.95)
+            interval::Union{Symbol,Nothing}=nothing, level::Real = 0.95,
+            symmetric::Bool = false)
 
 If `interval` is `nothing` (the default), return the predicted response of model
 `mm` from covariate values `newX` and, optionally, an `offset`. If `interval` is
-`:confidence`, also return upper and lower bounds for a given coverage `level`.
+`:confidence`, also return upper and lower bounds for a given coverage `level`. By 
+default (`symmetric = false`) the intervals are constructed by applying the inverse link
+to intervals for the linear predictor, if `symmetric=true`, the intervals are symmetric 
+around the point estimates.
 """
 function predict(mm::AbstractGLM, newX::AbstractMatrix;
                  offset::FPVector=eltype(newX)[],
-                 interval::Union{Symbol,Nothing}=nothing, level::Real=0.95)
+                 interval::Union{Symbol,Nothing}=nothing, level::Real=0.95, symmetric=false)
     eta = newX * coef(mm)
     if !isempty(mm.rr.offset)
         length(offset) == size(newX, 1) ||
@@ -525,23 +529,34 @@ function predict(mm::AbstractGLM, newX::AbstractMatrix;
     else
         length(offset) > 0 && throw(ArgumentError("fit without offset, so value of `offset` kw arg does not make sense"))
     end
-    mu = [linkinv(Link(mm), x) for x in eta]
+    mu = linkinv.(Link(mm), eta)
 
     if interval === nothing
         return mu
     elseif interval == :confidence
+        normalquantile = quantile(Normal(), (1. + level)/2)
         # Use Delta method to estimate variance in two steps
+        # (2nd step varies as `symmetric` is `true` or `false`)
         # 1. Estimate variance for eta based on variance for coefficients
         #    through the diagonal of newX*vcov(mm)*newX'
         vcovXnewT = vcov(mm)*newX'
         vareta = [dot(view(newX, i, :), view(vcovXnewT, :, i)) for i in axes(newX,1)]
-        # 2. Now compute the variance for mu based on variance of eta
-        varmu = vareta .* [abs2(mueta(Link(mm), x)) for x in eta]
+        
+        if symmetric
+            # 2. Now compute the variance for mu based on variance of eta and
+            # construct intervals based on that
+            varmu = vareta .* [abs2(mueta(Link(mm), x)) for x in eta]
+            lower = mu .- normalquantile .* sqrt.(varmu)
+            upper = mu .+ normalquantile .* sqrt.(varmu)
+        elseif !symmetric 
+            # 2. Construct intervals for eta, then apply inverse link
+            lower = linkinv.(Link(mm), eta .- normalquantile .* sqrt.(vareta))
+            upper = linkinv.(Link(mm), eta .+ normalquantile .* sqrt.(vareta))
+        end
     else
         error("only :confidence intervals are defined")
     end
-    ciwidth = quantile(Normal(), (1. - level)/2) .* sqrt.(varmu)
-    (prediction = mu, lower = mu .+ ciwidth, upper = mu .- ciwidth)
+    (prediction = mu, lower = lower, upper = upper)
 end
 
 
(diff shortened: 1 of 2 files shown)


nalimilan:
Thanks!

@mkborregaard Since you implemented intervals for linear models, would you review this PR?

mkborregaard:
sure i'll have a look over the weekend

mkborregaard:
The predicton here seems based on a normal in the space of the input data - but shouldn't they be normal in the space of the link function then backtransformed?

nignatiadis:
Yes, I agree, since with your proposal the natural constraints on the predictions would be satisfied, while they may be violated with the current implementation. On the other hand,  if asymptotics have kicked in, then both intervals would be valid. 

The reason I implemented it in this way is 1) to conform with R's `glm` output which only returns standard errors (the intervals here are +- normal quantile * standard error with the standard error agreeing with R's) and 2) because I was interested in reproducing a figure from [CASI](https://web.stanford.edu/~hastie/CASI/) which had symmetric intervals around the logistic predictions.

mkborregaard:
Yes but I think R's `predict` function by default also returns the predictions in the space of the link function, right?

mkborregaard:
Which I think we shouldn't emulate

nignatiadis:
Yes that is correct. I also think it's not a great default to return predictions on link space, although it can be useful to have it as an option  (but that is for another issue..). However the R standard error I am talking about is the standard error you get if you explicitly ask for predictions on the scale of the response, i.e., if you do something like 
```r
predict(glm_object, newdata = data, type =  "response", se.fit = TRUE)
```

mkborregaard:
yes, but is it a bit strange isn't it? :-)

nignatiadis:
It is a bit strange but not too much.. At the end of the day these confidence intervals are not very trustworthy anyway (not even for the coefficients..) and only kick in for very large n/p (at which point both methods we  have been discussing are valid). To me they are mostly useful in that they provide a *rough* sense of variability.

In any case, to summarize our discussion, the pros are:

**Delta method** (current PR): Symmetric intervals, agree with `R` functions.

**Retransformation** (your suggestion): Satisfy natural constraints, my guess is that they might have better coverage in finite samples.

So I guess we both prefer your suggestion. Providing both choices would clutter the API too much, right?

nalimilan:
I agree returning symmetric confidence intervals is weird. That may make some sense for R as it only returns standard errors, but here we return both ends so better use the more appropriate method. Am I right that symmetric CIs may go beyond 0 or 1 for logistic regression?

Anyway, IMO providing both choices would be good, that function isn't too complex anyway and it would be nice to allow reproducing R's results exactly.

nignatiadis:
Yes you are right about the intervals going beyond to 0 to 1 potentially for logistic regression. I will add the other option for intervals sometime later this week.

mkborregaard:
But just to be sure - R doesn't have functionality for this, right? They just ask you to compute it yourself? I'd be in favour of returning the 95% confidence intervals that have the property that they contain the 95% closest lines to the mean line

nignatiadis:
No, I think R only returns standard errors and no confidence intervals (at least that is true for the `glm.predict` function). Which means we can match their output if we construct approximately 68% intervals :P. @mkborregaard I am not completely sure what you mean by the last sentence. So not including both options?

I tried to check what other languages do and came up with the following (i.e., seems like no default code, just do it yourself) :
* This cross-validated thread: https://stats.stackexchange.com/questions/354098/calculating-confidence-intervals-for-a-logistic-regression
* Also a Stackoverflow thread where the accepted answer gives Python code for both interval constructions we are discussing here:
https://stackoverflow.com/questions/47414842/confidence-interval-of-probability-prediction-from-logistic-regression-statsmode
* Agresti in his [Categorical data analysis](https://www.wiley.com/en-us/Categorical+Data+Analysis%2C+3rd+Edition-p-9780470463635) textbook seems to also use the transformation approach (i.e., compute intervals at linear predictor scale, then apply inverse link).

mkborregaard:
Here's what R's ggplot2 package does (borrowing the logistic smooth example from it's documentation):
```R
ggplot(rpart::kyphosis, aes(Age, as.numeric(Kyphosis) - 1)) +
    geom_jitter(height = 0.05) +
    binomial_smooth()
```
Screenshot 2020-04-22 at 19 37 35
So it seems that the default (or at least most used) R approach is also the Retransformation approach. Indeed I don't think there's an option to change that.

nignatiadis:
Oh that's a very cool finding! I did not even realize `geom_smooth` supported options beyond `lm` and the default smoother (..but then again I am spending less time in R these days). I will update the pull request over the weekend!

nignatiadis:
@mkborregaard  finally got around to updating this. Let me know what you think. I have included both symmetric and asymmetric intervals with default being asymmetric ones (for the reasons we discussed above). Furthermore, I added some tests by comparing to ggplot2 stat_smooth output, see code below for reproducibility.

```r
library(ggplot2)

logistic_fun <- function(x) exp(x)/(1+exp(x))
xs <- seq(0,1, length=50)
probs <- logistic_fun( -2 + 4*xs) 
set.seed(1)
ys <- rbinom(50, 1, probs)

df <- data.frame(x=xs, y=ys)

# get output from ggplot smooth
# see https://stackoverflow.com/questions/9789871/method-to-extract-stat-smooth-line-fit
ggres <- ggplot(df, aes(x=x,y=y)) + geom_smooth(method="glm", method.args = list(family = "binomial"))
tmpbuild <- ggplot_build(ggres)
data <- tmpbuild$data[[1]]

idx <- seq(1, 80, length=5)
data_sub <- data[idx,] # values to compare against
x_sub <- data_sub$x

# now also check se.fit from `glm` call
tmp_glm <- glm(y~x, df, family=binomial())
newdata = data.frame(x=x_sub)
predict(tmp_glm, newdata=newdata, type="response", se.fit=TRUE)$se.fit
```

nignatiadis:
@mkborregaard: tiny bump in case you have time to look at this!

nignatiadis:
Another tiny bump, it would be great to have this merged!

mkborregaard:
When a man says he'll do something he'll do it, no need to remind him every 6 months! :trollface: :joy:

mkborregaard (approved):
LGTM :+1:
Really, really sorry about the delay.

test/runtests.jl
```diff
@@ -458,7 +458,15 @@ end
 
     newX = rand(5, 2)
     newY = logistic.(newX * coef(gm11))
-    @test isapprox(predict(gm11, newX), newY)
+    gm11_pred1 = predict(gm11, newX)
+    gm11_pred2 = predict(gm11, newX; interval=:confidence, symmetric=true)
+    gm11_pred3 = predict(gm11, newX; interval=:confidence, symmetric=false)
+    @test gm11_pred1 == gm11_pred2.prediction == gm11_pred3.prediction≈ newY
+    se_pred = [0.19904587484129196, 0.18029108261296775,
+                   0.3290573571361879, 0.11536024793564569, 0.23972290956210984]
```
nalimilan, line 466:
```suggestion
    se_pred = [0.19904587484129196, 0.18029108261296775,
               0.3290573571361879, 0.11536024793564569, 0.23972290956210984]
```

test/runtests.jl
```diff
@@ -458,7 +458,15 @@ end
...
+    @test gm11_pred2.lower ≈ gm11_pred2.prediction .- quantile(Normal(), 0.975).*se_pred #
```
nalimilan, line 467:
```suggestion
    @test gm11_pred2.lower ≈ gm11_pred2.prediction .- quantile(Normal(), 0.975).*se_pred
```

test/runtests.jl
```diff
@@ -504,6 +512,34 @@ end
...
+    preds_symmetric = predict(gm, newX, interval=:confidence, symmetric=true)
+
+    @test preds_asymmetric.prediction == preds_symmetric.prediction
+    @test preds_asymmetric.prediction ≈ ggplot_prediction atol=1e-3
+    @test preds_asymmetric.lower ≈ ggplot_lower atol=1e-3
+    @test preds_asymmetric.upper ≈ ggplot_upper atol=1e-3
+
+    @test preds_symmetric.upper .-  preds_symmetric.lower ≈ 2 .* 1.96 .* R_glm_se atol=1e-3
+
+    # now fit model at level so that CI width is equal to standard error
```
nalimilan, line 538: That sounds like a weird test IMHO. It's really the same as checking `2 .* 1.96 .* R_glm_se` as above.

nalimilan (commented):
Cool. Just a few final remarks then.

src/glmfit.jl
```diff
@@ -522,9 +529,38 @@ function predict(mm::AbstractGLM, newX::AbstractMatrix;
...
+        # (2nd step varies as `symmetric` is `true` or `false`)
+        # 1. Estimate variance for eta based on variance for coefficients
+        #    through the diagonal of newX*vcov(mm)*newX'
+        vcovXnewT = vcov(mm)*newX'
+        vareta = [dot(view(newX, i, :), view(vcovXnewT, :, i)) for i in axes(newX,1)]
+        
+        if symmetric
+            # 2. Now compute the variance for mu based on variance of eta and
+            # construct intervals based on that
+            varmu = vareta .* [abs2(mueta(Link(mm), x)) for x in eta]
```
nignatiadis, line 548:
```suggestion
            stdmu = stdeta .* abs.(mueta.(Link(mm), eta))
            lower = mu .- normalquantile .* stdmu
            upper = mu .+ normalquantile .* stdmu
```

src/glmfit.jl
```diff
@@ -522,9 +529,38 @@ function predict(mm::AbstractGLM, newX::AbstractMatrix;
...
+            lower = mu .- normalquantile .* sqrt.(varmu)
+            upper = mu .+ normalquantile .* sqrt.(varmu)
```
nignatiadis, line 550:
```suggestion
            stdmu = stdeta .* abs.(mueta.(Link(mm), eta))
            lower = mu .- normalquantile .* stdmu
            upper = mu .+ normalquantile .* stdmu
```

test/runtests.jl
nignatiadis: Fully agree, going to remove this!

src/glmfit.jl
```diff
@@ -507,13 +507,20 @@ function dispersion(m::AbstractGLM, sqr::Bool=false)
...
-an offset.
+    predict(mm::AbstractGLM, newX::AbstractMatrix; offset::FPVector=eltype(newX)[],
+            interval::Union{Symbol,Nothing}=nothing, level::Real = 0.95,
+            symmetric::Bool = false)
+
+If `interval` is `nothing` (the default), return the predicted response of model
+`mm` from covariate values `newX` and, optionally, an `offset`. If `interval` is
+`:confidence`, also return upper and lower bounds for a given coverage `level`. By 
+default (`symmetric = false`) the intervals are constructed by applying the inverse link
+to intervals for the linear predictor, if `symmetric=true`, the intervals are symmetric
```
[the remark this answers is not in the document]
nignatiadis: That is true. I guess it depends on what you mean by "correct" :D. These intervals have correct coverage under old school asymptotics (number of features fixed, number of samples to infinity, design matrix well-behaved...). I know that does not mean a lot, but I would not rigorously trust the "asymmetric" intervals either. 

I am hoping we can keep this option as my original goal for this PR was to reproduce a figure from [CASI](https://web.stanford.edu/~hastie/CASI/) which had symmetric intervals around the logistic predictions. If you don't like this, maybe we can instead also provide a standard error option like R does? Although that would make the interface clunkier I think.

src/glmfit.jl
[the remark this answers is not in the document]
nignatiadis: Slightly changed this suggested change (to use `stdeta` instead of `vareta`).

nignatiadis:
Thanks @mkborregaard and @nalimilan for taking the time to review this! I did most of the major changes (but I am not sure why nightly fails).

src/glmfit.jl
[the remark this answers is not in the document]
nalimilan: OK, if Efron and Hastie use that... :-D

Does this approach have a name in the literature or existing software though? "symmetric" is rather the consequence of computing the interval on the response scale instead of the transformed scale, but it doesn't really say what's the difference (why would somebody prefer non-symmetric intervals? :-p).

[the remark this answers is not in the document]
nignatiadis: Ok so after rereading that part, Efron/Hastie plot predictions +- 1 standard error and say it's the standard error (they don't call it a confidence interval or anything like that).
  
I searched a bit more:
* It appears that STATA (https://www.stata.com/support/faqs/statistics/standard-error-predicted-probability/) uses "symmetric" intervals (though they don't give a name to these).
*  Cox, C., & Ma, G. (1995). "Asymptotic confidence bands for generalized nonlinear regression models". Biometrics, 142-150. use the name " asymmetric intervals" for the transformation intervals (I agree that's not good though).
* [Xu and Long (2005)](https://journals.sagepub.com/doi/pdf/10.1177/1536867X0500500405) use the names "endpoint transformation intervals" and "delta method intervals" for the asymmetric, resp. symmetric intervals. I think that sounds more reasonable?

One proposal then would be to change the keyword argument from `symmetric` to `interval_method` and then in the docs:

#
By default (`interval_method = :endpoint_transformation`) the intervals are constructed by applying the inverse link
to intervals for the linear predictor. If `interval_method = :delta`, the intervals are constructed by the delta method, i.e., by linearization of the predicted response around the linear predictor. The `:delta` method intervals are symmetric around the point estimates, but do not respect natural parameter constraints (e.g., the lower bound for a probability could be negative).
# 
What do you think?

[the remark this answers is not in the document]
nalimilan: OK, `:delta` indeed sounds like a better description. Though you mention that method in a comment that applies to both branches in the code so it probably needs adjusting.

For the default method, maybe just `:endpoint_transformation` is quite long? Maybe `:transformation` would be enough? Or maybe we can find a term that describes how the values are computed before the transformation?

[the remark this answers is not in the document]
nignatiadis: Sounds good! I will make sure to adjust that comment when I update the file.  `:transformation` sounds good to me as well. (I don't have a good idea about how to name it any other way; before transformation it's just x'\hat{\beta}, so just a linear transformation of the "approximately" Gaussian \hat{\beta}.)

src/glmfit.jl
```diff
@@ -535,23 +540,25 @@ function predict(mm::AbstractGLM, newX::AbstractMatrix;
...
             stdmu = stdeta .* abs.(mueta.(Link(mm), eta))
             lower = mu .- normalquantile .* stdmu
             upper = mu .+ normalquantile .* stdmu
-        elseif !symmetric 
+        elseif interval_method == :transformation
             # 2. Construct intervals for eta, then apply inverse link
             lower = linkinv.(Link(mm), eta .- normalquantile .* stdeta)
             upper = linkinv.(Link(mm), eta .+ normalquantile .* stdeta)
+        else
+            error("interval_method can be only :transformation or :delta")
```
nalimilan, line 561:
```suggestion
            throw(ArgumentError("interval_method can be only :transformation or :delta"))
```

src/glmfit.jl
```diff
@@ -507,13 +507,25 @@ function dispersion(m::AbstractGLM, sqr::Bool=false)
...
-an offset.
+    predict(mm::AbstractGLM, newX::AbstractMatrix; offset::FPVector=eltype(newX)[],
+            interval::Union{Symbol,Nothing}=nothing, level::Real = 0.95,
+            interval_method::Symbol = :transformation)
+
+If `interval` is `nothing` (the default), return the predicted response of model
+`mm` from covariate values `newX` and, optionally, an `offset`. If `interval` is
+`:confidence`, also return upper and lower bounds for a given coverage `level`. By default
+(`interval_method = :transformation`) the intervals are constructed by applying the
+inverse link to intervals for the linear predictor. If `interval_method = :delta`,
```
nalimilan, line 518:
```suggestion
```suggestion
Return the predicted response of model `mm` from covariate values `newX` and,
optionally, an `offset`.

If `interval=:confidence`, also return upper and lower bounds for a given coverage `level`.
By default (`interval_method = :transformation`) the intervals are constructed by applying
the inverse link to intervals for the linear predictor. If `interval_method = :delta`,
```

src/glmfit.jl
```diff
@@ -522,7 +534,36 @@ function predict(mm::AbstractGLM, newX::AbstractMatrix;
...
+            upper = mu .+ normalquantile .* stdmu
+        elseif interval_method == :transformation
+            # 2. Construct intervals for eta, then apply inverse link
+            lower = linkinv.(Link(mm), eta .- normalquantile .* stdeta)
+            upper = linkinv.(Link(mm), eta .+ normalquantile .* stdeta)
+        else
+            error("interval_method can be only :transformation or :delta")
+        end
+    else
+        error("only :confidence intervals are defined")
```
nalimilan, line 564:
```suggestion
        throw(ArgumentError("only :confidence intervals are defined"))
```

test/runtests.jl
```diff
@@ -504,6 +512,31 @@ end
...
+    preds_transformation = predict(gm, newX, interval=:confidence, interval_method=:transformation)
+    preds_delta = predict(gm, newX, interval=:confidence, interval_method=:delta)
+
+    @test preds_transformation.prediction == preds_delta.prediction
+    @test preds_transformation.prediction ≈ ggplot_prediction atol=1e-3
+    @test preds_transformation.lower ≈ ggplot_lower atol=1e-3
+    @test preds_transformation.upper ≈ ggplot_upper atol=1e-3
+
+    @test preds_delta.upper .-  preds_delta.lower ≈ 2 .* 1.96 .* R_glm_se atol=1e-3
+    @test_throws ErrorException predict(gm, newX, interval=:confidence, interval_method=:undefined_method)
```
nalimilan, line 537:
```suggestion
    @test_throws ArgumentError predict(gm, newX, interval=:confidence, interval_method=:undefined_method)
    @test_throws ArgumentError predict(gm, newX, interval=:undefined)
```

nalimilan (approved):
Thanks, looks good! Just a few cosmetic suggestions.

src/glmfit.jl
nignatiadis, line 518:
```suggestion
Return the predicted response of model `mm` from covariate values `newX` and,
optionally, an `offset`.

If `interval=:confidence`, also return upper and lower bounds for a given coverage `level`.
By default (`interval_method = :transformation`) the intervals are constructed by applying
the inverse link to intervals for the linear predictor. If `interval_method = :delta`,
```

PR #367, opened 2020-04-03
Allow version 0.6 of CSV

andreasnoack:
Not sure why CompatHelper didn't handle this one. Possibly because it's just a test dependency. cc @DilumAluthge


DilumAluthge:
Yeah, right now compat helper only handles required dependencies; it doesn’t do anything with test dependencies. It would be cool to add that as a feature though.

PR #369, opened 2020-04-14
convert all eltypes of GlmResp args to the eltype of the first

dsweber2:
min example: `GLM.GlmResp(randn(10), Binomial(), LogitLink(), randn(Float32, 10), randn(10))` Gets stuck in a loop because `float` won't convert `Float64` and `Float32` to the same type. I just turned the first one into a float, and then made the rest match that type of float explicitly.

### DIFF
--- a/src/glmfit.jl
+++ b/src/glmfit.jl
@@ -60,7 +60,9 @@ end
 
 function GlmResp(y::AbstractVector{<:Real}, d::D, l::L, off::AbstractVector{<:Real}, 
                  wts::AbstractVector{<:Real}) where {D, L}
-        GlmResp(float(y), d, l, float(off), float(wts))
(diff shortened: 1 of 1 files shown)


dsweber2:
digging a bit further, it seems the problem I was trying to address with this was actually caused in Lasso.jl, which was inconsistent about handing `Float32` and `Float64` to GLM.jl. I think this pull is still worth changing, but there are several changes elsewhere that are also needed if the input is somewhat inconsistent about the type of float involved.

nalimilan:
Thanks. Though note that `float` only makes a copy when necessary, contrary to what would happen with this PR. Can you change it to use `convert` instead? Also better use `promote_type` to choose the appropriate destination type.

PR #393, opened 2020-12-11
32-bit macOS does not exist

nalimilan:


kleinschmidt:
Thanks, good catch.  Any idea why the builds on nightly are timing out?

kleinschmidt (approved):

nalimilan:
No idea. And on Windows we even get errors... :-/

Issue #413, opened 2021-03-20
predict gives incorrect CIs when collinear columns have been dropped


nalimilan:
As noted at https://github.com/JuliaStats/GLM.jl/pull/410, `predict` appears to give incorrect results when some predictors are dropped because of collinearity. We should really do something about this.

```julia
julia> x = [1.0 1.0 2.0
            1.0 2.0 3.0
            1.0 -1.0 0.0];

julia> y = [1.0, 3.0, -2.0];

julia> m1 = lm(x, y, dropcollinear=true)
LinearModel{GLM.LmResp{Array{Float64,1}},GLM.DensePredChol{Float64,CholeskyPivoted{Float64,Array{Float64,2}}}}:

Coefficients:
─────────────────────────────────────────────────────────────────
        Coef.  Std. Error       t  Pr(>|t|)  Lower 95%  Upper 95%
─────────────────────────────────────────────────────────────────
x1   0.0       NaN         NaN       NaN     NaN        NaN
x2   2.07143     0.257539    8.04    0.0787   -1.20092    5.34378
x3  -0.428571    0.174964   -2.45    0.2468   -2.65169    1.79455
─────────────────────────────────────────────────────────────────


julia> m2 = lm(x[:, 1:2], y, dropcollinear=false)
LinearModel{GLM.LmResp{Array{Float64,1}},GLM.DensePredChol{Float64,Cholesky{Float64,Array{Float64,2}}}}:

Coefficients:
────────────────────────────────────────────────────────────────
        Coef.  Std. Error      t  Pr(>|t|)  Lower 95%  Upper 95%
────────────────────────────────────────────────────────────────
x1  -0.428571    0.174964  -2.45    0.2468  -2.65169     1.79455
x2   1.64286     0.123718  13.28    0.0479   0.070872    3.21484
────────────────────────────────────────────────────────────────


julia> p1 = predict(m1, x, interval=:confidence);

julia> p2 = predict(m2, x[:, 1:2], interval=:confidence);

julia> @test p1.prediction ≈ p2.prediction
Test Passed

julia> @test p1.upper ≈ p2.upper
Test Failed at REPL[254]:1
  Expression: p1.upper ≈ p2.upper
   Evaluated: [3.354342064529162; 5.728437323463455; 1.6152034019232366] ≈ [3.2437098232940285; 5.727181955909349; 1.200919478058661]
ERROR: There was an error during testing

julia> @test p1.lower ≈ p2.lower
[...]

PR #415, opened 2021-03-26
Cooks Distance tests

ericqu:
I added some tests, and an export. and use pretty much the original code. 

I wanted to try to do a PR and learn something along the way.

Unfortunately, there are a few errors, I am hoping you can help me with it.

I did ``` dev GLM``` in  the package manager, and then tried  ```test GLM``` but got some errors in the relevant test:

the first error is 

```
r2(t_lm) = 0.9532076162447004
Linear Model Cooks Distance - refers to PR #368 and issue #414: Error During Test at /home/eric/.julia/dev/GLM/test/runtests.jl:63
  Test threw exception
  Expression: isapprox(st_df.CooksD, cooksdistance(t_lm))
  MethodError: no method matching cooksdistance(::StatsModels.TableRegressionModel{LinearModel{GLM.LmResp{Array{Float64,1}},GLM.DensePredChol{Float64,CholeskyPivoted{Float64,Array{Float64,2}}}},Array{Float64,2}})
  Closest candidates are:
    cooksdistance(!Matched::LinearModel) at /home/eric/.julia/dev/GLM/src/lm.jl:251
  Stacktrace:
   [1] top-level scope at /home/eric/.julia/dev/GLM/test/runtests.jl:63
   [2] top-level scope at /home/eric/Repos/julia_install/julia-1.5.3/share/julia/stdlib/v1.5/Test/src/Test.jl:1115
   [3] top-level scope at /home/eric/.julia/dev/GLM/test/runtests.jl:53
```
Which I believe is a problem with the function signature.
I added a call to r2 as it also use a ```obj::LinearModel``` as an argument and does not throw an error. 
I thought about not forcing a type, but I guessed we want to make sure this a linear model and a generalized one.

Hence, I think I am missing something obvious although not sure what yet. 

The second error is:
```
r2(t_lm_w) = 0.9525895777180596
Linear Model Cooks Distance - refers to PR #368 and issue #414: Error During Test at /home/eric/.julia/dev/GLM/test/runtests.jl:52
  Got exception outside of a @test
  DomainError with -1.8726905285161664:
  sqrt will only return a complex result if called with a complex argument. Try sqrt(Complex(x)).
  Stacktrace:
   [1] throw_complex_domainerror(::Symbol, ::Float64) at ./math.jl:33
   [2] sqrt at ./math.jl:573 [inlined]
   [3] _broadcast_getindex_evalf at ./broadcast.jl:648 [inlined]
   [4] _broadcast_getindex at ./broadcast.jl:621 [inlined]
   [5] getindex at ./broadcast.jl:575 [inlined]
   [6] macro expansion at ./broadcast.jl:932 [inlined]
   [7] macro expansion at ./simdloop.jl:77 [inlined]
   [8] copyto! at ./broadcast.jl:931 [inlined]
   [9] copyto! at ./broadcast.jl:886 [inlined]
   [10] copy at ./broadcast.jl:862 [inlined]
   [11] materialize at ./broadcast.jl:837 [inlined]
   [12] stderror(::LinearModel{GLM.LmResp{Array{Float64,1}},GLM.DensePredChol{Float64,CholeskyPivoted{Float64,Array{Float64,2}}}}) at /home/eric/.julia/dev/GLM/src/linpred.jl:224
   [13] coeftable(::LinearModel{GLM.LmResp{Array{Float64,1}},GLM.DensePredChol{Float64,CholeskyPivoted{Float64,Array{Float64,2}}}}; level::Float64) at /home/eric/.julia/dev/GLM/src/lm.jl:195
   [14] coeftable(::LinearModel{GLM.LmResp{Array{Float64,1}},GLM.DensePredChol{Float64,CholeskyPivoted{Float64,Array{Float64,2}}}}) at /home/eric/.julia/dev/GLM/src/lm.jl:194
   [15] coeftable(::StatsModels.TableRegressionModel{LinearModel{GLM.LmResp{Array{Float64,1}},GLM.DensePredChol{Float64,CholeskyPivoted{Float64,Array{Float64,2}}}},Array{Float64,2}}; kwargs::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}) at /home/eric/.julia/packages/StatsModels/FJnLj/src/statsmodel.jl:174
   [16] coeftable at /home/eric/.julia/packages/StatsModels/FJnLj/src/statsmodel.jl:174 [inlined]
   [17] show(::IOContext{Base.GenericIOBuffer{Array{UInt8,1}}}, ::StatsModels.TableRegressionModel{LinearModel{GLM.LmResp{Array{Float64,1}},GLM.DensePredChol{Float64,CholeskyPivoted{Float64,Array{Float64,2}}}},Array{Float64,2}}) at /home/eric/.julia/packages/StatsModels/FJnLj/src/statsmodel.jl:185
   [18] show_delim_array(::IOContext{Base.GenericIOBuffer{Array{UInt8,1}}}, ::Tuple{StatsModels.TableRegressionModel{LinearModel{GLM.LmResp{Array{Float64,1}},GLM.DensePredChol{Float64,CholeskyPivoted{Float64,Array{Float64,2}}}},Array{Float64,2}}}, ::Char, ::Char, ::Char, ::Bool, ::Int64, ::Int64) at ./show.jl:776
   [19] show_delim_array at ./show.jl:761 [inlined]
   [20] show(::IOContext{Base.GenericIOBuffer{Array{UInt8,1}}}, ::Tuple{StatsModels.TableRegressionModel{LinearModel{GLM.LmResp{Array{Float64,1}},GLM.DensePredChol{Float64,CholeskyPivoted{Float64,Array{Float64,2}}}},Array{Float64,2}}}) at ./show.jl:794
   [21] _show_default(::IOContext{Base.GenericIOBuffer{Array{UInt8,1}}}, ::Any) at ./show.jl:406
   [22] show_default at ./show.jl:389 [inlined]
   [23] show(::IOContext{Base.GenericIOBuffer{Array{UInt8,1}}}, ::Any) at ./show.jl:384
   [24] sprint(::Function, ::MethodError; context::Pair{Symbol,Bool}, sizehint::Int64) at ./strings/io.jl:103
   [25] Test.Error(::Any, ::Any, ::Any, ::Any, ::Any) at /home/eric/Repos/julia_install/julia-1.5.3/share/julia/stdlib/v1.5/Test/src/Test.jl:162
   [26] do_test(::Test.ExecutionResult, ::Any) at /home/eric/Repos/julia_install/julia-1.5.3/share/julia/stdlib/v1.5/Test/src/Test.jl:518
   [27] top-level scope at /home/eric/.julia/dev/GLM/test/runtests.jl:66
   [28] top-level scope at /home/eric/Repos/julia_install/julia-1.5.3/share/julia/stdlib/v1.5/Test/src/Test.jl:1115
   [29] top-level scope at /home/eric/.julia/dev/GLM/test/runtests.jl:53
   [30] include(::String) at ./client.jl:457
   [31] top-level scope at none:6
   [32] eval(::Module, ::Any) at ./boot.jl:331
   [33] exec_options(::Base.JLOptions) at ./client.jl:272
   [34] _start() at ./client.jl:506
```

here it seems that when used with weights when calling the stderrror (``` stderror(x::LinPredModel) = sqrt.(diag(vcov(x)))``` )I suppose some entries in the diag vcov are negative. Could it be related to the weights?


ericqu:
For reference, I used the following program to generate the test data with SAS.

```sas
proc reg data=work.import;
	model Y = XA XB;
	output out=RegOut pred=Pred cookd=CooksD;
run;quit;

proc reg data=work.import;
	weight ww;
	model Y = XA XB;
	output out=RegOutW pred=Pred cookd=CooksD;
run;quit;
```

ericqu:
So looking further into the first issue. 
It appears that when calling the ```lm``` function to do the fit it eventually calls fit in ```StatsModels``` which in turn will redirect calls to r2 and adjr2 to the implementation in LinearModel when adequate (which explains why my call to r2 was working). 
This would suggest that to implement the ```cooksdistance``` (or any additional methods) there would be a needs to make some changes in StatsModels as well. While this seems useful for some "generic" functions, for specific ones it would seem useful to short-circuit this. And I am not sure how to that adequately.

ericqu:
By adding the following lines and alias method:
```julia
using StatsModels

function cooksdistance(obj :: StatsModels.TableRegressionModel)
    cooksdistance(obj.model)
end  
```
While this manages to get rid of the problems. I am not sure this is the way to go as it introduces a dependency on StatsModel, or is it alright?

For the weighted version, it appears that the results differ between SAS and the implementation. I will look into it.

kleinschmidt:
Yeah this is a wart of the way that GLM integrates with StatsModels at the moment.  See https://github.com/JuliaStats/StatsModels.jl/issues/32 for some discussion.  The short version is that GLM doesn't at the moment know anything about the formula syntax (`y ~ 1 + x`), it only knows about a `y` vector and an `X` matrix.  The original idea way back in the day was that StatsModels (formerly part of DataFrames) would hijack the `fit(::RegressionModel, ::Formula, ::DataFrame, ...)` methods, generate the arrays needed to fit the model, and wrap the whole thing in its own struct (now `TableRegressionModel`).  The hope was that you could The now-obvious shortcoming of that method is the one you're seeing now: only functions which are "blessed" in StatsModels themselves work with that wrapper (see [here](https://github.com/JuliaStats/StatsModels.jl/blob/0d0c2bbd56b183491e8c831a92168d4265e4ea68/src/statsmodel.jl#L117-L133)).

In the long term teh plan is to get rid of that altogether and have packages liek GLM take a dependency on StatsModels, but I'm afraid that keeps getting pushed back because there are some design issues.  See #339 for @nalimilan's progress in re-working GLM along those lines.

I'm afraid that it's not totally clear to me where that leaves otherwise good contributions like this.  I'd hate to see this PR languish for these upstream reasons but I don't see a way around it at this point.

ericqu:
Okay, that is disappointing.

I will still look into the Cook's distance with the weights scenario to complete this task and amend the PR accordingly. This is my first PR, so you might see some weird issues such as ```.vscode/settings.json``` which I will fix once I know more about GitHub.

I am struck that so much effort is spent establishing an architecture supporting different models. To take the example of the formula discussed in the above conversation. It matters little to me as a user if I use a different formula syntax between GLM LM or MixedModel (for which I will probably still need to do something special to define the fixed and random effects). For instance, the syntax in SAS/Proc REG (see above) is different, but that is a shallow learning effort to use one coming from the other. In contrast to not having features (such as the Cook's distance) that will dictate which tools I use.
While I understand that there needs to be a tradeoff between the inefficiencies resulting from completely independent models (say code duplication) and needed architecture to alleviate their costs, from my (user) perspective, I balance that with the features that interested parties could develop with a lower barrier of entry. 

Maybe the MixedModels guys have found solutions?
I see in (https://github.com/JuliaStats/MixedModels.jl/blob/main/src/linearmixedmodel.jl) the following is using a similar signature and is not part of StatsModel.:
```julia
"""
    feL(m::LinearMixedModel)
Return the lower Cholesky factor for the fixed-effects parameters, as an `LowerTriangular`
`p × p` matrix.
"""
function feL(m::LinearMixedModel)
    XyL = m.L[end]
    k = size(XyL, 1)
    inds = Base.OneTo(k - 1)
    LowerTriangular(view(XyL, inds, inds))
end
```
I will keep looking further.

nalimilan:
> I'm afraid that it's not totally clear to me where that leaves otherwise good contributions like this. I'd hate to see this PR languish for these upstream reasons but I don't see a way around it at this point.

Why not just add `cooksdistance` to the list of methods that StatsModels delegate to the underlying model? We can't stop adding features waiting for the switch to the new framework, given the absence of progress on that front.

ericqu:
Happy Monday everyone!

So I made some changes:

- I added a wrapper so that when called with a StatsModels.TableRegressionModel then use the LinearModel.
- I tested with the results from SAS and R for the Cook's Distance with weights, and unfortunately, the existing code did not give accurate results. It turns out that calculating the Cooks's Distance when the weights are normalized is relatively simple as only the residuals need to be weighted. 
- I added only a couple of tests so that the testing time isn't too long (I could add some more if you want to).
- It turns out that only the "F test for model comparison" is broken, although I think it is unrelated as it was broken from the beginning.

Hence I think this is good to go. Please let me know your comments or if you want me to change something.
I struggled with git/GitHub; it seems ok to me now. If it isn't, please let me know as well.

src/lm.jl
```diff
@@ -243,3 +243,46 @@ function confint(obj::LinearModel; level::Real=0.95)
...
+    u = residuals(obj)
+    mse = dispersion(obj,true)
+    k = dof(obj)-1
+    d_res = dof_residual(obj)
+    X = modelmatrix(obj)
+    wts = obj.rr.wts
+    if isempty(wts)
+        hii = diag(X * inv(X' * X) * X')
+    else
+        if mean(wts) == 1. # assume that the weights are normalised if the mean is 1
```
nalimilan, line 264: What does "normalized" mean in this context? In GLM.jl weights are frequency weights (i.e. integers most of the time), so their mean should probably never be 1. This differs from e.g. R.

src/lm.jl
```diff
@@ -243,3 +243,46 @@ function confint(obj::LinearModel; level::Real=0.95)
...
+    cooksdistance(obj :: StatsModels.TableRegressionModel)
+wrapper from TableRegressionModel to LinearModel. Will produce error when use with something else than LinearModel.
+"""
+function cooksdistance(obj :: StatsModels.TableRegressionModel)
+    if isa(obj.model , LinearModel)
+        cooksdistance(obj.model)
+    else
+        error("cooksdistance not implemented for model type: ", typeof(obj.model))
+    end
+end
```
nalimilan, line 288: Can you make a PR to add this to StatsModels instead?

src/lm.jl
ericqu: I see, well I assumed that the weights were weights and not frequencies.

Also the degree of freedom function seems to use weights as weights and not frequencies:
`dof(x::LinearModel) = length(coef(x)) + 1` (line 159 in lm.jl) in case of frequencies then this should be the sum of the frequencies there, right? I realised that I have be confused between weights and frequencies.

If the caller is thinking about weights, then I assumed it would be relatively simple for the caller to normalized the weights and since the beginning of the code was already done by the original author. 

Now if we want to be strict that they are frequencies, then I am afraid my analytical skills are not up to the work of finding the adequate calculation. I would just put `error` on line 264.

Let me know your preferred way.

src/lm.jl
ericqu: sure, can you tell me where (which file) I should put the code?
Is there is a risk that the team from StatsModels will reject it? as I suppose that both packages will need to make a release at relatively the same time.

nalimilan: It's at https://github.com/JuliaStats/StatsModels.jl/blob/0d0c2bbd56b183491e8c831a92168d4265e4ea68/src/statsmodel.jl#L123.

The team is the same, so just wait until @kleinschmidt gives his opinion about it. :-)

src/lm.jl
nalimilan: > Also the degree of freedom function seems to use weights as weights and not frequencies:
> `dof(x::LinearModel) = length(coef(x)) + 1` (line 159 in lm.jl) in case of frequencies then this should be the sum of the frequencies there, right? I realised that I have be confused between weights and frequencies.

`dof` gives the number of *consumed* degrees of freedom. The number of *residual* DOF is given by `dof_residual`, which is computed as `nobs(m) - dof(m) + 1`. `nobs` is in turn defined as the sum of weights:
https://github.com/JuliaStats/GLM.jl/blob/b39253f18e5d370c15ca8c7ac8d0f09c2dc04373/src/linpred.jl#L250

> Now if we want to be strict that they are frequencies, then I am afraid my analytical skills are not up to the work of finding the adequate calculation. I would just put `error` on line 264.

Usually frequency weights are the easier kind of weights to reason about since the results should be exactly equivalent to fitting the same model without weights but on data with each row repeated a number of times equal to its weight. But we would have to decide whether the distance for a given row should represent the effect of removing one observation corresponding to this row, or all of them (probably the former makes more sense). Anyway if you don't feel like implementing this we can just throw an error in the presence of weights for now.
if you don't 

Cc: @tbeason

test/runtests.jl
```diff
@@ -49,6 +49,21 @@ linreg(x::AbstractVecOrMat, y::AbstractVector) = qr!(simplemm(x)) \ y
...
+        XB=[1.8, 7.8, 11.8, 20.5, 17.3, 15.8], 
+        W=[0.6, 1.2, 1.2, 0.6, 1.2, 1.2],
+        CooksD=[1.7122291956, 18.983407026, 0.000118078, 0.8470797843, 0.0715921999, 0.1105843157], 
+        CooksDW=[0.7371619575, 30.979165951, 0.0112480511, 0.5168548491, 0.0600217986, 0.1270980264] )
+
+    t_lm = lm(@formula(Y ~ XA + XB), st_df)
+    @test isapprox(st_df.CooksD, cooksdistance(t_lm))
+    t_lm_w = lm(@formula(Y ~ XA + XB), st_df, wts = st_df.W)
+    @test isapprox(st_df.CooksDW, cooksdistance(t_lm_w))
+end
```
nalimilan, line 65: It would be useful to have a test for a model without an intercept, and for a model with two collinear columns (as we recently found buggy results in that case).

tbeason:
I'm afraid I won't be much help here! I am no expert, I just put my original PR together from the formulas I found online. I had tested it against the most common Python implementation and got identical results IIRC. I ended up not needing the functionality as I moved in a different direction, but I do think it is good to have something like this available somewhere.

Also, no need to cite me in the docstring, thanks for thinking of me though :smile:

src/lm.jl
nalimilan: Ah, actually there's a third step, as the (empty) function itself should be defined in StatsBase at https://github.com/JuliaStats/StatsBase.jl/blob/master/src/statmodels.jl. So that's 3 PRs!

nalimilan: @Nosferican You might be interested in this as you define `leverage` in Econometrics.jl.

test/runtests.jl
ericqu: Tests amended. I wasn't sure how collinear the variables needed to be. They have a variance inflation factor ≊ 250 in my test; I hope it is alright.

src/lm.jl
ericqu: @nalimilan, I understand the frequencies the same way you do. Although I am not sure there is any for us to decide with regards to the Cooks's distance as the results should match what is present in other packages (like SAS). I want to implement it but I realized that I am laking the skills to do it correctly. If there is a known formula let me know and I will plug it in.
If it is helpful I could add the tests for the frequencies, it is not an issue.

ericqu (commented):
Tests updated, awaiting further feedback.

The following points remains open as I do not know how to fix them.

- It turns out that the "F test for model comparison" is broken, although I think it is unrelated as it was broken from the beginning.
- and "NegativeBinomial NegativeBinomialLink Fixed θ " is broken as well but only for Julia nightly. It seems unrelated to the Cook's Distance.

test/runtests.jl
nalimilan: Sorry, I meant 100% collinear variables, like `1:10` and `2:11`. The point is just to test that results are correct when a variable is automatically omitted.

src/lm.jl
nalimilan: Let's just throw an error in the presence of weights if we don't know the formula.

src/lm.jl
nalimilan: @kleinschmidt is OK, so can you file PRs in StatsBase and StatsModels?

nalimilan:
Yes, don't worry about the failures, they are due to unrelated printing changes in Julia 1.6.

test/runtests.jl
ericqu: Good catch! Indeed it doesn't work. 
I think the problem is that the call to ```modelmatrix(obj)``` returns the same matrix no matter if lm/fit was called with dropcollinear=true or false as if the collinear variable is not dropped. And subsequently, the code fails at trying to invert the matrix.
My wishful thinking is that the cook's distance is meant to be called after fitting the model and hence should expect the collinear variable(s) to be dropped already. What do you think @nalimilan ?
I see that Issue #413 is on a similar topic; could the root cause be the same?

src/lm.jl
ericqu: Ok, I will do it when everything else is green.

test/runtests.jl
nalimilan: OK. Throwing an error is better than silently returning incorrect results. Please just turn the collinearity test into `@test_throws SingularException ...`. The error could probably be fixed by subsetting the `X` matrix to exclude columns for which `coef` is exactly zero, but that doesn't have to be fixed immediately.

src/lm.jl
ericqu: ok done:

- https://github.com/JuliaStats/StatsBase.jl/pull/679
- https://github.com/JuliaStats/StatsModels.jl/pull/221

src/lm.jl
ericqu: done

ericqu:
@nalimilan could you kindly review the PR again? Unexpectedly the compilation im Staatsmodel does not work quite well, it is not clear to me if that is expected.
Of course since the delegation is now in Stats Model the code here does not compile. That's expected, right?

ericqu:
@nalimilan, I noticed that the PR is not merged, is there anything I could do to help finalize it?

nalimilan:
Yes you'll need a StatsModels release first, and then bump the minimum versions for StatsBase and StatsModels in Project.toml.

kleinschmidt:
I've merged the StatsModels PR and tagged a new version 0.6.22: https://github.com/JuliaRegistries/General/pull/34686

src/lm.jl
```diff
@@ -243,3 +243,27 @@ function confint(obj::LinearModel; level::Real=0.95)
...
+
+"""
+    cooksdistance(obj::LinearModel)
+
+Compute [Cook's distance](https://en.wikipedia.org/wiki/Cook%27s_distance)
+for each observation in linear model `obj`, giving an estimate of the influence
+of each data point.
+Currently only implemented for LinearModel models without weights.
+"""
+function cooksdistance(obj::LinearModel)
```
palday, line 255:
```suggestion
function StatsBase.cooksdistance(obj::LinearModel)
```

.gitignore
```diff
@@ -1,3 +0,0 @@
-docs/build
-docs/site
-docs/Manifest.toml
```
nalimilan, line 3: Please revert this removal.

nalimilan (commented):
Should be OK at last with two minor fixes!

palday (changes requested):
We definitely want `cooksdistance` to be a method of `StatsBase.cooksdistance`, so this requires that the compat entry for StatsBase is updated in  `Project.toml`:

```toml
StatsBase = "0.33.5"
```

src/GLM.jl
```diff
@@ -17,7 +17,8 @@ module GLM
     import SpecialFunctions: erfc, erfcinv, digamma, trigamma
     export coef, coeftable, confint, deviance, nulldeviance, dof, dof_residual,
            loglikelihood, nullloglikelihood, nobs, stderror, vcov, residuals, predict,
-           fitted, fit, fit!, model_response, response, modelmatrix, r2, r², adjr2, adjr²
+           fitted, fit, fit!, model_response, response, modelmatrix, r2, r², adjr2, adjr², 
+           cooksdistance
```
nalimilan, line 21: Also add `cooksdistance` to the list of imports from StatsBase.

nalimilan:
Also need to require the latest StatsModels as test don't pass without it.

src/lm.jl
```diff
@@ -243,3 +243,27 @@ function confint(obj::LinearModel; level::Real=0.95)
...
+"""
+function StatsBase.cooksdistance(obj::LinearModel)
+    u = residuals(obj)
+    mse = dispersion(obj,true)
+    k = dof(obj)-1
+    d_res = dof_residual(obj)
+    X = modelmatrix(obj)
+    wts = obj.rr.wts
+    if isempty(wts)
+        hii = diag(X * inv(X' * X) * X')
```
palday, line 263: This is a pretty expensive operation for computing the diagonal of the hat matrix. I need to write out some algebra to optimize this, but basically anytime you saw an inverse of a matrix of arbitrary size and dimension in numeric code, something has gone wrong. (channeling @dmbates here in my attitudes, but maybe he has the formula handy without needing to write some scribbles involving a Cholesky.....)

src/lm.jl
```diff
@@ -243,3 +243,27 @@ function confint(obj::LinearModel; level::Real=0.95)
...
+for each observation in linear model `obj`, giving an estimate of the influence
+of each data point.
+Currently only implemented for LinearModel models without weights.
+"""
+function StatsBase.cooksdistance(obj::LinearModel)
+    u = residuals(obj)
+    mse = dispersion(obj,true)
+    k = dof(obj)-1
+    d_res = dof_residual(obj)
+    X = modelmatrix(obj)
```
palday, line 260: Use `crossmodelmatrix` here, then you get `X` and `X'X` back, but `X'X` has been wrapped in `Symmetric`, which allows the linear algebra methods below to be more specialized.

src/lm.jl
palday: Actually, I wondering now if we should define (maybe in StatsBase?)

1. `hatmatrix` to match `crossmodelmatix`.
2. `diaghatmatrix` that is an efficient evaluation of *only* the diagonal of the hat matrix.

(thoughts @ararslan and @nalimilan ?)

palday:
@ericqu Sorry for the terse and incomplete reviews I'm taking quick looks during coffee breaks at the day job, so as annoying as many review rounds can be, I hope getting feedback sooner is nonetheless helpful.

src/lm.jl
palday: Okay, assuming that `X` is full rank (i.e. there are no collinear predictors, which your code already does with `inv`), then you can use the QR decomposition to compute the hat matrix more efficiently and in a numerically stable way:

```julia
julia> using LinearAlgebra

julia> X = [1 0; 1 1; 1 0; 1 1]
4×2 Matrix{Int64}:
 1  0
 1  1
 1  0
 1  1

julia> hat = X * inv(X'X) * X'
4×4 Matrix{Float64}:
 0.5  0.0  0.5  0.0
 0.0  0.5  0.0  0.5
 0.5  0.0  0.5  0.0
 0.0  0.5  0.0  0.5

julia> Q = qr(X).Q[1:size(X,1), 1:size(X,2)]
4×2 Matrix{Float64}:
 -0.5   0.5
 -0.5  -0.5
 -0.5   0.5
 -0.5  -0.5

julia> Q*Q'
4×4 Matrix{Float64}:
 0.5          5.55112e-17  0.5          8.32667e-17
 5.55112e-17  0.5          8.32667e-17  0.5
 0.5          8.32667e-17  0.5          1.11022e-16
 8.32667e-17  0.5          1.11022e-16  0.5

julia> round.(Q*Q'; digits=1)
4×4 Matrix{Float64}:
 0.5  0.0  0.5  0.0
 0.0  0.5  0.0  0.5
 0.5  0.0  0.5  0.0
 0.0  0.5  0.0  0.5

julia> hat - Q*Q' .< eps(Float64) # everything within machine epsilon 
4×4 BitMatrix:
 1  1  1  1
 1  1  1  1
 1  1  1  1
 1  1  1  1
```

src/lm.jl
ericqu: done

src/lm.jl
ericqu: Well, my intention was to write some tests to finalize the implementation of Cook's Distance. 
It feels to me that the optimizations you're suggestions would optimize the code, but I am not sure where to start as to code it or to test it, nor in which package it makes the most sense to introduce it.
Could this be done in subsequent PR, by someone more skilled than me in optimizing the code and with a better knowledge of the intricacies of GLM, StatsModel, and StatsBase?

ericqu:
@palday thank you for taking the time to review and comment. I implemented the quick change about the `crossmodelmatrix`. 
Probably because of my lack of experience (this is my first PR), it is rather difficult for me to test implemented change locally: I used `pkg> dev GLM`, but this doesn't push back the changes to my own fork of GLM, I am not sure how to make changes to additional package(s) while being able to test the results locally before committing to PRs.

src/lm.jl
ericqu: @palday well the collinearity is still an issue. It is dealt with only in the sense that invoking the calculation under the wrong condition provides an error.

test/runtests.jl
```diff
@@ -49,6 +49,36 @@ linreg(x::AbstractVecOrMat, y::AbstractVector) = qr!(simplemm(x)) \ y
...
+
+    # linear regression, no intercept 
+    t_lm_noint = lm(@formula(Y ~ XA +0), st_df)
+    @test isapprox(st_df.CooksD_noint, cooksdistance(t_lm_noint))
+
+    # linear regression, two collinear variables (Variance inflation factor ≊ 250)
+    t_lm_multi = lm(@formula(Y ~ XA + XB), st_df)
+    @test isapprox(st_df.CooksD_multi, cooksdistance(t_lm_multi))
+
+    # linear regression, two full collinear variables (XC = 2 XA) hence should get the same results as base
```
palday, line 75: this isn't quite true: one of the collinear variables should be pivoted out and set to zero (`dropcollinear=true`) or error. 
```suggestion
    # linear regression, two full collinear variables (XC = 2 XA) hence should get the same results as the original
    # after pivoting
```

src/lm.jl
```diff
@@ -243,3 +243,28 @@ function confint(obj::LinearModel; level::Real=0.95)
...
+    wts = obj.rr.wts
+    cmm = crossmodelmatrix(obj)
+    if isempty(wts)
+        hii = diag(X * inv(cmm) * X')
```
palday, line 264:
```suggestion
    X, XtX = crossmodelmatrix(obj)
    k == size(X,2) || throw(ArgumentError("Models with collinear terms are not currently supported."))
    if isempty(wts)
        hii = diag(X * inv(XtX) * X')
```

test/runtests.jl
```diff
@@ -49,6 +49,36 @@ linreg(x::AbstractVecOrMat, y::AbstractVector) = qr!(simplemm(x)) \ y
...
+    t_lm_colli = lm(@formula(Y ~ XA + XC), st_df, dropcollinear=true)
+    ## Currently the test fails as the collinear variable is not droped from a ```modelmatrix(obj)``` call.
+    @test_throws SingularException isapprox(st_df.CooksD_base, cooksdistance(t_lm_colli))
```
palday, line 78:
```suggestion
    @test_throws ArgumentError isapprox(st_df.CooksD_base, cooksdistance(t_lm_colli))
```

palday (commented):
Getting very close. I'll let the other reviewers comment on style. 

@nalimilan I'll add my QR decomposition trick in a follow-up PR. I can specialize even further for the choldpred to use the existing Cholesky decomposition instead of doing a new QR.

src/lm.jl
nalimilan: Thanks @palday. I suspected there were better ways to do this. Though by default we use the Cholesky decomposition, so I guess we'd better whatever decomposition has already been computed (and stored) in the predictor object?

palday: @nalimilan Yep, I found some notes for dealing with QR or Cholesky efficiently.

test/runtests.jl
```diff
@@ -49,6 +49,36 @@ linreg(x::AbstractVecOrMat, y::AbstractVector) = qr!(simplemm(x)) \ y
...
 
+@testset "Linear Model Cooks Distance - refers to PR #368 and issue #414" begin
+    st_df = DataFrame( 
+        Y=[6.4, 7.4, 10.4, 15.1, 12.3 , 11.4],
+        XA=[1.5, 6.5, 11.5, 19.9, 17.0, 15.5],
+        XB=[1.8, 7.8, 11.8, 20.5, 17.3, 15.8], 
+        XC=[3., 13., 23., 39.8, 34., 31.],
+        CooksD_base=[1.4068501943, 0.176809102, 0.0026655177, 1.0704009915, 0.0875726457, 0.1331183932], 
+        CooksD_noint=[0.0076891801, 0.0302993877, 0.0410262965, 0.0294348488, 0.0691589296, 0.0273045538], 
+        CooksD_multi=[1.7122291956,	18.983407026, 0.000118078, 0.8470797843, 0.0715921999, 0.1105843157 ],
```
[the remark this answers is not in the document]
ericqu: yes, I mentioned that there: https://github.com/JuliaStats/GLM.jl/pull/415#issuecomment-808164128
should I put in the code comment as well?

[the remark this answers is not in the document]
palday: Yes, something like 

```julia
# reference values from SAS proc reg
```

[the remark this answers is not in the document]
ericqu: done

src/lm.jl
```diff
@@ -243,3 +243,28 @@ function confint(obj::LinearModel; level::Real=0.95)
...
+for each observation in linear model `obj`, giving an estimate of the influence
+of each data point.
+Currently only implemented for LinearModel models without weights.
+"""
+function StatsBase.cooksdistance(obj::LinearModel)
+    u = residuals(obj)
+    mse = dispersion(obj,true)
+    k = dof(obj)-1
+    d_res = dof_residual(obj)
+    X, XtX = crossmodelmatrix(obj)
```
palday, line 260: oops, I messed this up in my last suggestion, should be:
```suggestion
    X = modelmatrix(obj)
    XtX = crossmodelmatrix(obj)
```

ericqu:
> We definitely want `cooksdistance` to be a method of `StatsBase.cooksdistance`, so this requires that the compat entry for StatsBase is updated in `Project.toml`:
> 
> ```toml
> StatsBase = "0.33.5"
> ```

@palday or @nalimilan I made the change to address that. However, it keeps being presented as an issue for the PR. Is there anything that could be done?

palday (approved):
CI failures seem unrelated.

I want to get that explicit matrix inversion changed as soon as possible, but have that flagged in an issue.

I'm happy when @nalimilan is happy.

src/lm.jl
```diff
@@ -243,3 +243,29 @@ function confint(obj::LinearModel; level::Real=0.95)
...
     quantile(TDist(dof_residual(obj)), (1. - level)/2.) * [1. -1.]
 end
+
+"""
+    cooksdistance(obj::LinearModel)
+
+Compute [Cook's distance](https://en.wikipedia.org/wiki/Cook%27s_distance)
+for each observation in linear model `obj`, giving an estimate of the influence
+of each data point.
+Currently only implemented for LinearModel models without weights.
```
nalimilan, line 253:
```suggestion
Currently only implemented for linear models without weights.
```

test/runtests.jl
```diff
@@ -49,6 +49,38 @@ linreg(x::AbstractVecOrMat, y::AbstractVector) = qr!(simplemm(x)) \ y
...
+@testset "Linear Model Cooks Distance - refers to PR #368 and issue #414" begin
+    st_df = DataFrame( 
+        Y=[6.4, 7.4, 10.4, 15.1, 12.3 , 11.4],
+        XA=[1.5, 6.5, 11.5, 19.9, 17.0, 15.5],
+        XB=[1.8, 7.8, 11.8, 20.5, 17.3, 15.8], 
+        XC=[3., 13., 23., 39.8, 34., 31.],
+        # values from SAS proc reg
+        CooksD_base=[1.4068501943, 0.176809102, 0.0026655177, 1.0704009915, 0.0875726457, 0.1331183932], 
+        CooksD_noint=[0.0076891801, 0.0302993877, 0.0410262965, 0.0294348488, 0.0691589296, 0.0273045538], 
+        CooksD_multi=[1.7122291956,	18.983407026, 0.000118078, 0.8470797843, 0.0715921999, 0.1105843157 ],
```
nalimilan, line 61:
```suggestion
        CooksD_multi=[1.7122291956, 18.983407026, 0.000118078, 0.8470797843, 0.0715921999, 0.1105843157],
```

test/runtests.jl
```diff
@@ -49,6 +49,38 @@ linreg(x::AbstractVecOrMat, y::AbstractVector) = qr!(simplemm(x)) \ y
     @test coef(lm3) == coef(lm4) ≈ [11, 1, 2, 3, 4]
 end
 
+@testset "Linear Model Cooks Distance - refers to PR #368 and issue #414" begin
```
nalimilan, line 52:
```suggestion
@testset "Linear Model Cook's Distance" begin
```

test/runtests.jl
```diff
@@ -49,6 +49,38 @@ linreg(x::AbstractVecOrMat, y::AbstractVector) = qr!(simplemm(x)) \ y
...
+    @test isapprox(st_df.CooksD_noint, cooksdistance(t_lm_noint))
+
+    # linear regression, two collinear variables (Variance inflation factor ≊ 250)
+    t_lm_multi = lm(@formula(Y ~ XA + XB), st_df)
+    @test isapprox(st_df.CooksD_multi, cooksdistance(t_lm_multi))
+
+    # linear regression, two full collinear variables (XC = 2 XA) hence should get the same results as the original
+    # after pivoting
+    t_lm_colli = lm(@formula(Y ~ XA + XC), st_df, dropcollinear=true)
+    ## Currently the test fails as the collinear variable is not droped from a ```modelmatrix(obj)``` call.
```
nalimilan, line 79:
```suggestion
    # Currently fails as the collinear variable is not dropped from `modelmatrix(obj)`
```

nalimilan (approved):
Thanks!

Issue #224, opened 2018-05-02
Least square would fail, depending on the column ordering of X


richardkwo:
[data.txt](https://github.com/JuliaStats/GLM.jl/files/1966024/data.txt)

I encounter `ERROR: step-halving failed` for simple least squares.

Please use the attached `data.txt` to reproduce the error on a simple least square with a 200 x 6 design. 

```julia
using GLM

data = readdlm("data.txt")
y = data[:,1]
X = data[:,2:6]
glm(hcat(X, ones(size(X,1))), y, Normal(), IdentityLink())
```
When intercept is added as the last column, this works fine.
```
Coefficients:
      Estimate Std.Error   z value Pr(>|z|)
x1    -1.25395  0.166745  -7.52018   <1e-13
x2   -0.142528  0.113096  -1.26023   0.2076
x3    -0.02223  0.113199 -0.196381   0.8443
x4    0.111343   0.12555  0.886845   0.3752
x5   0.0216354  0.121658  0.177838   0.8589
x6    0.107489  0.147946  0.726543   0.4675
```

However, if the intercept is added as the first column, it will fail. 
```julia
glm(hcat(ones(size(X,1)), X), y, Normal(), IdentityLink())
```
I get `ERROR: step-halving failed at beta0 = [0.107489, -1.25395, -0.142528, -0.02223, 0.111343, 0.0216354].`, with the following
```
Stacktrace:
 [1] _fit!(::GLM.GeneralizedLinearModel{GLM.GlmResp{Array{Float64,1},Distributions.Normal{Float64},GLM.IdentityLink},GLM.DensePredChol{Float64,Base.LinAlg.Cholesky{Float64,Array{Float64,2}}}}, ::Bool, ::Int64, ::Float64, ::Float64, ::Void) at ~/.julia/v0.6/GLM/src/glmfit.jl:196
 [2] fit!(::GLM.GeneralizedLinearModel{GLM.GlmResp{Array{Float64,1},Distributions.Normal{Float64},GLM.IdentityLink},GLM.DensePredChol{Float64,Base.LinAlg.Cholesky{Float64,Array{Float64,2}}}}) at ~/.julia/v0.6/GLM/src/glmfit.jl:220
[...]

PR #416, opened 2021-04-05
Fix tests on Julia 1.6

nalimilan:
`@sprintf` has been fixed to use two digits as requested, but existing
tests use a single digit.


kleinschmidt (approved):
If CI is happy I'm happy.  The Windows failures look unrelated to this but may be real failures?  I don't know well enough to say...

nalimilan:
Funny that it happens only on Windows. The test is supposed to throw a `ConvergenceException` but for some reason it fails elsewhere. The error is similar to https://github.com/JuliaStats/GLM.jl/issues/224. Cc: @andreasnoack

andreasnoack:
The line search fails. Something like ten iterations without improvement. It's indeed odd that it only happens on Windows. I'll run it a VM to see if I can reproduce it.

andreasnoack:
I cannot reproduce this error on my Windows VM with 1.6.0.

PR #423, opened 2021-04-25
Verbose show with R² and F-statistic

OndrejSlamecka:
Adds R² and F-statistic, bringing the output closer to what users might expect when coming from R.

Example:

```
julia> ols_lin.model
LinearModel{GLM.LmResp{Vector{Float64}}, GLM.DensePredChol{Float64, LinearAlgebra.CholeskyPivoted{Float64, Matrix{Float64}}}}

Coefficients:
─────────────────────────────────────────────────────────────
     Coef.  Std. Error      t  Pr(>|t|)  Lower 95%  Upper 95%
─────────────────────────────────────────────────────────────
x1  -442.0     54.5539  -8.10    <1e-09  -551.688   -332.312
x2    51.0      1.8619  27.39    <1e-30    47.2564    54.7436
─────────────────────────────────────────────────────────────

R² = 0.9399
F-statistic: 750.3 on 1 and 50 degrees of freedom, p-value: 0.02898
```

### DIFF
--- a/src/linpred.jl
+++ b/src/linpred.jl
@@ -224,7 +224,38 @@ end
 stderror(x::LinPredModel) = sqrt.(diag(vcov(x)))
 
 function show(io::IO, obj::LinPredModel)
-    println(io, "$(typeof(obj)):\n\nCoefficients:\n", coeftable(obj))
+    println(io, typeof(obj))
+    println(io)
+    println(io, "Coefficients:")
+    println(io, coeftable(obj))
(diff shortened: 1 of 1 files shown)


OndrejSlamecka:
On second thought, there's no reason this can't be implemented in StatsModels, I'll make a PR there.

nalimilan:
R² and F-statistics are specific to linear models so StatsModels can't print these in general.

Issue #400, opened 2021-01-27
easy printing of fstat


compleathorseplayer:
Hi All.   What about the following for easy access to ftest()?

``` 
ftest(lmodel)=(1/(1-r2(lmodel))-1)*(length(predict(lmodel))-
    length(coef(lmodel)))/(length(coef(lmodel))-1)   
```

I have been giving this manually to my students, but it would be nice to have it
available as a method (at the moment ftest() with one argument fails)

I tried to make this into a Pull request but have never done that and got bogged
down with 'branch comparisons'.   Could someone please submit this if you think it is a good idea?

Maybe the function should be called ```fstat()``` to be analogous to the ```r2()``` diagnostic (a value, as opposed to a table)

andy-aa:
It also needs easy access to the p-value for the F-statistic.
```julia
function fstat(ols::LinearModel)
    k = length(coef(ols)) - 1
    n = nobs(ols)
    F = (1 / (1 - r2(ols)) - 1) * (n - k - 1) / k
    p = ccdf(FDist(k, n - k - 1), F)
    return F, p
end
```

palday:
I believe #424 addressed this.

PR #424, opened 2021-04-25
Implement fstatistic

OndrejSlamecka:
To be used by https://github.com/JuliaStats/StatsModels.jl/pull/225

I am not sure how much sense does `nobs::Float64` make but `Float64` is the return type of `nobs`.

### DIFF
--- a/src/ftest.jl
+++ b/src/ftest.jl
@@ -1,3 +1,10 @@
+struct SingleFTestResult
+    nobs::Int
+    dof::Int
+    fstat::Float64
+    pval::Float64
+end
+
 mutable struct FTestResult{N}
     nobs::Int
     ssr::NTuple{N, Float64}
@@ -27,6 +34,37 @@ end
 _diffn(t::NTuple{N, T}) where {N, T} = ntuple(i->t[i]-t[i+1], N-1)
 
 _diff(t::NTuple{N, T}) where {N, T} = ntuple(i->t[i+1]-t[i], N-1)
+
+"""
+    ftest(mod::LinearModel)
+
+Perform an F-test to determine whether model `mod` fits significantly better
+than the null model (i.e. which includes only the intercept).
+
+```jldoctest
+julia> dat = DataFrame(Result=[1.1, 1.2, 1, 2.2, 1.9, 2, 0.9, 1, 1, 2.2, 2, 2],
+                       Treatment=[1, 1, 1, 2, 2, 2, 1, 1, 1, 2, 2, 2]);
+
+julia> model = lm(@formula(Result ~ 1 + Treatment), dat);
+
+julia> ftest(model.model)
+F-test against the null model:
+F-statistic: 241.62 on 12 observations and 1 degrees of freedom, p-value: <1e-07
+```
+"""
+function ftest(mod::LinearModel)
+    hasintercept(mod) || throw(ArgumentError("ftest only works for models with an intercept"))
+
+    rss = deviance(mod)
+    tss = nulldeviance(mod)
+
+    n = Int(nobs(mod))
+    p = dof(mod) - 2 # -2 for intercept and dispersion parameter
+    fstat = ((tss - rss) / rss) * ((n - p - 1) / p)
+    fdist = FDist(p, dof_residual(mod))
+
+    SingleFTestResult(n, p, promote(fstat, ccdf(fdist, abs(fstat)))...)
+end
 
 """
     ftest(mod::LinearModel...; atol::Real=0.0)
@@ -85,9 +123,6 @@ F-test: 3 models fitted on 12 observations
 ```
 """
 function ftest(mods::LinearModel...; atol::Real=0.0)
-    if length(mods) < 2
-        Base.depwarn("Passing less than two models to ftest is deprecated", :ftest)
-    end
     if !all(==(nobs(mods[1])), nobs.(mods))
         throw(ArgumentError("F test is only valid for models fitted on the same data, " *
                             "but number of observations differ"))
@@ -125,6 +160,12 @@ function ftest(mods::LinearModel...; atol::Real=0.0)
     fstat = (NaN, (MSR1 ./ MSR2)...)
     pval = (NaN, ccdf.(FDist.(abs.(Δdf), dfr_big), abs.(fstat[2:end]))...)
     return FTestResult(Int(nobs(mods[1])), SSR, df, r2.(mods), fstat, pval)
+end
+
+function show(io::IO, ftr::SingleFTestResult)
+    print(io, "F-test against the null model:\nF-statistic: ", StatsBase.TestStat(ftr.fstat), " ")
+    print(io, "on ", ftr.nobs, " observations and ", ftr.dof, " degrees of freedom, ")
+    print(io, "p-value: ", PValue(ftr.pval))
 end
 
 function show(io::IO, ftr::FTestResult{N}) where N
--- a/src/linpred.jl
+++ b/src/linpred.jl
@@ -255,3 +255,5 @@ coef(x::LinPred) = x.beta0
 coef(obj::LinPredModel) = coef(obj.pp)
 
 dof_residual(obj::LinPredModel) = nobs(obj) - dof(obj) + 1
+
+hasintercept(m::LinPredModel) = any(i -> all(==(1), view(m.pp.X , :, i)), 1:size(m.pp.X, 2))
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "GLM"
 uuid = "38e38edf-8417-5370-95a0-9cbb8c7f171a"
-version = "1.5.1"
+version = "1.6.0"
 
 [deps]
 Distributions = "31c24e10-a181-5473-b8eb-7969acd0382f"
--- a/src/GLM.jl
+++ b/src/GLM.jl
@@ -15,10 +15,11 @@ module GLM
                       fitted, fit, model_response, response, modelmatrix, r2, r², adjr2, adjr², PValue
     import StatsFuns: xlogy
     import SpecialFunctions: erfc, erfcinv, digamma, trigamma
+    import StatsModels: hasintercept
     export coef, coeftable, confint, deviance, nulldeviance, dof, dof_residual,
            loglikelihood, nullloglikelihood, nobs, stderror, vcov, residuals, predict,
            fitted, fit, fit!, model_response, response, modelmatrix, r2, r², adjr2, adjr², 
-           cooksdistance
+           cooksdistance, hasintercept
 
     export
         # types
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -643,7 +643,62 @@ end
     @test preds_delta.upper .-  preds_delta.lower ≈ 2 .* 1.96 .* R_glm_se atol=1e-3
     @test_throws ArgumentError predict(gm, newX, interval=:confidence, interval_method=:undefined_method)
     @test_throws ArgumentError predict(gm, newX, interval=:undefined)
-end 
+end
+
+@testset "F test comparing to null model" begin
+    d = DataFrame(Treatment=[1, 1, 1, 2, 2, 2, 1, 1, 1, 2, 2, 2.],
+                  Result=[1.1, 1.2, 1, 2.2, 1.9, 2, .9, 1, 1, 2.2, 2, 2],
+                  Other=categorical([1, 1, 2, 1, 2, 1, 3, 1, 1, 2, 2, 1]))
+    mod = lm(@formula(Result~Treatment), d).model
+    othermod = lm(@formula(Result~Other), d).model
+    nullmod = lm(@formula(Result~1), d).model
+    bothmod = lm(@formula(Result~Other+Treatment), d).model
+    nointerceptmod = lm(reshape(d.Treatment, :, 1), d.Result)
+
+    ft1 = ftest(mod)
+    ft1base = ftest(nullmod, mod)
+    @test ft1.nobs == ft1base.nobs
+    @test ft1.dof ≈ dof(mod) - dof(nullmod)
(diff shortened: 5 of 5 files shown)


src/linpred.jl
```diff
@@ -227,13 +227,11 @@ function show(io::IO, obj::LinPredModel)
     println(io, "$(typeof(obj)):\n\nCoefficients:\n", coeftable(obj))
 end
 
-modelframe(obj::LinPredModel) = obj.fr
```
palday, line 230: why is this being removed?

src/linpred.jl
```diff
@@ -227,13 +227,11 @@ function show(io::IO, obj::LinPredModel)
...
 modelmatrix(obj::LinPredModel) = obj.pp.X
 response(obj::LinPredModel) = obj.rr.y
 
 fitted(m::LinPredModel) = m.rr.mu
 predict(mm::LinPredModel) = fitted(mm)
-formula(obj::LinPredModel) = modelframe(obj).formula
```
palday, line 236: why is this being removed?

src/lm.jl
```diff
@@ -185,6 +185,31 @@ function adjr2(obj::LinearModel)
...
+    fstat = ((tss - rss) / p) / (rss / (n - p - 1))
+    fdist = FDist(n, p)
+
+    FStatistic(n, p, fstat, ccdf.(fdist, abs(fstat)))
+end
+
+function show(io::IO, fstat::FStatistic)
+    print(io, "F-statistic: ", round(fstat.f, sigdigits=4), " ")
+    print(io, "on ", fstat.ndims, " and ", fstat.nobs, " degrees of freedom, ")
+    print(io, "p-value: ", round(fstat.fprob, sigdigits=4))
```
palday, line 210:
```suggestion
    print(io, "p-value: ", PValue(p))
```

src/lm.jl
```diff
@@ -185,6 +185,31 @@ function adjr2(obj::LinearModel)
     1 - (1 - r²(obj))*(n-1)/(n-p)
 end
 
+struct FStatistic
+    nobs::Float64
+    ndims::Int64
+    f::Float64
+    fprob::Float64
+end
```
palday, line 193: If you expect this to always be the return type of `fstatistic`, then it should be included in `StatsModels` with the function stub.

palday, line 193:
```suggestion
struct FStatistic{T<:Real}
    df1::T
    df2::T
    f::T
    p::T
end
```

Note that the denominator degrees of freedom in a an F test do not have to be an integer (e.g. when using the Welch-Satterthwaite correction). Likewise, neither the numerator nor the denominator are exactly the number of observations nor the number of dimensions. They are degrees of freedom and there are various cases where it makes sense for these to be something than integers. (This also comes up in mixed models, where the effective degrees of freedom for test statistics is somewhat different than just the number of parameters.)

src/lm.jl
```diff
@@ -185,6 +185,31 @@ function adjr2(obj::LinearModel)
...
+
+function StatsModels.fstatistic(obj::LinearModel)
+    rss = deviance(obj)
+    tss = nulldeviance(obj)
+
+    n = nobs(obj)
+    p = dof(obj) - 2 # -2 for dispersion parameter and intercept
+    fstat = ((tss - rss) / p) / (rss / (n - p - 1))
```
palday, line 201:
```suggestion
    fstat = ((tss - rss) * (n - p - 1)) / (rss * p)
```

Reduce the number of divisions with algebra :smile:

src/lm.jl
```diff
@@ -185,6 +185,31 @@ function adjr2(obj::LinearModel)
...
+    fdist = FDist(n, p)
+
+    FStatistic(n, p, fstat, ccdf.(fdist, abs(fstat)))
```
palday, line 204: why the broadcast `.` here?

src/linpred.jl
OndrejSlamecka: I didn't find any `LinPredModel` here that would have this attribute. There's `LinearModel` which doesn't and through `AbstractGLM` there's `GeneralizedLinearModel` which also doesn't have `fr`.

src/linpred.jl
OndrejSlamecka: This is dependent on my assessment for `modelframe` above.

src/lm.jl
OndrejSlamecka: Thanks, I'll add that.

src/lm.jl
OndrejSlamecka: I'd like to ask yesterday's me that question :- )

src/lm.jl
OndrejSlamecka: Thank you, I did not know about non-integer DOF so mentioning Welch-Satterthwaite correction gave me something to study!

There's one problem in the resulting code, that the arguments to FStatistic don't have the same type. I solved this with 
`    FStatistic(promote(n, p, fstat, ccdf(fdist, abs(fstat)))...)` but I am not certain that's the right Julia way.

OndrejSlamecka:
Thank you @palday for the suggestions, I implemented them, except un-deleting the two modelframe related functions for which I commented above. Is it desirable to also `export fstatistic` in `GLM.jl`?

palday:
> Thank you @palday for the suggestions, I implemented them, except un-deleting the two modelframe related functions for which I commented above. Is it desirable to also `export fstatistic` in `GLM.jl`?

I only focused on some Julia/style issues; I'll let @nalimilan or @kleinschmidt comment on API choices. :smile:

samusz:
Seems related to [issue #400 ](https://github.com/JuliaStats/GLM.jl/issues/400), if I am not mistaken.

nalimilan:
Sorry for not commenting earlier @OndrejSlamecka. How about making this a method of `ftest`? Since the p-value is printed, this is not just a function to get the F-statistic. That also simplifies the API.

OndrejSlamecka:
So in https://github.com/JuliaStats/GLM.jl/blob/504e5186c87b9bc3c7af70a69d570d329703b52a/src/ftest.jl#L88 turn, for example the following one into the one below? (To be clear, I just made up 42 and 0.42.)

```
julia> ftest(nullmodel.model, model.model)
F-test: 2 models fitted on 12 observations
─────────────────────────────────────────────────────────────────
     DOF  ΔDOF     SSR     ΔSSR       R²     ΔR²        F*  p(>F)
─────────────────────────────────────────────────────────────────
[1]    2        3.2292           -0.0000                         
[2]    3     1  0.1283  -3.1008   0.9603  0.9603  241.6234  <1e-7
─────────────────────────────────────────────────────────────────
```

```
julia> ftest(nullmodel.model, model.model)
F-test: 2 models fitted on 12 observations
─────────────────────────────────────────────────────────────────
     DOF  ΔDOF     SSR     ΔSSR       R²     ΔR²        F*  p(>F)
─────────────────────────────────────────────────────────────────
[1]    2        3.2292           -0.0000              42.0   0.42      
[2]    3     1  0.1283  -3.1008   0.9603  0.9603  241.6234  <1e-7
─────────────────────────────────────────────────────────────────
```

That'd be nice API-wise. A disadvantage is that it wouldn't show when viewing the model summary (https://github.com/JuliaStats/StatsModels.jl/pull/225) which is something to consider.

If you want to go ahead with the "extending `ftest`" approach, I'll update the code and the comment (I reckon it's a matter of dropping/replacing the two NaNs here https://github.com/JuliaStats/GLM.jl/blob/504e5186c87b9bc3c7af70a69d570d329703b52a/src/ftest.jl#L125).

nalimilan:
Adding this information to the first row of the table could be confusing, as it doesn't have the same meaning as in other rows. I was just thinking about using the `ftest` function with a single argument. The returned object could be a different type from `FTestResult`, similar to your `FStatistic` type (but with a name like `SingleFTestResult` or `FTest1Result`). Can you also put the new code in ftest.jl?

OndrejSlamecka:
@nalimilan I've done that now. Please let me know what you think.

src/ftest.jl
```diff
@@ -1,3 +1,10 @@
+struct SingleFTestResult{T<:Real}
+    df1::T
+    df2::T
```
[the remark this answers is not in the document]
palday: Always integers for classical ANOVA, but e.g. the Greenhouse-Geiser and Huyn-Feldt corrections work be changing the df to potentially non integers. Not sure if we want to leave room to support that in future releases.

src/ftest.jl
```diff
@@ -127,6 +155,12 @@ function ftest(mods::LinearModel...; atol::Real=0.0)
     return FTestResult(Int(nobs(mods[1])), SSR, df, r2.(mods), fstat, pval)
 end
 
+function show(io::IO, fstat::SingleFTestResult)
+    print(io, "F-statistic: ", round(fstat.f, sigdigits=4), " ")
```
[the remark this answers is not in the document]
palday: Would it make sense to use `StatsBase.TestStatistic` instead of `@printf` here?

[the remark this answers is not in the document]
nalimilan: Well I think I added that type in StatsBase only because it was needed to get two decimals while still using the matrix printing code. Other than that it doesn't have much benefit, does it?

That said, maybe we should print only 2 decimal digits rather than 4 for consistency.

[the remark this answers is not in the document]
palday: Consistency is why I suggested it: if all test statistics use `TestStatistic`, then they'll all be printed the same.

src/ftest.jl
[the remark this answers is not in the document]
nalimilan: Hmm, I guess changing the type later wouldn't be considered as breaking, right? We already use `Int` for `FTest` too FWIW.

src/ftest.jl
```diff
@@ -1,3 +1,10 @@
+struct SingleFTestResult{T<:Real}
+    df1::T
+    df2::T
+    f::T
+    p::T
```
nalimilan, line 5: Like for `FTest`, we can require that all values are `Float64`, as it's not useful to use other types in practice. Then the `T` type parameter can be dropped completely.
```suggestion
    f::Float64
    p::Float64
```

src/ftest.jl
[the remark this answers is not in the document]
palday: But yes, I agree it would be great to print numbers that are integers as integers, whether or not the underlying type is `Int`.

src/ftest.jl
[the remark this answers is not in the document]
nalimilan: Yeah, makes sense.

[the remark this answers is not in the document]
OndrejSlamecka: Interesting discussion! I made it `    print(io, "F-test against the null model:\nF-statistic: ", TestStat(fstat.f), " ")`

src/ftest.jl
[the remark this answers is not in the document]
OndrejSlamecka: I think this has gone from Int to parametric in the past. Maybe because of StatsModels but it seems we gave up on integrating with that here and also FTestResult{N} uses Int => I made it Int.

src/ftest.jl
```diff
@@ -78,16 +85,37 @@ F-test: 3 models fitted on 12 observations
...
      DOF  ΔDOF     SSR     ΔSSR       R²     ΔR²        F*   p(>F)
 ──────────────────────────────────────────────────────────────────
-[1]    2        3.2292           -0.0000                          
+[1]    2        3.2292           -0.0000
 [2]    3     1  0.1283  -3.1008   0.9603  0.9603  241.6234   <1e-7
 [3]    5     2  0.1017  -0.0266   0.9685  0.0082    1.0456  0.3950
 ──────────────────────────────────────────────────────────────────
 ```
+
+For a single model the F-statistic is printed.
```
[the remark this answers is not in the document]
OndrejSlamecka: Right. I split the descriptions.

src/ftest.jl
```diff
@@ -28,6 +35,34 @@ _diffn(t::NTuple{N, T}) where {N, T} = ntuple(i->t[i]-t[i+1], N-1)
...
+```
+"""
+function ftest(mod::LinearModel)
+    rss = deviance(mod)
+    tss = nulldeviance(mod)
+
+    n = nobs(mod)
+    p = dof(mod) - 2 # -2 for dispersion parameter and intercept
+    fstat = ((tss - rss) * (n - p - 1)) / (rss * p)
+    fdist = FDist(n, p)
```
nalimilan, line 61: This isn't the correct formula, is it? I'd expect something like `FDist(p-1, p)`.
Also this doesn't give the same result as the existing `ftest` method for the example given above in the docstring. Have you checked this against other implementations? BTW, can you add tests for the new method?

src/ftest.jl
```diff
@@ -1,3 +1,10 @@
+struct SingleFTestResult
+    dof::Int
+    nulldof::Int
```
nalimilan, line 3: Actually it would make sense to store `nobs` instead, for consistency with `FTestResult`. It's also nicer for interpretation. Then I'd print the number of observations in the output (like for the existing `ftest`).

src/ftest.jl
```diff
@@ -28,6 +35,34 @@ _diffn(t::NTuple{N, T}) where {N, T} = ntuple(i->t[i]-t[i+1], N-1)
...
+F-test against the null model:
+F-statistic: 241.62 on 12 and 1 degrees of freedom, p-value: 0.0502
+```
+"""
+function ftest(mod::LinearModel)
+    rss = deviance(mod)
+    tss = nulldeviance(mod)
+
+    n = nobs(mod)
+    p = dof(mod) - 2 # -2 for dispersion parameter and intercept
```
nalimilan, line 59: Maybe this method should check that `mod` includes the intercept? Otherwise I'm afraid results will be misleading.

nalimilan (approved):
Thanks! This is something we could consider moving to StatsModels so that other packages don't have to reimplement it. Would probably be worth filing an issue there to discuss it, if you feel like it.

PR #433, opened 2021-06-06
Update the perf script.

dmbates:
The perf/glm.jl script was using out-of-date constructions.  This includes updates to recent versions of the Random and StatsModels packages.  It turns out that most of the effort in fitting a large-ish Bernoulli model like this is in the linear algebra.  Using Julia v1.7.0-DEV just switching to MKL saves about 30% of the fitting time on my laptop.
```julia
julia> @time glm(pred, resp, Bernoulli());
  4.994811 seconds (46.76 k allocations: 2.052 GiB, 2.71% gc time, 0.44% compilation time)

julia> using MKL

julia> @time glm(pred, resp, Bernoulli());
  3.549232 seconds (40 allocations: 2.049 GiB, 2.88% gc time)
```

### DIFF
--- a/perf/glm.jl
+++ b/perf/glm.jl
@@ -1,21 +1,23 @@
-using GLM, DataFrames, StatsFuns
-glm(@formula(y ~ 1), DataFrame(y = float(bitrand(10))), Bernoulli())
-
+using GLM, Random, StatsModels
+                                # create a column table with dummy response
 n = 2_500_000
-Random.seed!(1234321)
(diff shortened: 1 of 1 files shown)


andreasnoack:
I think we'll have to adjust the random numbers in our tests to get tests passing on nightly. I'll take a look and then restart tests here once that has been fixed.

ViralBShah:
Should we merge this?

dmbates:
I would say yes, let's go ahead and merge it.

Issue #431, opened 2021-05-25
predict() gives wrong results when dropcollinear = true


githubtomtom:
the following example shows that `predict()` is wrong when `dropcollinear = true`:
1. `:confidence` and `:prediction` give the **same** predicitons
2. lower and upper bounds are unreasonably **wide**

```julia
Random.seed!(1)
X = randn(10, 3);
X[:, 2] .= X[:, 1] .- X[:, 3];
y = randn(10);

julia> mod1 = GLM.lm(X, y; dropcollinear = false)
GLM.LinearModel{GLM.LmResp{Vector{Float64}}, GLM.DensePredChol{Float64, LinearAlgebra.Cholesky{Float64, Matrix{Float64}}}}:

Coefficients:
─────────────────────────────────────────────────────────────────
        Coef.  Std. Error      t  Pr(>|t|)   Lower 95%  Upper 95%
─────────────────────────────────────────────────────────────────
x1   1.22558    3.65892e7   0.00    1.0000  -8.65198e7  8.65198e7
x2  -0.982437   3.65892e7  -0.00    1.0000  -8.65198e7  8.65198e7
x3  -0.5        3.65892e7  -0.00    1.0000  -8.65198e7  8.65198e7
─────────────────────────────────────────────────────────────────


julia> mod2 = GLM.lm(X, y; dropcollinear = true)
GLM.LinearModel{GLM.LmResp{Vector{Float64}}, GLM.DensePredChol{Float64, LinearAlgebra.CholeskyPivoted{Float64, Matrix{Float64}}}}:

Coefficients:
─────────────────────────────────────────────────────────────────
       Coef.  Std. Error       t  Pr(>|t|)   Lower 95%  Upper 95%
─────────────────────────────────────────────────────────────────
x1  0.0       NaN         NaN       NaN     NaN         NaN
x2  0.243147    0.302187    0.80    0.4443   -0.453697    0.93999
x3  0.725583    0.439743    1.65    0.1375   -0.288465    1.73963
─────────────────────────────────────────────────────────────────

julia> GLM.predict(mod1, X[1:1, :], interval = :confidence)
(prediction = [0.13054650028641046], lower = [-0.10362885085068049], upper = [0.36472185142350144])
julia> GLM.predict(mod1, X[1:1, :], interval = :prediction)
(prediction = [0.13054650028641046], lower = [-2.4585555782536694], upper = [2.71964857882649])

[...]

PR #410, opened 2021-03-09
Fix predict with collinear=true

nalimilan:
We need to reorder rows and columns when using pivoted Cholesky decomposition.

Fixes #409.

PR #436, opened 2021-06-21
Fix more `predict` confidence intervals when `dropcollinear=true`

nalimilan:
The previous fix (https://github.com/JuliaStats/GLM.jl/pull/410) handled incorreclty cases where pivoted Cholesky decomposition permutes columns in a non-monotonous way. This is because the inverted permutation has to be used, but that bug isn't visible in simple examples.

Fixes https://github.com/JuliaStats/GLM.jl/issues/431.


ararslan (approved):
Very nice catch!

PR #438, opened 2021-06-22
Throw an error in `predict` with intervals for rank-deficient models

nalimilan:
`predict` currently gives incorrect intervals when some variables have been dropped due to collinearity. Throw an error instead to
avoid misleading users, until we find a fix.

See https://github.com/JuliaStats/GLM.jl/pull/410#issuecomment-793920775.

### DIFF
--- a/src/lm.jl
+++ b/src/lm.jl
@@ -245,6 +245,11 @@ function predict(mm::LinearModel, newx::AbstractMatrix;
     end
     if interval === nothing
         return retmean
+    elseif mm.pp.chol isa CholeskyPivoted &&
+        mm.pp.chol.rank < size(mm.pp.chol, 2)
+        throw(ArgumentError("prediction intervals are currently not implemented " *
+                            "when some independent variables have been dropped " *
+                            "from model due to collinearity"))
     end
     length(mm.rr.wts) == 0 || error("prediction with confidence intervals not yet implemented for weighted regression")
     chol = cholesky!(mm.pp)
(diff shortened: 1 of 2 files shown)


src/lm.jl
```diff
@@ -245,6 +245,11 @@ function predict(mm::LinearModel, newx::AbstractMatrix;
     end
     if interval === nothing
         return retmean
+    elseif mm.pp.chol isa CholeskyPivoted &&
+        mm.pp.chol.rank < size(mm.pp.chol, 2)
+        throw(ArgumentError("prediction intervals are currently not implemented " *
+                            "when some independent variables have been dropped " *
+                            "from model due to collinearity"))
```
ararslan, line 252:
```suggestion
                            "from the model due to collinearity"))
```

test/runtests.jl
```diff
@@ -600,6 +600,23 @@ end
 
     # Deprecated argument value
     @test predict(m1, x, interval=:confint) == p1
+
+    # Prediction intervals would give incorrect results when some variables
+    # have been dropped due to collinearity (#410)
+    x = [1.0 1.0 2.0
+         1.0 2.0 3.0
+         1.0 -1.0 0.0];
```
ararslan, line 608:
```suggestion
         1.0 -1.0 0.0]
```

ararslan (approved):
LGTM otherwise, or ignore comments if you disagree

src/lm.jl
nalimilan: Glad I'm not the only one to make suggestions about determinants! :-D

ararslan: Perhaps it's a habit I picked up from you at some point 😉

nalimilan:
Thanks! I've filed https://github.com/JuliaStats/GLM.jl/pull/439 to release 1.4.3.

src/lm.jl
nalimilan: Argh..

PR #439, opened 2021-06-22
Bump version to 1.5.0

nalimilan:


bkamins:
It should technically be 1.5 as we added `cooksdistance` which is exported.

bkamins (approved):

nalimilan:
Good catch!

Issue #442, opened 2021-07-18
prediction fails on custom vectors


bkamins:
In https://github.com/JuliaStats/GLM.jl/blob/96f6eeba0f93c1271982e75575e303b354a2b412/src/linpred.jl#L129 (and related lines) we require `Vector{T}` in `delbeta!`. However, in general the user might pass some other type of `AbstractVector` and prediction fails.

I am not too much into the GLM.jl package to know the best approach to fix it (the safest one would be to make the signatures more flexible force conversion to `Vector`).

CC @quinnj + @nalimilan - I have caught it by checking what things the usage of `ChainedVector` can break, but this time the issue is on GLM.jl side.

nalimilan:
Given that this code is relatively low-level and relies on `BlasReal`, it seems safest to convert to `Vector{T}`. `ldiv!` for `Cholesky` seems to support only `StridedArray` objects anyway.

nalimilan:
That said, the vector is only used in `mul!(p.delbeta, transpose(p.X), r)`, so as long as `mul!` works it's probably fine to allow any kind of vector. If it works maybe it's OK to widen the signature.

bkamins:
I meant to widen the signature and internally do `convert(Vector{T}, whatever_was_passed)`, so for `Vector` it would be a no-op.

nalimilan:
Yeah that's what I understood. But in my second comment I wondered whether just widening the signature would be enough.

bkamins:
I have checked that `mul!` is fully generic, so I have opened https://github.com/JuliaStats/GLM.jl/pull/443.

andreasnoack:
Do you have a reproducer for this? It would be useful to see the exact code path that triggers this. As mentioned in https://github.com/JuliaStats/GLM.jl/pull/443#issuecomment-884476200, I'd think that it's better to initialize `r` differently to ensure that it's a `Vector`. It will be much easier to ensure reasonable performance with a more restrictive signature.

bkamins:
I have included the reproducers in tests of #443. Here they are repeated (notice that in #443 I have spotted that also `cholpred` requires fixing not only `delbeta!`):
```
julia> using GLM, SparseArrays

julia> X = sprand(5, 3, 0.5)
5×3 SparseMatrixCSC{Float64, Int64} with 7 stored entries:
 0.370538  0.226452   ⋅
 0.20658    ⋅         ⋅
  ⋅         ⋅        0.0668819
[...]

PR #443, opened 2021-07-21
Allow any vector in prediction

bkamins:
Fixes https://github.com/JuliaStats/GLM.jl/issues/442

### DIFF
--- a/src/linpred.jl
+++ b/src/linpred.jl
@@ -126,12 +126,12 @@ function cholesky(p::DensePredChol{T}) where T<:FP
 end
 cholesky!(p::DensePredQR{T}) where {T<:FP} = Cholesky{T,typeof(p.X)}(p.qr.R, 'U', 0)
 
-function delbeta!(p::DensePredChol{T,<:Cholesky}, r::Vector{T}) where T<:BlasReal
+function delbeta!(p::DensePredChol{T,<:Cholesky}, r::AbstractVector{T}) where T<:BlasReal
     ldiv!(p.chol, mul!(p.delbeta, transpose(p.X), r))
     p
 end
 
-function delbeta!(p::DensePredChol{T,<:CholeskyPivoted}, r::Vector{T}) where T<:BlasReal
+function delbeta!(p::DensePredChol{T,<:CholeskyPivoted}, r::AbstractVector{T}) where T<:BlasReal
     ch = p.chol
     delbeta = mul!(p.delbeta, adjoint(p.X), r)
     rnk = rank(ch)
@@ -148,7 +148,7 @@ function delbeta!(p::DensePredChol{T,<:CholeskyPivoted}, r::Vector{T}) where T<:
     p
 end
 
-function delbeta!(p::DensePredChol{T,<:Cholesky}, r::Vector{T}, wt::Vector{T}) where T<:BlasReal
+function delbeta!(p::DensePredChol{T,<:Cholesky}, r::AbstractVector{T}, wt::AbstractVector{T}) where T<:BlasReal
     scr = mul!(p.scratchm1, Diagonal(wt), p.X)
     cholesky!(Hermitian(mul!(cholfactors(p.chol), transpose(scr), p.X), :U))
     mul!(p.delbeta, transpose(scr), r)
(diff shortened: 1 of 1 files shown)


bkamins:
Do we want some tests of this change?

andreasnoack:
The residual vector here is created by this package so I'm wondering if we shouldn't initialize it as a `Vector` instead of allowing `r` to be an `AbstractVector`. I think it could easily be a performance trap since it can easily hit a slow generic fallback.

bkamins:
This was mas initial thought as you can see in #442. I have opened a PR to make sure the issue is fixed (in general if someone tries to make a prediction using data read-in by CSV.jl it is most likely to fail as CSV.jl produces `ChainedVector` now - so this is a pretty pressing thing).

However, I have not studied all the design of GLM.jl enough to make the best decision about what should be done here (so I would prefer to delegate it so core GLM.jl maintainers).

nalimilan:
If this gives poor performance for some custom array types, these should probably implement `mul!`, which is a quite fundamental operation.

I'd say a test would be good, both to ensure this is safe and to avoid regression in the future. Maybe just using a range is OK?

bkamins:
> Maybe just using a range is OK?

Range is OK, but I will also use a `view` and sparse vector then.

test/runtests.jl
```diff
@@ -502,6 +502,24 @@ end
...
+                 fit(GeneralizedLinearModel, Matrix(X), sparse(y))]
+    gmdense = fit(GeneralizedLinearModel, Matrix(X), y)
+
+    for gmsparse in gmsparsev
+        @test isapprox(deviance(gmsparse), deviance(gmdense))
+        @test isapprox(coef(gmsparse), coef(gmdense))
+        @test isapprox(vcov(gmsparse), vcov(gmdense))
+    end
+end
+
```
bkamins, line 521:
```suggestion
    gmsparsev = [fit(LinearModel, X, y),
                 fit(LinearModel, X, sparse(y)),
                 fit(LinearModel, Matrix(X), sparse(y))]
    gmdense = fit(LinearModel, Matrix(X), y)

    for gmsparse in gmsparsev
        @test isapprox(deviance(gmsparse), deviance(gmdense))
        @test isapprox(coef(gmsparse), coef(gmdense))
        @test isapprox(vcov(gmsparse), vcov(gmdense))
    end
end
```

bkamins:
I have added tests that prompted more changes that were required.

src/linpred.jl
```diff
@@ -185,15 +185,22 @@ function SparsePredChol(X::SparseMatrixCSC{T}) where T
         similar(X))
 end
 
-cholpred(X::SparseMatrixCSC) = SparsePredChol(X)
+cholpred(X::SparseMatrixCSC, ::Bool=false) = SparsePredChol(X)
```
nalimilan, line 188: Is this because we don't have any implementation of the sparse Cholesky decomposition? Also, the fact that this was missing indicates that this method wasn't tested at all, right?
```suggestion
# No implementation of pivoted Cholesky decomposition is available
cholpred(X::SparseMatrixCSC, pivot::Bool=false) = SparsePredChol(X)
```

bkamins: > the fact that this was missing indicates that this method wasn't tested at all, right?

Yes

> we don't have any implementation of the pivoted sparse Cholesky decomposition?

We would have to ask someone who knows `LinearAlgebra` better. I just changed it to make the tests pass. Currently `SparsePredChol` does not allow passing `pivot` argument, so I have not changed it.

andreasnoack:
I've taken a look at the reproducer. I'll propose that we change the `LmResp` constructor to something like
```julia
function LmResp(y::FPVector, wts::FPVector=similar(y, 0))
    newy = convert(Vector, y)
    newwts = convert(Vector, wts)
    return LmResp{typeof(newy)}(fill!(similar(newy), 0), similar(newy, 0), newwts, newy)
end
```
instead of loosening the `Vector` signature in `delbeta!`. As mentioned earlier, it will make it easier to ensure that things work and stay fast.

nalimilan:
https://github.com/JuliaStats/GLM.jl/issues/444 raises the question of how to handle views of `Vector`s. Is there any reason to convert them to `Vector`? Maybe we should at least allow them in the signature?

andreasnoack:
It's always simpler to avoid type parameters and I see little benefit in allowing arbitrary arrays type unless we have an actual demand for it such as e.g. getting the package working on GPUs. For now, I think we should just convert the input to a `Vector`. I believe that should fix #444.

bkamins:
As I am not maintaining GLM.jl please do not feel attached to my PR. I am OK with any solution that would work (and it can be refined later) - I just want GLM.jl to work with `view`s (as this was the case that also prompted me to open this PR)

nalimilan:
Any solution that fixes views is OK for me so let's go with what @andreasnoack suggests.

nalimilan:
Do you feel like fixing GLMs (https://github.com/JuliaStats/GLM.jl/issues/444) while you're at it? :-p

bkamins:
This is fixed

PR #446, opened 2021-09-06
Allow for sparse arrays and views

andreasnoack:
This is an alternative to https://github.com/JuliaStats/GLM.jl/pull/443 as discussed in that PR. I'm converting the inputs to `Vector` since we don't really support anything else even though the structs are parametric. Eventually, we can consider loosening this conversion if we want to support e.g. `CuArrays` but it would have to be a more extensive rewrite to ensure that things work and that no slow paths are hit in important cases so, for now, I think this PR is the best way forward. cc @bkamins

### DIFF
--- a/src/glmfit.jl
+++ b/src/glmfit.jl
@@ -49,11 +49,15 @@ function GlmResp(y::V, d::D, l::L, η::V, μ::V, off::V, wts::V) where {V<:FPVec
     return GlmResp{V,D,L}(y, d, similar(y), η, μ, off, wts, similar(y), similar(y))
 end
 
-function GlmResp(y::V, d::D, l::L, off::V, wts::V) where {V<:FPVector,D,L}
-    η   = similar(y)
-    μ   = similar(y)
-    r   = GlmResp(y, d, l, η, μ, off, wts)
-    initialeta!(r.eta, d, l, y, wts, off)
+function GlmResp(y::FPVector, d::Distribution, l::Link, off::FPVector, wts::FPVector)
+    # Instead of convert(Vector{Float64}, y) to be more ForwardDiff friendly
+    _y   = convert(Vector{float(eltype(y))}, y)
+    _off = convert(Vector{float(eltype(off))}, off)
+    _wts = convert(Vector{float(eltype(wts))}, wts)
+    η    = similar(_y)
+    μ    = similar(_y)
+    r    = GlmResp(_y, d, l, η, μ, _off, _wts)
+    initialeta!(r.eta, d, l, _y, _wts, _off)
     updateμ!(r, r.eta)
     return r
 end
@@ -465,14 +469,14 @@ Fit a generalized linear model to data.
 $FIT_GLM_DOC
 """
 function fit(::Type{M},
-    X::Union{Matrix{T},SparseMatrixCSC{T}},
+    X::AbstractMatrix{<:FP},
     y::AbstractVector{<:Real},
     d::UnivariateDistribution,
     l::Link = canonicallink(d);
     dofit::Bool = true,
     wts::AbstractVector{<:Real}      = similar(y, 0),
     offset::AbstractVector{<:Real}   = similar(y, 0),
-    fitargs...) where {M<:AbstractGLM,T<:FP}
+    fitargs...) where {M<:AbstractGLM}
 
     # Check that X and y have the same number of observations
     if size(X, 1) != size(y, 1)
@@ -485,7 +489,7 @@ function fit(::Type{M},
 end
 
 fit(::Type{M},
-    X::Union{Matrix,SparseMatrixCSC},
+    X::AbstractMatrix,
     y::AbstractVector,
     d::UnivariateDistribution,
     l::Link=canonicallink(d); kwargs...) where {M<:AbstractGLM} =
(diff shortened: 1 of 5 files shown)


src/glmfit.jl
```diff
@@ -465,7 +469,7 @@ Fit a generalized linear model to data.
 $FIT_GLM_DOC
 """
 function fit(::Type{M},
-    X::Union{Matrix{T},SparseMatrixCSC{T}},
+    X::AbstractMatrix,
```
bkamins, line 472: below (line 496) we convert `X` with `float`, but here we do not do it. why?

bkamins:
The CI failures indicate that we use different code paths for actual computations with and without views (the differences are roudoff errors). This is surprising. What is the reason?

bkamins:
In general I am OK to close my PR and finalize this. Thank you!

andreasnoack:
> This is surprising. What is the reason?

I agree. I expected them to take the same code path and therefore be exactly equal. I'll try figure out what is going on.

bkamins:
@andreasnoack - what is still needed in this PR to merge it?

bkamins:
There is some strange test failure on Julia 1.0 (I would assume that the failing `==` test should pass). @andreasnoack Do you know the reason?

src/glmfit.jl
```diff
@@ -49,11 +49,15 @@ function GlmResp(y::V, d::D, l::L, η::V, μ::V, off::V, wts::V) where {V<:FPVec
...
-function GlmResp(y::V, d::D, l::L, off::V, wts::V) where {V<:FPVector,D,L}
-    η   = similar(y)
-    μ   = similar(y)
-    r   = GlmResp(y, d, l, η, μ, off, wts)
-    initialeta!(r.eta, d, l, y, wts, off)
+function GlmResp(y::FPVector, d::Distribution, l::Link, off::FPVector, wts::FPVector)
+    # Instead of convert(Vector{Float64}, y) to be more ForwardDiff friendly
+    _y   = float(convert(Vector, y))
+    _off = float(convert(Vector, off))
+    _wts = float(convert(Vector, wts))
```
nalimilan, line 56: Looks like I missed these:
```suggestion
    _y   = convert(Vector{float(eltype(y))}, y)
    _off = convert(Vector{float(eltype(off))}, off)
    _wts = convert(Vector{float(eltype(wts))}, wts)
```

nalimilan:
That failure is indeed weird. I wonder whether it happens at random or whether it's something specific to Julia 1.0 on Mac.

andreasnoack:
I can reproduce this locally with Julia 1.0
```
julia> coef(glm(X, y, Normal(), IdentityLink()))
2-element Array{Float64,1}:
 0.9999999999999999
 1.0

julia> coef(glm(view(X, 1:9, :), view(y, 1:9), Normal(), IdentityLink()))
2-element Array{Float64,1}:
 1.0
 1.0
```
I don't know why the two versions are taking a different code path somewhere. If somebody has any idea then I'd like to hear it.

andreasnoack:
This also fails on Julia 1.7 so I'll try to figure out where the divergence is happening.

bkamins:
There is the same problem for `lm` as for `glm`:
```
julia> coef(lm(X, y))
2-element Array{Float64,1}:
 1.0
 1.0000000000000004

julia> coef(lm(view(X, 1:9, :), view(y, 1:9)))
2-element Array{Float64,1}:
 0.9999999999999999
 1.0

julia> coef(glm(X, y, Normal(), IdentityLink()))
2-element Array{Float64,1}:
 0.9999999999999999
 0.9999999999999999

julia> coef(glm(view(X, 1:9, :), view(y, 1:9), Normal(), IdentityLink()))
2-element Array{Float64,1}:
 0.9999999999999999
 1.0
```

andreasnoack:
So this turned out to be pretty dumb. The view was `1:9` but he length was `10` so the two problems weren't identical. I just hadn't added any noise to the dependent variable so it was a perfect fit except for rounding differences. I've now added some noise and made the views the same lengths as the original data.

bkamins (approved):
Can you please also bump package version number in this PR and make a release after this?

Thank you!

nalimilan:
Good! Though it looks like my suggestions got reverted.

src/glmfit.jl
andreasnoack: The method at line 496 supposed to promote the arguments before this one is called. However, I can see that this one should then continue be restricted floats. I'll fix that.

andreasnoack:
> Though it looks like my suggestions got reverted.

Hm. Thanks for catching that. I tried to squash the intermediate commits while rebasing on master and I must somehow have ended up reverting those changes while doing that. I've reintroduced your suggestions.

andreasnoack:
I've now also included the tests from https://github.com/JuliaStats/GLM.jl/pull/466 and bumped the version.

src/linpred.jl
```diff
@@ -185,7 +185,7 @@ function SparsePredChol(X::SparseMatrixCSC{T}) where T
         similar(X))
 end
 
-cholpred(X::SparseMatrixCSC) = SparsePredChol(X)
+cholpred(X::SparseMatrixCSC, ::Bool=false) = SparsePredChol(X)
```
nalimilan, line 188:
```suggestion
cholpred(X::SparseMatrixCSC, pivot::Bool=false) = SparsePredChol(X)
```

nalimilan (approved):

Issue #441, opened 2021-07-02
License terms incomplete?


rcurtin:
Hi there!  Thank you for this excellent package.  I had a question about licensing.

The license in the repository is the MIT license, but due to GLM's dependency on Distributions, which in turn depends on StatsFuns, there is a dependency on the GPL-ed Rmath library via [Rmath.jl](https://github.com/JuliaStats/Rmath.jl).  But, that isn't noted here.  Hence my surprise when I realized that building and distributing something on top of GLM.jl would in turn mean that whatever I distributed would need to be under the terms of the GPL. :)

Is it possible to perhaps remove the Rmath dependency and use a different dependency for what that package is currently being used for?  (Please note, I am ignorant of whether that is easy and will take five minutes, or is a huge amount of work... just throwing it out there. :))  Or, at least maybe it is worth updating the license here to match something a little more like [FFTW.jl's](https://github.com/JuliaMath/FFTW.jl#license), or use a [THIRDPARTY.md file](https://github.com/JuliaLang/julia/blob/master/THIRDPARTY.md)?

In any case, this presents a bit of a blocker for my usage of GLM.jl, and so I'd really appreciate any input towards what a path to resolution could be.

Thanks!

rcurtin:
(see also https://github.com/JuliaStats/Rmath.jl/issues/75)

nalimilan:
As noted at https://github.com/JuliaStats/Rmath.jl/issues/75#issuecomment-912984117:
> I'd say it would be reasonable to make PRs to packages that depend on Rmath to add a THIRDPARTY.md file and a reference to it in LICENSE.md. Though of course it will be very hard to be exhaustive given that these are quite fundamental packages that are used indirectly by many packages (and new ones are added all the time).

rcurtin:
See #447. :+1:

PR #447, opened 2021-09-08
Update license to indicate GPL dependency.

rcurtin:
This comes from #441.  Happy to update the language---primarily I just wanted to indicate that there is indeed a GPL dependency here for now. :+1:

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -22,3 +22,8 @@ The GLM module is licensed under the MIT License:
 > LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
 > OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
 > WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+Note that the GLM module has an implicit dependency on the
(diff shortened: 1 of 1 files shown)


LICENSE.md
```diff
@@ -22,3 +22,8 @@ The GLM module is licensed under the MIT License:
 > LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
 > OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
 > WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+Note that the GLM module has an implicit dependency on the
```
nalimilan, line 26: Not sure what "implicit" means here:
```suggestion
Note that the GLM module depends on the
```

rcurtin: Hm, yeah, maybe there is a better word.  GLM.jl doesn't depend *directly* on Rmath, the dependency is via Distributions via StatsFuns.  Anyway, if you prefer I can take your suggestion, just let me know what you think now that I've clarified what I was going for. :)

nalimilan: Let's say indirect then. :-)
```suggestion
Note that the GLM module has an indirect dependency on the
```

nalimilan:
Thanks!

PR #451, opened 2021-11-02
Add `ftest` to the manual

nalimilan:
We should also mention it in the examples and not just in the API reference, but for now it's better than nothing.

Cc: @OndrejSlamecka

### DIFF
--- a/docs/src/api.md
+++ b/docs/src/api.md
@@ -63,6 +63,7 @@ fit
 ```@docs
 StatsBase.deviance
 GLM.dispersion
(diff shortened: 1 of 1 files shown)


bkamins (approved):

nalimilan:
Actually the manual isn't deployed currently (https://github.com/JuliaStats/GLM.jl/pull/452). And the block to which I added `ftest` is just a list of function names, without even expanding docstrings. The manual really needs some love!

PR #452, opened 2021-11-03
Fix manual deployment

nalimilan:
Currently deployment doesn't happen because some conditions related to Travis are not fulfilled. Hopefully using a newer version of the action will fix it.

### DIFF
--- a/.github/workflows/Documenter.yml
+++ b/.github/workflows/Documenter.yml
@@ -12,8 +12,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


bkamins (approved):

bkamins:
OK - I guess you know hot to test if it can be safely merged :)

nalimilan:
Looks like it worked now.

andreasnoack:
Thanks for fixing this. I've pushed a `v1.5.1+doc1` tag to trigger a docs deployment for version 1.5.1 and they are now up https://juliastats.org/GLM.jl/stable/.

PR #454, opened 2021-11-11
Docs for not using a DataFrame

KronosTheLate:
This PR adds documentation on how to do a fit without using a DataFrame, as suggested by @pdeffebach in [this thread](https://discourse.julialang.org/t/the-simplest-linear-fit-with-glm/71316/11).

### DIFF
--- a/docs/src/examples.md
+++ b/docs/src/examples.md
@@ -44,6 +44,42 @@ julia> round.(predict(ols), digits=5)
  6.83333
 ```
 
+### Without DataFrame
+Because a named tuple follows common table-interface defined in `Tables.jl` 
+(which is the one followed by a DataFrame), the problem can also be specified 
+with data in a named tuple as opposed to a `DataFrame`:
+```jldoctetst
+julia> using GLM
+
+julia> X=[1,2,3]
+3-element Vector{Int64}:
+ 1
+ 2
+ 3
+
+julia> Y=[2,4,7]
+3-element Vector{Int64}:
+ 2
+ 4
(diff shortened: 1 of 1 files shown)


nalimilan:
Thanks. Though rather than adding examples to a section which is currently organized by GLM type, it would be nice to explain in the "Home" section (e.g. in "Fitting GLM models") the different ways one can pass data, with a simple example. Currently that section is really lacking.

KronosTheLate:
Thanks yourself, for the package ^_^

When you say "Organized by GLM type", you mean with a dataframe as input?
In that case I agree, that a new section with the examples with data as arrays would be better.

I will have a look at what part of this I think would fit better in the "Home" section.

nalimilan:
> When you say "Organized by GLM type", you mean with a dataframe as input?

I mean that currently we have one example for each kind of GLM family (linear model, negative binomial...). Each of these can be fitted by passing a data frame, a named tuple, matrices, etc. Since these are orthogonal, better document them in different places to avoid duplicating all families for all input data types.

KronosTheLate:
Right, that makes sense.

Alternativly, a page on how to read, understand and write a `@formula` expression would be very helpful I think, because understanding it well is so key to being able to express the problem inside the GLM package. So perhaps a page dedicated to showing different `@formula`'s and the equivalent matrices would be good? And then, if the translation from `@formula` to matrix is well established, then the current documentation would be sufficient.

nalimilan:
Yes we really need more complete docs.

PR #340, opened 2019-10-28
Add support for rank deficient `GeneralizedLinearModel`

jason-xuan:
This pull request supersedes #314 because I don't have the write access to either REPOs.

In #314 an issue about slower converge has been discussed. This PR solves that issue and adds a test case about that issue.

cc: @Nosferican @nalimilan @jiahao @dmbates @andreasnoack @DilumAluthge

PR #458, opened 2021-11-26
Fix `coeftable` for saturated linear models

### LINKED ISSUE #456: Linear models failing to display with ambiguous datatypes with unhelpful error messages
The linear model is failing somewhere in `show` (I guess) if you give it `DataFrame`s with type `Any`.  MWE:
```julia
julia> df_test = DataFrame(date_numeric = [737810, 737841, 737869, 737900, 737930, 737961, 737991, 738022, 738053, 738083, 738114], num = Any[-74587.93, -74550.49, -74482.09, -74441.45, -74316.17, -74252.81, -73976.21, -73587.65, -73170.53, -72753.41, -72304.01])
11×2 DataFrame
 Row │ date_numeric  num
     │ Int64         Any
─────┼────────────────────────────
   1 │       737810  -74587.9
   2 │       737841  -74550.5
   3 │       737869  -74482.1
   4 │       737900  -74441.4
   5 │       737930  -74316.2
   6 │       737961  -74252.8
   7 │       737991  -73976.2
   8 │       738022  -73587.6
   9 │       738053  -73170.5
  10 │       738083  -72753.4
  11 │       738114  -72304.0

julia> model = fit(LinearModel, @formula(date_numeric ~ num), df_test)
ERROR: ArgumentError: FDist: the condition ν1 > zero(ν1) && ν2 > zero(ν2) is not satisfied.
Stacktrace:
  # ...

julia> df_test = DataFrame(date_numeric = [737810, 737841, 737869, 737900, 737930, 737961, 737991, 738022, 738053, 738083, 738114], num = Float64[-74587.93, -74550.49, -74482.09, -74441.45, -74316.17, -74252.81, -73976.21, -73587.65, -73170.53, -72753.41, -72304.01]);

julia> model = fit(LinearModel, @formula(date_numeric ~ num), df_test)
StatsModels.TableRegressionModel{LinearModel{GLM.LmResp{Vector{Float64}}, GLM.DensePredChol{Float64, LinearAlgebra.CholeskyPivoted
{Float64, Matrix{Float64}}}}, Matrix{Float64}}

date_numeric ~ 1 + num

Coefficients:
─────────────────────────────────────────────────────────────────────────────────────
                      Coef.    Std. Error       t  Pr(>|t|)  Lower 95%      Upper 95%
─────────────────────────────────────────────────────────────────────────────────────
(Intercept)   746741.0       1103.35       676.79    <1e-21  7.44245e5  749237.0
num                0.118876     0.0149383    7.96    <1e-04  0.0850827       0.152668
─────────────────────────────────────────────────────────────────────────────────────
```

Took me quite a while to figure out what exactly was happening here (realised something weird was happening when running `model = fit(LinearModel, @formula(date_numeric ~ num), df_test);` in the REPL worked, but didn't work when I didn't suppress the display (i.e., when I removed the semicolon).

nalimilan:
`coeftable` failed for saturated `LinearModel`s due to trying to compute F and T distributions with zero DOF.

Fixes #456.

### DIFF
--- a/src/lm.jl
+++ b/src/lm.jl
@@ -215,10 +215,18 @@ end
 
 function coeftable(mm::LinearModel; level::Real=0.95)
     cc = coef(mm)
-    se = stderror(mm)
-    tt = cc ./ se
-    p = ccdf.(Ref(FDist(1, dof_residual(mm))), abs2.(tt))
-    ci = se*quantile(TDist(dof_residual(mm)), (1-level)/2)
+    dofr = dof_residual(mm)
+    if dofr > 0
+        se = stderror(mm)
+        tt = cc ./ se
+        p = ccdf.(Ref(FDist(1, dof_residual(mm))), abs2.(tt))
+        ci = se*quantile(TDist(dof_residual(mm)), (1-level)/2)
+    else
+        se = fill(NaN, length(cc))
+        tt = fill(NaN, length(cc))
+        p = fill(NaN, length(cc))
+        ci = fill(NaN, length(cc))
+    end
     levstr = isinteger(level*100) ? string(Integer(level*100)) : string(level*100)
     CoefTable(hcat(cc,se,tt,p,cc+ci,cc-ci),
               ["Coef.","Std. Error","t","Pr(>|t|)","Lower $levstr%","Upper $levstr%"],
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -141,6 +141,37 @@ end
     @test isapprox(coef(m2p_dep_pos_kw), coef(m2p))
 end
 
+@testset "saturated linear model" begin
+    df = DataFrame(x=["a", "b", "c"], y=[1, 2, 3])
+    model = lm(@formula(y ~ x), df)
+    ct = coeftable(model)
+    @test dof_residual(model) == 0
+    @test dof(model) == 4
+    @test coef(model) ≈ [1, 1, 2]
+    @test all(isnan, hcat(ct.cols[2:end]...))
+
+    model = lm(@formula(y ~ 0 + x), df)
+    ct = coeftable(model)
+    @test dof_residual(model) == 0
+    @test dof(model) == 4
+    @test coef(model) ≈ [1, 2, 3]
+    @test all(isnan, hcat(ct.cols[2:end]...))
+
+    model = glm(@formula(y ~ x), df, Normal(), IdentityLink())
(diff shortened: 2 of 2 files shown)


andreasnoack:
I don't think the NaNs are right here. I guess the standard errors and t-values are already correct, i.e. Inf and zero respectively. I believe the CI should also just be `[-Inf,Inf]` and the P-value should be one.

nalimilan:
Actually that was my first thought, but I changed to NaN when I noticed that GLMs used that already and that R did the same. But I agree it sounds mathematically more correct to use Inf, 1.0 and 0.0. This changes the behavior of `coeftable` for GLMs though (for LMs an error was thrown so it's OK), but I guess we can consider this as non-breaking.

palday (approved):

palday:
@nalimilan I think the infinite CIs make sense here for a saturated model, but when the model becomes "oversaturated" i.e. rank deficient, then we return NaNs for the test statistics, right? I guess that's still coherent -- in the pivoted case, the coefficients are set to zero and there's not really values for the associated errors, i.e. the errors are Not A Number. In the saturated case, it's just that the uncertainty is infinite.

palday:
@nalimilan make a patch bump and then we can immediately tag a release :smile:

nalimilan:
Good point. For rank-deficient models, the PR used NaN for the t-value but not for the p-value and the CI. I've pushed a commit to fix this, with a test for `lm`. `glm` doesn't support rank-deficient models yet so I've left a comment to remember testing this as part of https://github.com/JuliaStats/GLM.jl/pull/340. BTW, note that with the PR `dispersion` will `Inf` for saturated rank-deficient models once we can fit them. This is better than returning `NaN`, right?

I've also bumped the minor version. That seems appropriate given that this PR changes the behavior a bit in use cases that are already supported in the current release, so this isn't just a small bugfix.

palday:
@nalimilan I agree with the minor instead of patch release.

Maybe @dmbates can comment on the origin of this convention and whether everything still makes sense? 

The one advantage for `p=1.0` that I can see from a naive perspective is that logical comparisons work as expected. For example, `NaN > 0.05` is false. But then again if a user is paying that little attention, then I guess there's nothing we can do to help them.

nalimilan:
@palday Let's merge?

nalimilan:
https://github.com/JuliaRegistries/General/pull/58360

PR #460, opened 2021-12-15
Update doctests for pivot

palday:
The doctests should have `CholeskyPivoted` and not `Cholesky` as the type parameter for the regression models because we default to rank reduction / pivoting in `[g]lm` calls. I also updated the display of `DataFrame`s for current DataFrames.jl.


docs/Project.toml
```diff
@@ -8,4 +8,5 @@ StatsBase = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
 StatsModels = "3eaba693-59b7-5ba5-a881-562e759f1c8d"
 
 [compat]
+DataFrames = "~1.3"
```
palday, line 11: Lock in the DataFrames version in docs so that the pretty printing remains the same.

nalimilan: I'm not sure I like this. Better get failures if printing changes so that we can keep the docs up to date? Otherwise we'll likely forget about that.

palday: I agree with your philosophy but we hadn't been failing on doctests even though that behavior changed a while back.  I don't mind getting rid of the compat restriction, but we need to be better about keeping the doctests up to date.

nalimilan: Indeed. We should pass `strict=true`, right?

palday: Uhh.... yes. I'll update that too.

palday: @nalimilan done

src/glmtools.jl
```diff
@@ -165,19 +166,21 @@ Return a 3-tuple of the inverse link, the derivative of the inverse link, and wh
 The variance function is returned as NaN unless the range of μ is (0, 1)
 
 # Examples
-```jldoctest
-julia> inverselink(LogitLink(), 0.0)
-(0.5, 0.25, 0.25)
+```jldoctest; setup = :(using GLM)
```
nalimilan, line 169: I wonder whether adding this in the manual wouldn't avoid the need to repeat this everywhere:
```julia

```@meta
CurrentModule = GLM
`` `

```

nalimilan (approved):

src/glmtools.jl
palday: Setting the current module that way doesn't seem to work for quick route of fixing outdated doctests:

```julia
julia> using Documenter, GLM
[ Info: Precompiling GLM [38e38edf-8417-5370-95a0-9cbb8c7f171a]

julia> doctest(GLM; fix=true)
```

But it's something we can revisit at some point.

nalimilan: Too bad...

PR #462, opened 2022-01-05
use `StatsModels.isnested` instead of defining new `issubmodel` func

kleinschmidt:
This just replaces the function called `issubmodel` with a method for `StatsModels.isnested` in order to provide support for `StatsModels.lrtest` and other goodies that assume that API.  `issubmodel` is deprecated and otherwise there are no changes to the code.

edit: I have not bumped the version.  I'll open a new PR for that once this is merged.

### DIFF
--- a/docs/src/api.md
+++ b/docs/src/api.md
@@ -65,7 +65,7 @@ StatsBase.deviance
 GLM.dispersion
 GLM.ftest
 GLM.installbeta!
-GLM.issubmodel
+StatsModels.isnested
 StatsBase.nobs
 StatsBase.nulldeviance
 StatsBase.predict
--- a/docs/src/api.md
+++ b/docs/src/api.md
@@ -65,10 +65,10 @@ StatsBase.deviance
 GLM.dispersion
 GLM.ftest
 GLM.installbeta!
-StatsModels.isnested
 StatsBase.nobs
 StatsBase.nulldeviance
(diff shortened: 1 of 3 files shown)


docs/src/api.md
```diff
@@ -65,7 +65,7 @@ StatsBase.deviance
 GLM.dispersion
 GLM.ftest
 GLM.installbeta!
-GLM.issubmodel
+StatsModels.isnested
```
palday, line 68: should the position be changed to maintained alphabetical (when including the module name) order?

src/ftest.jl
```diff
@@ -14,8 +14,10 @@ mutable struct FTestResult{N}
     pval::NTuple{N, Float64}
 end
 
+@deprecate issubmodel(mod1::LinPredModel, mod2::LinPredModel; atol::Real=0.0) StatsModels.isnested(mod1, mod2; atol=atol)
+
 """A helper function to determine if mod1 is nested in mod2"""
```
palday, line 19: should this docstring go and just lean on the `StatsModels` stub docstring?

palday (approved):

src/ftest.jl
kleinschmidt: the only reason not to do that would be if this explaind the specific algorithm that's used here but it doesn't so we can add that later if needed.

PR #466, opened 2022-01-20
support heterogeneous array types

### LINKED ISSUE #213: `GLM.lm(X, y)` fails when `X` is a `SubArray`
This works:
```julia
julia> xs = randn(46, 3)
46×3 Array{Float64,2}:
julia> ys = randn(46)
46-element Array{Float64,1}:
 julia> GLM.lm(xs, ys)
GLM.LinearModel{GLM.LmResp{Array{Float64,1}},GLM.DensePredChol{Float64,Base.LinAlg.Cholesky{Float64,Array{Float64,2}}}}:
```

This doesn't:
```julia
julia> GLM.lm(@view(xs[1:end, 1:end]), ys)
ERROR: MethodError: no method matching GLM.DensePredChol(::SubArray{Float64,2,Array{Float64,2},Tuple{UnitRange{Int64},UnitRange{Int64}},false}, ::Array{Float64,1}, ::Array{Float64,1}, ::Array{Float64,1}, ::Base.LinAlg.Cholesky{Float64,Array{Float64,2}}, ::Array{Float64,2}, ::Array{Float64,2})
Closest candidates are:
  GLM.DensePredChol(::Array{T<:Union{Float32, Float64},2}, ::Array{T<:Union{Float32, Float64},1}, ::Array{T<:Union{Float32, Float64},1}, ::Array{T<:Union{Float32, Float64},1}, ::C, ::Array{T<:Union{Float32, Float64},2}, ::Array{T<:Union{Float32, Float64},2}) where {T<:Union{Float32, Float64}, C} at /.julia/v0.6/GLM/src/linpred.jl:91

Stacktrace:
 [1] GLM.DensePredChol(::SubArray{Float64,2,Array{Float64,2},Tuple{UnitRange{Int64},UnitRange{Int64}},false}, ::Bool) at /.julia/v0.6/GLM/src/linpred.jl:103
 [2] lm(::SubArray{Float64,2,Array{Float64,2},Tuple{UnitRange{Int64},UnitRange{Int64}},false}, ::Array{Float64,1}) at /.julia/v0.6/GLM/src/lm.jl:150
```

I understand the view on the array should be equivalent, so it’s probably an improvement to be made to the GLM package?

piever:
Fix #213 

This simply performs a conversion step before building `GLMResp` or `LMResp` to make sure that all arrays have the same type before proceeding to construct the object.

It still does not allow to pass an abstract matrix as the first argument to `GLM.glm`, because of [this signature](https://github.com/JuliaStats/GLM.jl/blob/be96833e470b914cd68ba007b3e47d127357be54/src/glmfit.jl#L468), but I imagine that can be addressed separately.

### DIFF
--- a/src/glmfit.jl
+++ b/src/glmfit.jl
@@ -49,11 +49,12 @@ function GlmResp(y::V, d::D, l::L, η::V, μ::V, off::V, wts::V) where {V<:FPVec
     return GlmResp{V,D,L}(y, d, similar(y), η, μ, off, wts, similar(y), similar(y))
 end
 
-function GlmResp(y::V, d::D, l::L, off::V, wts::V) where {V<:FPVector,D,L}
+function GlmResp(y::FPVector, d::D, l::L, off::FPVector, wts::FPVector) where {D,L}
     η   = similar(y)
     μ   = similar(y)
-    r   = GlmResp(y, d, l, η, μ, off, wts)
-    initialeta!(r.eta, d, l, y, wts, off)
+    V   = typeof(η)
+    r   = GlmResp(convert(V, y), d, l, η, μ, convert(V, off), convert(V, wts))
+    initialeta!(r.eta, d, l, r.y, r.wts, r.offset)
     updateμ!(r, r.eta)
     return r
 end
--- a/src/lm.jl
+++ b/src/lm.jl
@@ -28,8 +28,12 @@ mutable struct LmResp{V<:FPVector} <: ModResp  # response in a linear model
     end
 end
 
-LmResp(y::FPVector, wts::FPVector=similar(y, 0)) = 
-    LmResp{typeof(y)}(fill!(similar(y), 0), similar(y, 0), wts, y)
+function LmResp(y::FPVector, wts::FPVector=similar(y, 0))
+    μ   = fill!(similar(y), 0)
+    off = similar(y, 0)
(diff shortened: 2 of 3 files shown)


kleinschmidt:
Looks like you also made the change to support `AbstractMatrix` in `glm`?

kleinschmidt (changes requested):
This looks great! Thank you for the tests!!  Assuming CI is happy, can you bump the patch (this is not breaking AFAICT) and we should be good to go.

piever:
Thanks for the quick review!

> Looks like you also made the change to support `AbstractMatrix` in `glm`?

Yes, in the end it was a very small change, so it felt cleaner to include it here.

I've bumped the patch number, it would actually be great to tag a release after merging this (so I can remove the workarounds in AlgebraOfGraphics).

kleinschmidt:
I think it's worth looking at how this interacts with #443 and #446 so I've requested review from @andreasnoack as well.

andreasnoack:
So https://github.com/JuliaStats/GLM.jl/pull/446 is slightly more general since it also covers a fix for sparse X which I don't think is covered by the changes here so I'll suggest that we go with that PR. I've taken the extra tests from this PR since they were more extensive than what had written.

piever:
Sure! Closing this as it was superseded by #446. Glad the tests were useful!

PR #467, opened 2022-01-24
fix predict docstring and remove trailing dim

piever:
This PR fixes the docstring of `predict(::LinearModel, x; interval, level)`, which claimed to return a matrix with 3 columns, whereas it is actually a named tuple with three fields (`prediction`, `lower`, `upper`).

It also fixes the types of the returned `lower` and `upper`. Previously, `prediction` was a vector (IMO, the correct shape), whereas `lower` and `upper` were `N x 1` matrices (IMO, the incorrect shape). Now, all three are vectors.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "GLM"
 uuid = "38e38edf-8417-5370-95a0-9cbb8c7f171a"
-version = "1.6.2"
+version = "1.7.0"
 
 [deps]
 Distributions = "31c24e10-a181-5473-b8eb-7969acd0382f"
--- a/src/lm.jl
+++ b/src/lm.jl
@@ -236,7 +236,7 @@ end
 
 If `interval` is `nothing` (the default), return a vector with the predicted values
 for model `mm` and new data `newx`.
-Otherwise, return a 3-column matrix with the prediction and
+Otherwise, return a vector with the predicted values as well as
 the lower and upper confidence bounds for a given `level` (0.95 equates alpha = 0.05).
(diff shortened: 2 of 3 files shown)


nalimilan:
Thanks. The docstring fix is fine, but I'm afraid changing the return type would be breaking so we should wait until 2.0.

piever:
> I'm afraid changing the return type would be breaking so we should wait until 2.0.

I see your point, but I am not sure it is so clear cut. Is there a way that we could verify whether this change breaks any downstream package?

Even though it is possible, it seems unlikely to me that changing from `N x 1` matrix to vector would break code in practice. If we can test that this does not cause breakage in packages that depend on GLM, could it be done for the 1.7 release?

Otherwise, it seems overly laborious to change now the docstring to document the (IMO incorrect) return type, add a milestone to not forget about this, and then come back to it and update code and docstring.

Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "GLM"
 uuid = "38e38edf-8417-5370-95a0-9cbb8c7f171a"
-version = "1.6.1"
+version = "1.6.2"
```
nalimilan, line 3:
```suggestion
version = "1.7.0"
```

nalimilan (approved):

nalimilan:
Let's merge this and https://github.com/JuliaStats/GLM.jl/pull/458 and then tag 1.7.0.

piever:
Thanks for accepting the change! I've rebased and force-pushed to fix the conflicts (on the Project.toml file).

PR #477, opened 2022-04-17
Do not mention pseudo-R² in manual

nalimilan:
`r2` does not work for GLMs currently (even with two arguments) as they don't implement `nullloglikelihood`.
Also add a mention about `adjr2`.

If somebody has an idea regarding how `nullloglikelihood` could be implemented... Maybe all we need to do is define for each link what the prediction would be for all observations under the null model, and call `loglik_obs` on that like we do for `loglikelihood`?

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -152,7 +152,8 @@ Many of the methods provided by this package have names similar to those in [R](
 - `dof_residual`: degrees of freedom for residuals, when meaningful
 - `glm`: fit a generalized linear model (an alias for `fit(GeneralizedLinearModel, ...)`)
 - `lm`: fit a linear model (an alias for `fit(LinearModel, ...)`)
-- `r2`: R² of a linear model or pseudo-R² of a generalized linear model
+- `r2`: R² of a linear model
+- `adjr2`: adjusted R² of a linear model
 - `stderror`: standard errors of the coefficients
 - `vcov`: estimated variance-covariance matrix of the coefficient estimates
 - `predict` : obtain predicted values of the dependent variable from the fitted model


palday:
@nalimilan The null model is "just" `y ~ 1`, so it would be trivial to create a null model using the response of the original and a column of 1's as `X`.

Or am I missing something ?

nalimilan:
Yeah, I just wonder whether we should compute this more efficiently. After looking into it, I have a commit here that computes the null deviance and log-likelihood correctly simply by defining the null model as taking the predicted response to be the mean response. This works for all models we currently test, except those with offsets. Do you think there exist a direct formula for this case too?

palday:
@nalimilan did you push the commit? For Poisson at least I think there might be a few easy/fast cases even with offset. I'll have to work through the algebra later to be sure there.

nalimilan:
I've just filed https://github.com/JuliaStats/GLM.jl/pull/479. Let me know if you can find the formulas. Though if some cases have no closed-form solution we'll have to rely on fitting the null model at least for them...

FWIW, R's `?glm` has this warning about offsets:
> null.deviance: The deviance for the null model, comparable with
> ‘deviance’. The null model will include the offset, and an
> intercept if there is one in the model.  Note that this will
> be incorrect if the link function depends on the data other
> than through the fitted mean: specify a zero offset to force
> a correct calculation.

palday (approved):
approving for now, feel free to merge (or not) until we get #479 done

ParadaCarleton:
Does `r^2` have to return a pseudo-r2? I ask because there are many definitions of pseudo-r2, so it can be confusing. This could also cause mistakes for users expecting the function to return only the actual `r^2`. (I would assume this myself.)

nalimilan:
`r2` only returns the pseudo-R² if you pass a second argument specifying which variant you want.

nalimilan:
Superseded by https://github.com/JuliaStats/GLM.jl/pull/479.

PR #479, opened 2022-04-18
Implement `nulldeviance` and `nullloglikelihood` for GLMs

nalimilan:
Models with offets are not supported for now.


src/glmfit.jl
```diff
@@ -248,6 +248,30 @@ end
...
+    isempty(r.offset) ||
+        throw(ErrorException("models with offsets are currently not supported"))
+
+    dev = zero(eltype(y))
+    if length(wts) == length(y)
+        mu = mean(y, weights(wts))
+        @inbounds for i in eachindex(y, wts)
+            dev += wts[i] * devresid(d, y[i], mu)
+        end
+    else
```
palday, line 266: `wts` should either be empty of same length as `y`, right?

palday:
From McCullagh and Nelder: 
image

nalimilan:
What am I supposed to conclude from this excerpt? :-D AFAICT the call to `devresid` is equivalent to these formulas, right? But I'm still not sure how to handle offsets.

src/glmfit.jl
nalimilan: Yeah, `isempty(wts)` would probably be clearer, and more robust since we would get an error anyway if lengths don't match. I just copied the pattern used in `loglikelihood` but I'll change all functions to use `isempty`.

palday:
Yeah, the excerpt was also just for my own note-taking as the closest thing I found to an efficient closed-form formula for the null deviance. 

I think offsets are just always awkward to handle. The offset is added to mu, right? Then we have mean(mu+offset) = mean(mu) + mean(offset) and can just use that in all our calculations, which will give the same problems you mentioned in R (https://github.com/JuliaStats/GLM.jl/pull/477#issuecomment-1101745178).

For models without an intercept, the null model is actually the same as forcing the mean to zero, right? I'm not sure if that will work numerically for some distributions, but maybe we should warn/error in those cases.

nalimilan:
> I think offsets are just always awkward to handle. The offset is added to mu, right? Then we have mean(mu+offset) = mean(mu) + mean(offset) and can just use that in all our calculations, which will give the same problems you mentioned in R ([#477 (comment)](https://github.com/JuliaStats/GLM.jl/pull/477#issuecomment-1101745178)).

Not exactly, but close: the offset is included in `eta`, i.e. `eta = XB + offset`.

After some more investigations, I've made some progress: I'm able to compute the null deviance for a Poisson model with a log link and an offset. I'm not sure why it doesn't work for a Normal model with a log link and an offset. I'll see whether I can find a solution, and if not I will at least push what I have.

> For models without an intercept, the null model is actually the same as forcing the mean to zero, right? I'm not sure if that will work numerically for some distributions, but maybe we should warn/error in those cases.

I wondered the same thing. The null model is defined as the one with only the intercept in the docstring currently. For models without an intercept, it could make sense to define it as the model for which all predicted values are 0. Though these things [are confusing](https://stats.stackexchange.com/questions/26176/removal-of-statistically-significant-intercept-term-increases-r2-in-linear-mo/26205#26205), so throwing an error would be safer, at least as long as users don't request that feature. This matters especially because people may call `r2` for GLMs without realizing what null model is taken as the reference.

src/glmfit.jl
```diff
@@ -269,6 +308,41 @@ function loglikelihood(m::AbstractGLM)
...
+            ϕ = nulldeviance(m)/length(y)
+            @inbounds for i in eachindex(y)
+                ll += loglik_obs(d, y[i], mu, 1, ϕ)
+            end
+        end
+    else
+        nullm = fit(GeneralizedLinearModel,
+                    fill(1.0, length(y), 1), y, d, L(), wts=wts, offset=offset,
+                    maxiter=m.maxiter, minstepfac=m.minstepfac,
+                    atol=m.atol, rtol=m.rtol)
```
nalimilan, line 340: I'm not sure whether this is really needed, as the null model should converge much more quickly than more complex models. Maybe we don't need to use a low tolerance just to compute the deviance.

Note that this requires storing these values in new `GeneralizedLinearModel` fields (which could be useful in its own right).

nalimilan:
I had code that computes the null deviance/log-likelihood for Poisson models with offsets, but I'm afraid it loses precision significantly in some cases due to the back and forth between operations on the response scale and operations on the linear predictor scale. For a Binomial model with large offsets, this gave an infinite deviance. Given that we need a fallback that does fit the null model for other distributions anyway, I figured it would be simpler to always use it when there are offsets. I've kept the intermediate commits if you want to have a look.

nalimilan:
Another issue is that for models fitted with `negbin`, the null model is taken to be the one with the same theta parameter, which will have a worse fit than calling `negbin` again. Maybe `negbin` should fill a field in the model object to indicate that theta was chosen automatically so that we throw an error from `nulldeviance` and `nullloglikelihood` instead of returning misleading results? (We can't always throw an error for `NegativeBinomial` as users can specify theta manually.)

palday (approved):

src/glmfit.jl
```diff
@@ -248,14 +255,46 @@ end
...
+                            "null deviance cannot be interpreted"))
+    r      = m.rr
+    wts    = weights(r.wts)
+    y      = r.y
+    d      = r.d
+    offset = r.offset
+    dev = zero(eltype(y))
+    if isempty(offset) # Faster method
+        if !isempty(wts)
+            mu = mean(y, weights(wts))
```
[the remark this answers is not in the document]
nalimilan: Good catch!

nalimilan:
Given that we now have a clear definition the null model when theere's no intercept in StatsAPI (https://github.com/JuliaStats/StatsAPI.jl/pull/14) and an implementation for linear models (https://github.com/JuliaStats/GLM.jl/pull/481), I've pushed changes to support the no intercept case.

src/glmfit.jl
```diff
@@ -257,31 +257,32 @@ end
...
     offset = r.offset
-    dev = zero(eltype(y))
+    hasint = hasintercept(m)
+    dev    = zero(eltype(y))
     if isempty(offset) # Faster method
         if !isempty(wts)
-            mu = mean(y, weights(wts))
+            mu = hasint ?
+                mean(y, wts) :
+                linkinv(r.link, zero(eltype(y))*zero(eltype(wts))/1)
```
palday, line 272:
```suggestion
                linkinv(r.link, zero(eltype(y)) * zero(eltype(wts)) / 1)
```

This is `0 * 0 / 1`, isn't it?

src/glmfit.jl
```diff
@@ -257,31 +257,32 @@ end
...
             @inbounds for i in eachindex(y, wts)
                 dev += wts[i] * devresid(d, y[i], mu)
             end
         else
-            mu = mean(y)
+            mu = hasint ? mean(y) : linkinv(r.link, zero(eltype(y))/1)
```
palday, line 277: This is another `0/1`. Why the division by 1?

src/glmfit.jl
```diff
@@ -257,31 +257,32 @@ end
...
             @inbounds for i in eachindex(y)
                 dev += devresid(d, y[i], mu)
             end
         end
     else
+        X = fill(1.0, length(y), hasint ? 1 : 0)
```
palday, line 283:
```suggestion
        X = fill(1.0, length(y), hasint)
```

src/glmfit.jl
```diff
@@ -310,34 +311,32 @@ function loglikelihood(m::AbstractGLM)
...
     wts    = r.wts
     y      = r.y
     d      = r.d
     offset = r.offset
+    hasint = hasintercept(m)
     ll  = zero(eltype(y))
     if isempty(r.offset) # Faster method
         if !isempty(wts)
-            mu = mean(r.y, weights(wts))
+            mu = hasint ? mean(y, weights(wts)) : linkinv(r.link, zero(ll)/1)
```
palday, line 324: I feel like I'm missing something obvious here -- why dies `0 / 1` keep popping up?

src/glmfit.jl
ararslan: Yes but of the appropriate type, whereas `0 * 0 / 1` will always be a `Float64`.

palday: Yeah, I understood the type promotion of the zeros, but it was unclear to me why we needed to promote again for integer division.

nalimilan: I divide by one because `mean` divides by the number of elements. Otherwise for `Integer` inputs the result won't be float.

(I feel like these return type computation issues are really an area that need improvement in Julia.)

nalimilan:
@palday Good to go?

palday (approved):

PR #480, opened 2022-04-22
Adding geometric distribution for GLM.jl

ayushpatnaikgit:
The following is the summary of the changes in the following source files:

1. In docs/src/index.md:

- Added Geometric to the list of distribution in family, under the “Fitting GLM models” section

2. In src/GLM.jl:

- Exported Geometric

3. In src/glmtools.jl:

- Defined `LogLink` as canonical link function for `Geometric` distribution
- Defined `glmvar` function for `Geometric` distribution
- Defined `mustart` function for `Geometric` distribution
- Defined `devresid` function for `Geometric` distribution
- Defined `loglik_obs` function for `Geometric` distribution
- Added an example `glmvar` in docstring
- Added an example `mustart` in docstring

4. In test/runtests.jl:

- Added a test case of `Geometric` distribution with `LogLink`
- Added a test case of `Geometric` distribution with `InverseLink`


src/glmtools.jl
```diff
@@ -461,6 +477,11 @@ function loglik_obs end
 loglik_obs(::Bernoulli, y, μ, wt, ϕ) = wt*logpdf(Bernoulli(μ), y)
 loglik_obs(::Binomial, y, μ, wt, ϕ) = logpdf(Binomial(Int(wt), μ), Int(y*wt))
 loglik_obs(::Gamma, y, μ, wt, ϕ) = wt*logpdf(Gamma(inv(ϕ), μ*ϕ), y)
+# In Julia, a Geometric distribution characterizes the number of failures before
```
nalimilan, line 480:
```suggestion
# In Distributions.jl, a Geometric distribution characterizes the number of failures before 
```

test/runtests.jl
```diff
@@ -555,6 +555,36 @@ end
...
+    test_show(gm22)
+    @test dof(gm22) == 8
+    @test deviance(gm22) ≈ 137.8781581814965
+    @test loglikelihood(gm22) ≈ -548.3711276642073
+    @test aic(gm22) ≈ 1112.7422553284146
+    @test aicc(gm22) ≈ 1113.7933502189255
+    @test bic(gm22) ≈ 1136.6111083020812
+    @test coef(gm22)[1:7] ≈ [2.8978546663153897, -0.5701067649409168, 0.08040181505082235, 
+                            -0.4497584898742737, 0.08622664933901254, 0.3558996662512287, 
+                            0.29016080736927813]
```
nalimilan, line 570: Can you also test `stderror`?

```suggestion
    @test coef(gm22)[1:7] ≈ [2.8978546663153897, -0.5701067649409168, 0.08040181505082235, 
                            -0.4497584898742737, 0.08622664933901254, 0.3558996662512287, 
                             0.29016080736927813]
```

docs/src/index.md
```diff
@@ -34,8 +34,9 @@ functions are
            Bernoulli (LogitLink)
             Binomial (LogitLink)
                Gamma (InverseLink)
+               Geometric (LogLink)
```
nalimilan, line 37:
```suggestion
               Gamma (InverseLink)
           Geometric (LogLink)
```

nalimilan (commented):
Thanks! Have you checked the results against another implementation?

test/runtests.jl
mousum-github: Sure, I will update the test set accordingly.

mousum-github: > 

Since it is a special case of `Negative Binomial` distribution with `r = 1`, we checked with `Negative Binomial`.

nalimilan (approved):

src/glmtools.jl
```diff
@@ -419,6 +431,10 @@ function devresid(::Binomial, y, μ::Real)
     end
 end
 devresid(::Gamma, y, μ::Real) = -2 * (log(y / μ) - (y - μ) / μ)
+function devresid(::Geometric, y, μ::Real)
+    v = 2 * (xlogy(y, y / μ) + xlogy(y + 1, (μ + 1) / (y + 1)))
+    return μ == 0 ? oftype(v, NaN) : v
```
palday, line 436: I would probably move this check above to shortcircuit faster when mu==0.

src/glmtools.jl
```diff
@@ -419,6 +431,10 @@ function devresid(::Binomial, y, μ::Real)
     end
 end
 devresid(::Gamma, y, μ::Real) = -2 * (log(y / μ) - (y - μ) / μ)
+function devresid(::Geometric, y, μ::Real)
+    v = 2 * (xlogy(y, y / μ) + xlogy(y + 1, (μ + 1) / (y + 1)))
```
palday, line 435: The notes I have for squared deviance residuals are:

```julia
    v = 2 * (xlogy(y, y / μ) - xlogy(y + 1, (y + 1) / (μ + 1)))
```

but this is equivalent to the operation here. I have no idea which version will tend to perform better numerically in practice.

palday (approved):
It would be good to document where the numbers in the "Geometric LogLink" test set come from, but otherwise looks good to me.

src/glmtools.jl
mousum-github: Right - I will update the code similar to the following:
`μ == 0 ? NaN : 2 * (xlogy(y, y / μ) + xlogy(y + 1, (μ + 1) / (y + 1)))`

palday:
```julia
   μ == 0 && return oftype(μ, NaN)
   return 2 * (xlogy(y, y / μ) + xlogy(y + 1, (μ + 1) / (y + 1)))
```

is probably how I would do it, but it's really a matter of personal style and what @nalimilan thinks :smile:

nalimilan: I agree it sounds better to adapt the `NaN` type to that of the inputs. But I'm not sure whether it should be the type of `μ`, that of `y`, or `promote_type(typeof(μ), typeof(y))`...

nalimilan:
> It would be good to document where the numbers in the "Geometric LogLink" test set come from

I don't think we note this in the code usually, I think the implicit assumption is that values have been checked against R, either directly or indirectly (here, by checking against `NegativeBinomial` which was previously checked against R).

src/glmtools.jl
mousum-github: Looks like the following one is little faster
`2 * (xlogy(y, y / μ) - xlogy(y + 1, (y + 1) / (μ + 1)))`
Updated the code accordingly.

src/glmtools.jl
palday: `promote_type(typeof(μ), typeof(y))` seems the way to go. Given that we have an expression with `y/mu` and `xlogy` never returns a narrower type than the inputs, we can expect the result to be at least as wide as `y/mu`.

ViralBShah:
Bumping this - what remains to merge?

src/glmtools.jl
```diff
@@ -419,6 +431,10 @@ function devresid(::Binomial, y, μ::Real)
     end
 end
 devresid(::Gamma, y, μ::Real) = -2 * (log(y / μ) - (y - μ) / μ)
+function devresid(::Geometric, y, μ::Real)
+    μ == 0 && return oftype(μ, NaN)
```
nalimilan, line 435:
```suggestion
    μ == 0 && return convert(float(promote_type(typeof(μ), typeof(y))), NaN)
```

src/glmtools.jl
```diff
@@ -419,6 +431,10 @@ function devresid(::Binomial, y, μ::Real)
...
+    return 2 * (xlogy(y, y / μ) - xlogy(y + 1, (y + 1) / (μ + 1)))
+end
 devresid(::InverseGaussian, y, μ::Real) = abs2(y - μ) / (y * abs2(μ))
 function devresid(d::NegativeBinomial, y, μ::Real)
```
nalimilan, line 439: Can you adapt this method to use the same pattern for consistency?

nalimilan (commented):
Sorry, we kind of forgot about this. We just need to fix the discussion about the `devresid` return type.

src/glmtools.jl
mousum-github: changed `oftype(η, NaN)` to `convert(float(typeof(η)), NaN)` wherever possible in `inverselink` functions since `GLM.inverselink(GLM.IdentityLink(), 0)` is throwing an error.

Also, the `devresid` function for Negative Binomial is defined as how it is defined for Geometric.

src/glmtools.jl
```diff
@@ -222,22 +222,22 @@ end
...
-inverselink(::IdentityLink, η::Real) = η, one(η), oftype(η, NaN)
+inverselink(::IdentityLink, η::Real) = η, one(η), convert(float(typeof(η)), NaN)
 
 linkfun(::InverseLink, μ::Real) = inv(μ)
 linkinv(::InverseLink, η::Real) = inv(η)
 mueta(::InverseLink, η::Real) = -inv(abs2(η))
 function inverselink(::InverseLink, η::Real)
     μ = inv(η)
-    return μ, -abs2(μ), oftype(μ, NaN)
+    return μ, -abs2(μ), convert(float(typeof(μ)), NaN)
```
nalimilan, line 232: These are likely not needed as `μ = inv(η)` will (almost?) always be a float. Though I guess that doesn't hurt much.

nalimilan:
Thanks!

PR #481, opened 2022-04-25
Negative r2 for linear model with no intercept

mousum-github:
The Julia `lm` produces negative r2 (-0.15702479338842856) for the following:
`data = DataFrame(x = 60:70, y = 130:140)`
`mdl = lm(@formula(y ~ 0 + x), data)`
`r2(mdl)`

The certified value of r2 = 0.999365492298663
https://www.itl.nist.gov/div898/strd/lls/data/LINKS/DATA/NoInt1.dat

While investigating the reason, we found that the `nulldeviance` calculation is different. Ideally, if the model has the intercept term, then the `nulldeviance` = `y’y – n * mean(y)^2` and if the model does not have the intercept, then the `nulldeviance` = `y’y`.

In this PR, we have attempted to correct the `nulldeviance` calculation. 

The following is the summary of the changes:
- Added a new parameter `hasintercept` of type `Bool` in `nulldeviance` function with existing parameter type `LinResp`. 
- `nulldeviance(obj::LinearModel) = nulldeviance(obj.rr, hasintercept(obj))`
- Added a test case to test beta, standard error of the estimate, dispersion and r2 with the certified values given in the above link


ayushpatnaikgit:
@nalimilan @ViralBShah the PR is only failing on Julia nightly—windows. I don't think it's to do with anything that we've changed. What shall we do in this situation? It's happening with [PR 482](https://github.com/JuliaStats/GLM.jl/pull/482) as well.

ViralBShah:
Best would be to isolate a minimal test case and file a bug against Julia.

ViralBShah:
Actually do check the logs. This has to do with timezones, so ignore. Need to see what is going on though

nalimilan:
I think the TimeZones issue is a network problem, I've seen it before.

Regarding the PR itself, I think we'll have to adjust the definition of `nulldeviance` and `nullloglikelihood` to cover models with no intercept, as currently the null model is defined as the one with only the intercept, but this doesn't make much sense when there's no intercept. This is an issue with have recently discussed too at https://github.com/JuliaStats/GLM.jl/pull/479#issuecomment-1104452820. Cc: @palday

src/lm.jl
```diff
@@ -65,13 +65,16 @@ function deviance(r::LmResp)
     v
 end
 
-function nulldeviance(r::LmResp)
+function nulldeviance(r::LmResp, hasintercept::Bool=true)
```
nalimilan, line 68: Better define this method only for `LinearModel` if it cannot be computed only from information contained in the `LmResp` object.

src/lm.jl
```diff
@@ -65,13 +65,16 @@ function deviance(r::LmResp)
...
-    else 
-        m = mean(r.y, weights(r.wts))
+    m = 0.0
+    if(hasintercept)
+        if isempty(wts)
+            m = mean(y)
+        else 
+            m = mean(r.y, weights(r.wts))
+        end
     end
```
nalimilan, line 78:
```suggestion
    if hasintercept
        if isempty(wts)
            m = mean(y)
        else 
            m = mean(r.y, weights(r.wts))
        end
    else
        m = zero(eltype(y))
    end
```

src/lm.jl
```diff
@@ -200,7 +203,7 @@ deviance(obj::LinearModel) = deviance(obj.rr)
 
 For linear models, the deviance of the null model is equal to the total sum of squares (TSS).
 """
-nulldeviance(obj::LinearModel) = nulldeviance(obj.rr)
+nulldeviance(obj::LinearModel) = nulldeviance(obj.rr, hasintercept(obj))
 loglikelihood(obj::LinearModel) = loglikelihood(obj.rr)
 nullloglikelihood(obj::LinearModel) = nullloglikelihood(obj.rr)
```
nalimilan, line 208: Also change the definition of `nullloglikelihood` for consistency.

test/runtests.jl
```diff
@@ -207,6 +207,18 @@ end
...
+@testset "Linear model with no intercept" begin
+    # test case to test r2 for no intercept model
+    # https://www.itl.nist.gov/div898/strd/lls/data/LINKS/DATA/NoInt1.dat
+
+    data = DataFrame(x = 60:70, y = 130:140)
+    mdl = lm(@formula(y ~ 0 + x), data)
+    @test coef(mdl) ≈ [2.07438016528926]
+    @test stderror(mdl) ≈ [0.165289256198347E-01]
+    @test GLM.dispersion(mdl.model) ≈ 3.56753034006338
+    @test r2(mdl) ≈ 0.999365492298663
```
nalimilan, line 219: Can you also test other indicators (see other tests)?

src/lm.jl
mousum-github: Sure - there will be only one definition of nulldeviance (and as well as null loglikelihood).

src/lm.jl
mousum-github: Thanks. I will update the code.

src/lm.jl
mousum-github: Okay.

test/runtests.jl
mousum-github: Thanks for your suggestion. While testing the `adjusted r2`, I feel we may need to modify the `adjusted r2` also. 
Let me check and test with a few more test cases.

kleinschmidt (commented):
probably good to also test teh case when the model is are not fit to a formula?  e.g. the `lm(X::Matrix, y::Vector)` method.  AFAICT it _should_ Just Work TM (there are methods for `hasintercept` for formula-less models) but good to make sure it doesn't break :)

src/lm.jl
```diff
@@ -200,19 +173,44 @@ deviance(obj::LinearModel) = deviance(obj.rr)
...
 For linear models, the deviance of the null model is equal to the total sum of squares (TSS).
 """
-nulldeviance(obj::LinearModel) = nulldeviance(obj.rr)
-loglikelihood(obj::LinearModel) = loglikelihood(obj.rr)
-nullloglikelihood(obj::LinearModel) = nullloglikelihood(obj.rr)
+function nulldeviance(obj::LinearModel) 
+    y = obj.rr.y
+    wts = obj.rr.wts
+    
+    if(hasintercept(obj))
```
ararslan, line 180:
```suggestion
    if hasintercept(obj)
```

src/lm.jl
```diff
@@ -200,19 +173,44 @@ deviance(obj::LinearModel) = deviance(obj.rr)
...
 
-r2(obj::LinearModel) = 1 - deviance(obj)/nulldeviance(obj)
+    v = zero(eltype(y))*zero(eltype(wts))
+    if isempty(wts)
+        @inbounds @simd for i = 1:length(y)
+            v += abs2(y[i] - m)
+        end
+    else
+        @inbounds @simd for i = 1:length(y)
+            v += abs2(y[i] - m)*wts[i]
```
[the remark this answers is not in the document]
kleinschmidt: wont' `eachindex(y, wts)` find a "lowest common denominator" indexing?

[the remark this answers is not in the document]
kleinschmidt: e.g.

```julia 
julia> eachindex(rand(100, 1), rand(1, 1, 100))
Base.OneTo(100)
```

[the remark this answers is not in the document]
kleinschmidt: or even better lol
```julia
julia> eachindex(rand(100, 1), rand(1, 5, 20))
Base.OneTo(100)
```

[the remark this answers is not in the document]
ararslan: I think it will throw if `axes` aren't compatible, e.g. if one is an OffsetArray

nalimilan:
(For some reason I can't comment in the thread above.) I think `eachindex` is fine here. People shouldn't play fitting models where the response, predictors and/or weights have different axes. That would be asking for trouble. (Note that the current code is silently incorrect in that case anyway, so throwing an error would be an improvement.)

src/lm.jl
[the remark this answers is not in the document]
mousum-github: In the last commit, used `eachindex(y,wts)` instead of `i = 1:length(y)` in `nulldeviance` function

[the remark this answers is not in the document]
mousum-github: Also, added one more test case without providing any formula in the `lm` function as suggested by @kleinschmidt

test/runtests.jl
```diff
@@ -207,6 +207,74 @@ end
...
+        @test adjr2(mdl) ≈ 0.990022172949
+        @test nulldeviance(mdl) ≈ 41.00000000000000
+        @test deviance(mdl) ≈ 0.27272727272727
+        @test aic(mdl) ≈ 5.3199453808329
+        @test loglikelihood(mdl) ≈ -0.6599726904164597
+        @test nullloglikelihood(mdl) ≈ -8.179255266668315
+        @test predict(mdl) ≈ [2.909090909090908, 3.636363636363635, 4.363636363636362]
+    end
+    @testset "Test with without formula" begin
+        X = [4 5 6;]'
```
nalimilan, line 255:
```suggestion
        X = [4 5 6]'
```

nalimilan:
Thanks, looks good! I just wonder what release strategy we should adopt. Technically this could be considered as a breaking change (because it changes the definition of `nulldeviance` and `nullloglikelihood`, cf. https://github.com/JuliaStats/StatsAPI.jl/pull/14) or as a bugfix (because the previous result didn't make a lot of sense, and was even misleading for the R²). We probably don't want to tag GLM 2.0 just for htis. We could adopt an intermediate approach and print a warning (at least temporarily) for models without an intercept. Opinions?

palday:
@nalimilan I like the idea of issuing a warning for models without intercepts and a minor version bump -- it's a good thing to call out anyway that a lot of classical summary definitions are "weird" for models without intercepts.

src/lm.jl
```diff
@@ -200,19 +173,44 @@ deviance(obj::LinearModel) = deviance(obj.rr)
...
+    wts = obj.rr.wts
+    
+    if hasintercept(obj)
+        if isempty(wts)
+            m = mean(y)
+        else 
+            m = mean(y, weights(wts))
+        end
+    else
+        m = zero(eltype(y))
```
nalimilan, line 187: OK, let's try something like this then? It would also be good to adapt Project.toml to require StatsAPI 1.4, which I'll tag shortly (https://github.com/JuliaStats/StatsAPI.jl/pull/14), so that the docstring is consistent with what we do here.
```suggestion
        @warn("Starting from GLM.jl 1.8, null model is defined as having no predictor at all " *
              "when a model without an intercept is passed.")
        m = zero(eltype(y))
```

nalimilan:
@kleinschmidt Technically, it's possible to fit a model which has an intercept but for which `hasintercept` returns false by passing a model matrix which applies full dummy coding without a column of ones, right? Should we try to detect this?

nalimilan:
I don't have rights to edit your branch so I pushed the commit directly to master with the warning. Thanks!

Issue #469, opened 2022-02-09
add custom summary to types in GLM.jl


bkamins:
Example:

```
julia> model = lm(@formula(y ~ x), dataset)
StatsModels.TableRegressionModel{LinearModel{GLM.LmResp{Vector{Float64}},
 GLM.DensePredChol{Float64, LinearAlgebra.CholeskyPivoted{Float64,
 Matrix{Float64}}}}, Matrix{Float64}}

y ~ 1 + x

Coefficients:
───────────────────────────────────────────────────────────────────────
                Coef.  Std. Error     t  Pr(>|t|)  Lower 95%  Upper 95%
───────────────────────────────────────────────────────────────────────
(Intercept)  3.00009     1.12475   2.67    0.0257   0.455737   5.54444
x            0.500091    0.117906  4.24    0.0022   0.23337    0.766812
───────────────────────────────────────────────────────────────────────
```

Users are overwhelmed and confused about the meaning of
```
StatsModels.TableRegressionModel{LinearModel{GLM.LmResp{Vector{Float64}},
 GLM.DensePredChol{Float64, LinearAlgebra.CholeskyPivoted{Float64,
 Matrix{Float64}}}}, Matrix{Float64}}
```
part.

I am not sure what naming convention would be best as I do not know the exact type hierarchy in StatsModels.jl and GLM.jl, but maybe just omitting printing type parameters would be enough?

nalimilan:
Yes, we could probably just print `TableRegressionModel{LinearModel}`. And the day we get rid of `TableRegressionModel` this will just be `LinearModel`.

For GLMs, it would be useful to print something like `TableRegressionModel{GeneralizedLinearModel} with Binomial family and ProbitLink`.

bkamins:
@nalimilan - do you know the type hierarchy well enough to make this PR easily (I can look into it but I would need to learn GLM.jl and StatsModels.jl internals 😄)?

andreasnoack:
I think we can consider this fixed by https://github.com/JuliaStats/GLM.jl/pull/339. No it is printed as
```julia
GeneralizedLinearModel

Y ~ 1 + X

Coefficients:
─────────────────────────────────────────────────────────────────────────
                 Coef.  Std. Error      z  Pr(>|z|)  Lower 95%  Upper 95%
─────────────────────────────────────────────────────────────────────────
(Intercept)  -0.666667    0.62361   -1.07    0.2850   -1.88892   0.555586
[...]

PR #501, opened 2022-09-29
ditch negative zero in ftest output

### LINKED ISSUE #461: `ftest` and `-0.0`
Currently, `ftest` uses the unadjusted R2, which is constrained by definition to be on [0, 1]. As it stands in the tests and doctests, we have several cases where the floating point results for a null model give an R2 of `-0.0`. That's fine numerically, but realistically we know that this should be positive zero. Moreover, if you use a different BLAS (e.g. MKL), then you do get nonnegative `0.0`. This means that the doctests and tests for `show` methods can break based on the BLAS in a way that's much more annoying to fix than simply using `isapprox` instead of `isequal`.

Should we replace 

https://github.com/JuliaStats/GLM.jl/blob/3f9c8e44f85d1d997c38a30580ac447ac9bac55a/src/ftest.jl#L162

with

```julia
    return FTestResult(Int(nobs(mods[1])), SSR, df, max.(r2.(mods), 0), fstat, pval)
```
?

This is in some sense an upstream issue (i.e. the definition of `StatsBase.r2`), but I'm hesitant to apply a constraint at that point because a clearly negative R2 could be a decent test for pathological method definitions of `deviance` and `nulldeviance` for a particular type.

palday:
closes #461.

I also updated some docstring tests so that they would pass. (cf. also #469)


nalimilan:
Thanks. Just another though: maybe we should apply rounding only in `show`? The advantage is that we know we use 4 digits, so we can `-0.0000` to `0.0000` without applying an arbitrary threshold.

This wouldn't fix tests that check internal fields directly, but anyway if we only clamp small negative values to zero, we have to use `isapprox` with some tolerance to handle small positive values.

palday:
Can do, but mathematically this is non-adjusted R2, so should be strictly non-negative. Negative zero is just a floating point issue.

src/ftest.jl
```diff
@@ -187,15 +188,21 @@ function show(io::IO, ftr::FTestResult{N}) where N
     outrows[1, :] = ["", "DOF", "ΔDOF", "SSR", "ΔSSR",
                      "R²", "ΔR²", "F*", "p(>F)"]
 
+    # get rid of negative zero -- doesn't matter mathematically,
+    # but messes up doctests and various other things
+    # cf. Issue #461 
+
+    r2vals = [replace(@sprintf("%.4f", val),  "-0.0000" => "0.0000") for val in ftr.r2]
```
nalimilan, line 195:
```suggestion
    r2vals = [replace(@sprintf("%.4f", val), "-0.0000" => "0.0000") for val in ftr.r2]
```

nalimilan:
Yes, I agree, but the choice of an arbitrary threshold is annoying, and it seems safer to just store whatever `r2` returns to avoid hiding potential bugs in implementations.

nalimilan (approved):

PR #504, opened 2022-10-21
safer int for binomial loglik

### LINKED ISSUE #503: `Int` conversion in `loglik_obs(::Binomial)`
Currently, `loglik_obs(::Binomial)` forces the second argument to `logpdf(::Binomial, x)` to be be an integer:

https://github.com/JuliaStats/GLM.jl/blob/14597370eac4da245216455a91f5c0c7ccfae4a2/src/glmtools.jl#L515

(Sidebar: The PDF function will actually accept non integer values, returning -Inf if they're not within 1eps of an integer or the value of the corresponding integer if they are.)

As part of an analysis at the day job, I found a pathological case where we currently get an `InexactError`:

```julia
julia> using BenchmarkTools

julia> using Distributions

# values from a fitted model
julia> y, μ, wt, ϕ = 0.6376811594202898, 0.8492925285671102, 69.0, NaN # phi isn't used here
(0.6376811594202898, 0.8492925285671102, 69.0, NaN)

julia> # definition from GLM.jl
       loglik_obs(::Binomial, y, μ, wt, ϕ) = logpdf(Binomial(Int(wt), μ), Int(y*wt))
loglik_obs (generic function with 1 method)

julia> loglik_obs(Binomial(), y, μ,  wt, ϕ) # error
ERROR: InexactError: Int64(43.99999999999999)
Stacktrace:
 [1] Int64
   @ ./float.jl:788 [inlined]
 [2] loglik_obs(#unused#::Binomial{Float64}, y::Float64, μ::Float64, wt::Float64, ϕ::Float64)
   @ Main ./REPL[76]:2
 [3] top-level scope
   @ REPL[77]:1
```

If we look at the numbers creating the problem, we have

```julia
julia> y * wt
43.99999999999999

julia> nextfloat(y * wt)
44.0

julia> 44 / wt == y
true

julia> 44 / y == wt
true
``` 

So this really is just a floating point issue. We can fix this by defining a `safer_int` for "close enough"

```julia
function safer_int(x::T) where {T<:Base.IEEEFloat}
    r = round(Int, x)
    abs(x - r) <= eps(x) && return r
    throw(InexactError(:safer_int, T, x))
end

loglik_obs_safer(::Binomial, y, μ, wt, ϕ) = logpdf(Binomial(Int(wt), μ), safer_int(y*wt))
```

and then everything works:

```julia
julia> loglik_obs_safer(Binomial(), y, μ,  wt, ϕ) # works
-11.628163156011077
```

What happens to performance though for a non pathological case? Let's take a look

```julia
julia> y, μ, wt, ϕ = 1/3 , 0.5, 3, NaN
(0.3333333333333333, 0.5, 3, NaN)

julia> @benchmark loglik_obs(Binomial(), $y, $μ,  $wt, $ϕ)
BenchmarkTools.Trial: 10000 samples with 987 evaluations.
 Range (min … max):  50.616 ns … 86.458 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     50.785 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   51.102 ns ±  1.389 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

   ▆█▄                    ▁            ▁                      ▁
  ▆████▆▃▄▃▃▃▂▅▅▅▃▅▄▄▄▄▅▄▅██▆▇▇▇▆▅▄▃▅▅▆█▇▅▆▆▅▅▅▅▅▅▆▇▆▆▅▅▆▅▆▅▅ █
  50.6 ns      Histogram: log(frequency) by time      55.2 ns <

 Memory estimate: 0 bytes, allocs estimate: 0.

julia> @benchmark loglik_obs_safer(Binomial(), $y, $μ,  $wt, $ϕ)
BenchmarkTools.Trial: 10000 samples with 986 evaluations.
 Range (min … max):  53.414 ns … 88.446 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     53.542 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   53.847 ns ±  1.313 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  ▂█▇▃                          ▁▂▁    ▂▂                     ▂
  ████▆▇▅▄▄▃▄▅▄▅▄▅▄▄▄▅▃▆▅▃▃▄███████▇▆▆▇██▅▄▄▅▅▃▅▅▅▃▁▄▄▅▅▅▄▅▆▆ █
  53.4 ns      Histogram: log(frequency) by time      57.9 ns <

 Memory estimate: 0 bytes, allocs estimate: 0.
```

So only about 2-3ns worse.

Tested on Apple silicon:
```julia
julia> versioninfo()
Julia Version 1.8.2
Commit 36034abf260 (2022-09-29 15:21 UTC)
Platform Info:
  OS: macOS (arm64-apple-darwin21.3.0)
  CPU: 8 × Apple M1
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-13.0.1 (ORCJIT, apple-m1)
  Threads: 4 on 4 virtual cores
```

@nalimilan if you have no strong objections, I would just go ahead and open a PR to add this.

Hat-tip @ararslan who rubber-ducked this with me and @haberdashPI who found the pathological case.

palday:
closes #503

### DIFF
--- a/src/glmtools.jl
+++ b/src/glmtools.jl
@@ -507,7 +507,7 @@ Convert to Int, when `x` is within 1 eps of an integer.
 function _safer_int(x::T) where {T<:Base.IEEEFloat}
     r = round(Int, x)
     abs(x - r) <= eps(x) && return r
-    throw(InexactError(:safer_int, T, x))
+    throw(InexactError(nameof(T), T, x))
 end
 
 """
--- a/src/glmtools.jl
+++ b/src/glmtools.jl
@@ -500,17 +500,17 @@ dispersion_parameter(D) = true
 dispersion_parameter(::Union{Bernoulli, Binomial, Poisson}) = false
 
 """
-    _safer_int(x::T)
+    _safe_int(x::T)
     
 Convert to Int, when `x` is within 1 eps of an integer.
 """
-function _safer_int(x::T) where {T<:AbstractFloat}
+function _safe_int(x::T) where {T<:AbstractFloat}
     r = round(Int, x)
     abs(x - r) <= eps(x) && return r
     throw(InexactError(nameof(T), T, x))
 end
 
-_safer_int(x) = Int(x)
+_safe_int(x) = Int(x)
 
 """
     GLM.loglik_obs(D, y, μ, wt, ϕ)
@@ -525,7 +525,7 @@ The loglikelihood of a fitted model is the sum of these values over all the obse
 function loglik_obs end
 
 loglik_obs(::Bernoulli, y, μ, wt, ϕ) = wt*logpdf(Bernoulli(μ), y)
-loglik_obs(::Binomial, y, μ, wt, ϕ) = logpdf(Binomial(Int(wt), μ), _safer_int(y*wt))
+loglik_obs(::Binomial, y, μ, wt, ϕ) = logpdf(Binomial(Int(wt), μ), _safe_int(y*wt))
 loglik_obs(::Gamma, y, μ, wt, ϕ) = wt*logpdf(Gamma(inv(ϕ), μ*ϕ), y)
 # In Distributions.jl, a Geometric distribution characterizes the number of failures before 
 # the first success in a sequence of independent Bernoulli trials with success rate p.
--- a/src/glmtools.jl
+++ b/src/glmtools.jl
@@ -510,6 +510,8 @@ function _safer_int(x::T) where {T<:Base.IEEEFloat}
     throw(InexactError(nameof(T), T, x))
 end
 
+_safer_int(x) = Int(x)
+
 """
     GLM.loglik_obs(D, y, μ, wt, ϕ)
 
--- a/src/glmtools.jl
+++ b/src/glmtools.jl
@@ -500,6 +500,17 @@ dispersion_parameter(D) = true
 dispersion_parameter(::Union{Bernoulli, Binomial, Poisson}) = false
 
 """
+    _safer_int(x::T)
+    
+Convert to Int, when `x` is within 1 eps of an integer.
+"""
+function _safer_int(x::T) where {T<:Base.IEEEFloat}
+    r = round(Int, x)
+    abs(x - r) <= eps(x) && return r
+    throw(InexactError(:safer_int, T, x))
+end
+
+"""
     GLM.loglik_obs(D, y, μ, wt, ϕ)
 
 Returns `wt * logpdf(D(μ, ϕ), y)` where the parameters of `D` are derived from `μ` and `ϕ`.
@@ -512,7 +523,7 @@ The loglikelihood of a fitted model is the sum of these values over all the obse
 function loglik_obs end
 
 loglik_obs(::Bernoulli, y, μ, wt, ϕ) = wt*logpdf(Bernoulli(μ), y)
-loglik_obs(::Binomial, y, μ, wt, ϕ) = logpdf(Binomial(Int(wt), μ), Int(y*wt))
+loglik_obs(::Binomial, y, μ, wt, ϕ) = logpdf(Binomial(Int(wt), μ), _safer_int(y*wt))
 loglik_obs(::Gamma, y, μ, wt, ϕ) = wt*logpdf(Gamma(inv(ϕ), μ*ϕ), y)
 # In Distributions.jl, a Geometric distribution characterizes the number of failures before 
 # the first success in a sequence of independent Bernoulli trials with success rate p.
--- a/src/glmtools.jl
+++ b/src/glmtools.jl
@@ -504,7 +504,7 @@ dispersion_parameter(::Union{Bernoulli, Binomial, Poisson}) = false
     
 Convert to Int, when `x` is within 1 eps of an integer.
 """
-function _safer_int(x::T) where {T<:Base.IEEEFloat}
+function _safer_int(x::T) where {T<:AbstractFloat}
     r = round(Int, x)
     abs(x - r) <= eps(x) && return r
     throw(InexactError(nameof(T), T, x))
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1593,8 +1593,6 @@ end
 
     
 @testset "Floating point error in Binomial loglik" begin
-    @test GLM._safer_int(1.0) == 1
-    @test GLM._safer_int(1.0) isa Int
     @test_throws InexactError GLM._safer_int(1.3)
     @test GLM._safer_int(1) === 1
     # see issue 503
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1445,7 +1445,6 @@ end
         @test aic(mdl1) ≈ aic(mdl2)
         @test predict(mdl1) ≈ predict(mdl2)
     end
-
 end
 
 @testset "dropcollinear with GLMs" begin
@@ -1602,5 +1601,5 @@ end
     # 1. y * wt == 43.99999999999999 
     # 2. 44 / y == wt
     # 3. 44 / wt == y
-    @test GLM.loglik_obs(Binomial(), y, μ,  wt, ϕ) ≈ GLM.logpdf(Binomial(Int(wt), μ), 44)
+    @test GLM.loglik_obs(Binomial(), y, μ, wt, ϕ) ≈ GLM.logpdf(Binomial(Int(wt), μ), 44)
 end
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1593,8 +1593,8 @@ end
 
     
 @testset "Floating point error in Binomial loglik" begin
-    @test_throws InexactError GLM._safer_int(1.3)
-    @test GLM._safer_int(1) === 1
+    @test_throws InexactError GLM._safe_int(1.3)
+    @test GLM._safe_int(1) === 1
     # see issue 503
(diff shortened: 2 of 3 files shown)


test/runtests.jl
```diff
@@ -1591,3 +1592,15 @@ end
...
+@testset "Floating point error in Binomial loglik" begin
+    @test GLM._safer_int(1.0) == 1
+    @test GLM._safer_int(1.0) isa Int
+    # see issue 503
+    y, μ, wt, ϕ = 0.6376811594202898, 0.8492925285671102, 69.0, NaN
+    # due to floating point:
+    # 1. y * wt == 43.99999999999999 
+    # 2. 44 / y == wt
+    # 3. 44 / wt == y
+    @test GLM.loglik_obs(Binomial(), y, μ,  wt, ϕ) ≈ GLM.logpdf(Binomial(Int(wt), μ), 44)
```
kleinschmidt, line 1605:
```suggestion
    @test GLM.loglik_obs(Binomial(), y, μ, wt, ϕ) ≈ GLM.logpdf(Binomial(Int(wt), μ), 44)
```

test/runtests.jl
```diff
@@ -1445,6 +1445,7 @@ end
         @test aic(mdl1) ≈ aic(mdl2)
         @test predict(mdl1) ≈ predict(mdl2)
     end
+
```
kleinschmidt, line 1448:
```suggestion
```

kleinschmidt (approved):
LGTM, thanks for the quick fix!

src/glmtools.jl
```diff
@@ -499,6 +499,17 @@ false
...
 
+"""
+    _safer_int(x::T)
+    
+Convert to Int, when `x` is within 1 eps of an integer.
+"""
+function _safer_int(x::T) where {T<:Base.IEEEFloat}
+    r = round(Int, x)
+    abs(x - r) <= eps(x) && return r
+    throw(InexactError(:safer_int, T, x))
```
ararslan, line 510:
```suggestion
    throw(InexactError(nameof(T), T, x))
```

src/glmtools.jl
```diff
@@ -499,6 +499,17 @@ false
...
+end
```
ararslan, line 511:
```suggestion
end

_safer_int(x) = Int(x)
```

I haven't combed through the GLM internals in a bit but I _believe_ this would be technically breaking without a fallback method, since `y * wt` may not necessarily be `Float16`, `Float32`, or `Float64`, and currently it works as long as its type has a method for the `Int` constructor.

test/runtests.jl
```diff
@@ -1591,3 +1591,15 @@ end
     end
 end
 
+    
+@testset "Floating point error in Binomial loglik" begin
+    @test GLM._safer_int(1.0) == 1
+    @test GLM._safer_int(1.0) isa Int
```
ararslan, line 1597:
```suggestion
    @test GLM._safer_int(1.0) isa Int
    @test_throws InexactError GLM._safer_int(1.3)
    @test GLM._safer_int(1) === 1
```

src/glmtools.jl
palday: ah yeah good call

ararslan (approved):

src/glmtools.jl
```diff
@@ -499,6 +499,19 @@ false
 dispersion_parameter(D) = true
 dispersion_parameter(::Union{Bernoulli, Binomial, Poisson}) = false
 
+"""
+    _safer_int(x::T)
+    
+Convert to Int, when `x` is within 1 eps of an integer.
+"""
+function _safer_int(x::T) where {T<:Base.IEEEFloat}
```
nalimilan, line 507: Why not allow any `AbstractFloat`?

src/glmtools.jl
```diff
@@ -499,6 +499,19 @@ false
 dispersion_parameter(D) = true
 dispersion_parameter(::Union{Bernoulli, Binomial, Poisson}) = false
 
+"""
+    _safer_int(x::T)
```
nalimilan, line 503: Safer than what? :-) Why not just `_safe_int`?

test/runtests.jl
```diff
@@ -1591,3 +1591,17 @@ end
...
```
nalimilan, line 1597: This is redundant now:
```suggestion
```

nalimilan (commented):
Makes sense. Can you check performance not only with a microbenchmark like at https://github.com/JuliaStats/GLM.jl/issues/503 but also by calling `loglikelihood` on a model with large data?

src/glmtools.jl
palday: The logic here is based on IEEE 754 behavior though I guess it's a reasonable bit of logic for finite precision type with `eps()` defined. And `eps()` is defined on `AbstractFloat`, so okay :smile

palday:
```suggestion
function _safer_int(x::T) where {T<:AbstractFloat}
```

src/glmtools.jl
palday: Safer than `Int` itself. I'm not sure it's *safe* per se, but I can rename. It's an internal method anyway

palday:
@nalimilan 

Here's the benchmark:

```julia

using BenchmarkTools
using GLM
using Random

n = 1_000_000
y = rand(MersenneTwister(42), [1, 2, 3], n) ./ 3
wts = fill(3, n)

m = glm(@formula(y ~ 1), (; y), Binomial(); wts)

@benchmark loglikelihood(m) samples=100 seconds=10
```

for current master:
```julia
BenchmarkTools.Trial: 100 samples with 1 evaluation.
 Range (min … max):  78.540 ms … 96.700 ms  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     81.484 ms              ┊ GC (median):    0.00%
 Time  (mean ± σ):   81.709 ms ±  2.325 ms  ┊ GC (mean ± σ):  0.00% ± 0.00%

         ▂▃  ▂  ▁██                                            
  ▃▁▃▁▃▄▄███▆█▄▄███▇▇▄▄▁▃▆▄▃▃▁▁▁▃▁▄▁▁▁▃▁▁▄▁▁▁▁▁▁▁▁▁▁▁▁▃▁▁▁▁▁▃ ▃
  78.5 ms         Histogram: frequency by time        89.7 ms <

 Memory estimate: 16 bytes, allocs estimate: 1.
```

for this branch:
```julia
julia> @benchmark loglikelihood(m) samples=100 seconds=10
BenchmarkTools.Trial: 100 samples with 1 evaluation.
 Range (min … max):  83.523 ms … 91.340 ms  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     85.213 ms              ┊ GC (median):    0.00%
 Time  (mean ± σ):   85.559 ms ±  1.121 ms  ┊ GC (mean ± σ):  0.00% ± 0.00%

         ▁▁▄▇█▅ ▁       ▇▂                                     
  ▃▁▁▃▅▁▁████████▅██▆▃▆▆███▁▃▁▁▁▃▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▃ ▃
  83.5 ms         Histogram: frequency by time        91.1 ms <

 Memory estimate: 16 bytes, allocs estimate: 1.
```

So 4-5ms extra on a million observations and generally within the variability of current master.

This was tested on a different machine than #503, here's the new version info:
```julia
Julia Version 1.8.2
Commit 36034abf260 (2022-09-29 15:21 UTC)
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: 16 × Intel(R) Xeon(R) E-2288G CPU @ 3.70GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-13.0.1 (ORCJIT, skylake)
  Threads: 1 on 16 virtual cores
```

nalimilan (approved):

src/glmtools.jl
```diff
@@ -512,7 +525,7 @@ The loglikelihood of a fitted model is the sum of these values over all the obse
 function loglik_obs end
 
 loglik_obs(::Bernoulli, y, μ, wt, ϕ) = wt*logpdf(Bernoulli(μ), y)
-loglik_obs(::Binomial, y, μ, wt, ϕ) = logpdf(Binomial(Int(wt), μ), Int(y*wt))
+loglik_obs(::Binomial, y, μ, wt, ϕ) = logpdf(Binomial(Int(wt), μ), _safe_int(y*wt))
```
andreasnoack, line 528: An alternative would have been
```julia
loglik_obs(::Binomial, y, μ, wt, ϕ) = logpdf(Beta(y*wt + 1, (1 - y)*wt + 1), μ) - log(wt + 1)
```
It would throw when `y*wt` isn't roughly an integer, though.

ararslan: > It would throw when `y*wt` isn't roughly an integer, though.

That's what `_safe_int` does though?

palday: We could also just leave out the conversion -- `logpdf(::Binomial)` is defined for nearly integers as well, but returns `-Inf` if you're not near enough, but this would essentially be an untrapped error. I figured it would be better to catch that higher in the call stack so that the error is more interpretable. We could even return more informative ones like "response is not a multiple of weight".

Issue #426, opened 2021-04-27
Incorrect linear regression results


xianwenchen:
[Car-Training.csv](https://github.com/JuliaStats/GLM.jl/files/6384056/Car-Training.csv)

This is a bit urgent.

In a home exam that it is ongoing, I have the attached dataset, where students need to estimate models.

Here are the Julia codes:

```
using GLM
using DataFrames
using CSV

data = CSV.read( "Car-Training.csv", DataFrame )
model = @formula( Price ~ Year + Mileage )
results = lm( model, data )
```

The output is the following:
```
StatsModels.TableRegressionModel{LinearModel{GLM.LmResp{Array{Float64,1}},GLM.DensePredChol{Float64,LinearAlgebra.CholeskyPivoted{Float64,Array{Float64,2}}}},Array{Float64,2}}

Price ~ 1 + Year + Mileage

Coefficients:
─────────────────────────────────────────────────────────────────────────────────
                  Coef.    Std. Error       t  Pr(>|t|)    Lower 95%    Upper 95%
─────────────────────────────────────────────────────────────────────────────────
(Intercept)   0.0        NaN           NaN       NaN     NaN          NaN
Year          8.17971      0.167978     48.70    <1e-73    7.84664      8.51278
Mileage      -0.0580528    0.00949846   -6.11    <1e-7    -0.0768865   -0.0392191
─────────────────────────────────────────────────────────────────────────────────
```

The intercept is not estimated. The other two coefficients are not correct.

I tried R, SPSS, and Excel. All gave the same results that are different from Julia. I post the results from R below:

```
R> data <- read.csv("Car-Training.csv")
R> lm(Price ~ Year + Mileage, data = data )

Call:
lm(formula = Price ~ Year + Mileage, data = data)

R> summary( lm(Price ~ Year + Mileage, data = data ) )

Call:
lm(formula = Price ~ Year + Mileage, data = data)

Residuals:
   Min     1Q Median     3Q    Max 
 -2168   -835    -49    567   5402 

Coefficients:
             Estimate Std. Error t value Pr(>|t|)
(Intercept) -2.19e+06   3.52e+05   -6.21  1.1e-08
Year         1.10e+03   1.75e+02    6.25  9.0e-09
Mileage     -2.38e-02   9.84e-03   -2.42    0.017

Residual standard error: 1300 on 104 degrees of freedom
Multiple R-squared:  0.464,     Adjusted R-squared:  0.454 
[...]

Issue #375, opened 2020-06-05
PosDefException with simple linear regression when x/y ranges are small


mortenpi:
I'm getting a `PosDefException: matrix is not positive definite; Cholesky factorization failed` when doing a simple `lm(@formula(y ~ x), df)` when the range of `x` and `y` values is tiny compared to their absolute value. Should be possible to replicate with this dataset:

```julia
δ(n; q=3) = mod(n, q)/(q-1) - 1
function dataset(N)
    x0, Δx, δx = 13.5, 1e-8, 1e-8
    y0, Δy, δy = 200, 1e-7, 2e-7
    DataFrame(
        x = x0 .+ Δx .* (-N:N) .+ δx .* δ.(1:2N+1; q=3),
        y = y0 .+ Δy .* (-N:N) .+ δy .* δ.(3:2N+3; q=5),
    )
end
df = dataset(10)
lm(@formula(y ~ x), df)
```

I have a [full MWE here as a gist](https://gist.github.com/mortenpi/88a80a00cfbc9f6e504c24c5c07eb6d8#file-mwe-jl).

I can see how this is numerically an edge case, but I feel that it should ideally _just work_. For my purposes, I can pretty easily work around the issue by just shifting and scaling the `x` and `y` values, fitting and then scaling the coefficients back:

```julia
function linreg(df::DataFrame, x::Symbol, y::Symbol)
    xs, ys = df[:,x], df[:,y]
    xmin, xmax = extrema(xs)
    ymin, ymax = extrema(ys)
    _df = DataFrame(x=(xs .- xmin) ./ (xmax - xmin), y=(ys .- ymin) ./ (ymax - ymin))
    m = lm(@formula(y ~ x), _df)
    _b, _k = coef(m)
    k = _k * (ymax - ymin) / (xmax - xmin)
    b = _b * (ymax - ymin) + ymin - xmin*k
    return b, k
end
```

Seems to give a reasonable result:

![fit](https://user-images.githubusercontent.com/147757/83823443-2a64c900-a728-11ea-88d7-c087f876cc0e.png)

andreasnoack:
I think this is one of the relatively rare cases where you'd need the extra precision that the QR factorization provides relative to the Cholesky. The functionality for using QR is there but we don't expose it in a nice way so let's keep this one open to track the progress of that.

mortenpi:
[...]

Issue #500, opened 2022-09-29
Version 2.0 Breaking Changes


palday:
> I've noticed that some packages rely on `TableRegressionModel` to support GLM: https://github.com/jmboehm/RegressionTables.jl/issues/128, https://github.com/yufongpeng/AnovaBase.jl/issues/52 and https://github.com/yufongpeng/AnovaGLM.jl/issues/6. Even if they adapt to support the new approach, we'd better bump version to 2.0 to avoid any breakage. That can also be the occasion to drop some long-deprecated API. We should check whether we would like to make any other breaking changes. (A few other packages use `TableRegressionModel` for their own models, it would be good that they also stop using it but there's no hurry.)

_Originally posted by @nalimilan in https://github.com/JuliaStats/GLM.jl/issues/339#issuecomment-1242947968_

Here's a quick list of potentially issues that we might want to try to address as part of a push towards 2.0. Several are relatively straightforward, some could potentially be solved via more extensive documentation, and some will require Decisions to be made (e.g. all the stuff with weights).

- [ ] #339 
- [ ] remove deprecations (search the source for `deprecate` to catch "manual" deprecation warnings)
- [ ] all the issues related to handling of rank deficiency / multicollinearity (#449, #426, #413, #375, #280) because this involves potentially changing defaults 
- [ ] potentially exposing a way for the user to choose between QR/Cholesky with the formula interface? Maybe even defaulting to the slightly slower but more stable QR method?
- [ ] #483 
- [ ] #487
- [ ] #350
- [ ] #259
- [ ] #255
- [ ] #240

There are several other issues I would like to see addressed sooner rather than later, but all are technically nonbreaking, at least under [ColPrac](https://github.com/SciML/ColPrac) guidelines (e.g., changes to the show methods, as raised in #461 and #469).

mousum-github:
Right now we are working on GLM with QR decomposition in two steps
1. LM with QR
2. GLM with QR

Hope this will solve a few issues related to the `PosDefException` as mentioned above.

mousum-github:
I would like to have `Multiple dependent variables`, and `Quasi Likelihood` in GLM 2.0

nalimilan:
Nice to hear you're working on QR! I think we can wait until you finish that before tagging 2.0. OTC, multiple dependent variables and quasi-likelihood do not change current behavior so they can be added later (and we have to discuss whether they should live in this package or in a separate one).

nalimilan:
[...]

PR #207, opened 2017-12-18
Handle rank-deficient LinearModel.  Also changes in docs

dmbates:
For reasons that I am having difficulty remembering, I incorporated documentation changes in these changes to handle rank-deficient linear models as discussed in #200.

I realize this makes it harder to review these changes for rank-deficiency.  I can start a new branch and cherry-pick the rank-deficiency-related commits if that would be preferred.

PR #507, opened 2022-10-29
`QR` decomposition method in GLM.

mousum-github:
It is well known that the `Cholesky` decomposition is fast but numerically unstable for ill-conditioned design/model matrix. Also, some users prefer the `QR` decomposition method in the linear (or generalized linear) model.

In this PR, we have added a keyword argument `method` in the `lm`, `glm` and related functions to control which decomposition method will be used.  The `method` argument takes two values `:cholesky` (the default value and it uses `Cholesky`) and `:qr` (it uses `QR`).

Using QR decomposition in `lm`:
[Issue #426 ](https://github.com/JuliaStats/GLM.jl/issues/426) 
`lm(model, data; method=:qr)`

Coefficients:
- Intercept: -2.18759e6
- Year: 1096.15
- Mileage: -0.0238368

Loglikelihood: -917.7449640609644
r2: 0.46431799718947153
adjr2: 0.454016420212346

[Issue #375](https://github.com/JuliaStats/GLM.jl/issues/375) 
`lm(@formula(y ~ x), df; method=:qr, dropcollinear=false)`

Coefficients:
- Intercept: 65.7485
- x: 9.94455

Summary of additions/changes:

- `DensePredQR` - now supports both pivoted and unpivoted QR decomposition
- `invqr` - compute the inverse of X’WX  from QR decomposition
- `inverse` - a wrapper of `invqr` and `invchol`, used inside the `vcov` function.
- `delbeta!(p::DensePredQR{T,<:QRCompactWY}, r::Vector{T}, wt::Vector{T})`
- `delbeta!(p::DensePredQR{T,<:QRPivoted}, r::Vector{T})`
- `delbeta!(p::DensePredQR{T,<:QRPivoted}, r::Vector{T}, wt::Vector{T})`
-  Added `dropcollinear` and `method` keyword arguments in `negbin` function
- All test cases related to the linear model (lm) are replicated for `Cholesky` and `QR` decompositions 
- All test cases related to the glm are replicated for `Cholesky` and `QR` decompositions 
- Added one example of a linear model with the QR method.

To check and compare the performance of GLM with the QR method, we performed a logistic regression with 3,481,280 rows × 24 columns. Logistic regression took approximately (Intel i7 11th generation with 16GM RAM, Windows 11 OS),
- 56 seconds in R GLM
- 10 seconds in Julia GLM with Cholesky
- 30 seconds in Julia GLM with QR

Finally, @harsharora21 is one of the main contributors to this PR.


data/nasty.csv
```diff
@@ -0,0 +1,10 @@
+﻿LABEL,X,ZERO,MISS,BIG,ONE,HUGE,TINY,ROUND
```
bkamins, line 1: this data is small - I think it will be easier to just embed it in a test set to keep things in one place.

src/linpred.jl
```diff
@@ -35,35 +35,49 @@ end
...
+    T = eltype(X)
+    F = pivot ? pivoted_qr(X) : qr(X)
+    DensePredQR(Matrix{T}(X),
+        zeros(T, p),
+        zeros(T, p),
+        zeros(T, p),
+        F)
+end
+
+#DensePredQR(X::Matrix, beta0::Vector, pivot::Bool=false) = DensePredQR{eltype(X)}(X, beta0, pivot)
```
bkamins, line 77: why do we have here these commented lines?

palday:
@mousum-github for the method kwarg, I would be explicit and use `:qr` or `:cholesky` (instead of `:stable` and `:fast`) and explain the difference in the docs. QR and Cholesky are after all not the only methods for solving the least squares problem. :smile:

data/nasty.csv
mousum-github: Thanks for your suggestion. Updated the code accordingly

src/linpred.jl
mousum-github: Removed the code.

ayushpatnaikgit:
> @mousum-github for the method kwarg, I would be explicit and use `:qr` or `:cholesky` (instead of `:stable` and `:fast`) and explain the difference in the docs. QR and Cholesky are after all not the only methods for solving the least squares problem. smile

Shall we do a poll? 

[❤️](https://emojipedia.org/red-heart/) for `:stable` / `:fast`
👍 for `:qr` / `:cholesky`

We will add `QR` decomposition for `glm` after this, so it's best to get it right once and for all.

ViralBShah:
@palday Just a note that it might be useful to use [LinearSolve.jl](https://github.com/SciML/LinearSolve.jl) for the linear solver APIs, especially once we have optional dependencies support.

ayushpatnaikgit:
> > @mousum-github for the method kwarg, I would be explicit and use `:qr` or `:cholesky` (instead of `:stable` and `:fast`) and explain the difference in the docs. QR and Cholesky are after all not the only methods for solving the least squares problem. smile
> 
> Shall we do a poll?
> 
> [heart](https://emojipedia.org/red-heart/) for `:stable` / `:fast` +1 for `:qr` / `:cholesky`
> 
> We will add `QR` decomposition for `glm` after this, so it's best to get it right once and for all.

@mousum-github it looks like we should go ahead with `:qr`/`:cholesky`.

mousum-github:
> > > @mousum-github for the method kwarg, I would be explicit and use `:qr` or `:cholesky` (instead of `:stable` and `:fast`) and explain the difference in the docs. QR and Cholesky are after all not the only methods for solving the least squares problem. smile
> > 
> > 
> > Shall we do a poll?
> > [heart](https://emojipedia.org/red-heart/) for `:stable` / `:fast` +1 for `:qr` / `:cholesky`
> > We will add `QR` decomposition for `glm` after this, so it's best to get it right once and for all.
> 
> @mousum-github it looks like we should go ahead with `:qr`/`:cholesky`.

okay - we will change accordingly

test/runtests.jl
```diff
@@ -275,6 +275,187 @@ end
...
+        m2p_dep_pos = lm(Xmissingcell, ymissingcell, true; method=:stable)
+        @test_logs (:warn, "Positional argument `allowrankdeficient` is deprecated, use keyword " *
+                    "argument `dropcollinear` instead. Proceeding with positional argument value: true") fit(LinearModel, Xmissingcell, ymissingcell, true)
+        @test isa(m2p_dep_pos.pp.qr, QRPivoted)
+        @test rank(m2p_dep_pos.pp.qr.R) == rank(m2p.pp.qr.R)
+        @test isapprox(deviance(m2p_dep_pos), deviance(m2p))
+        @test isapprox(coef(m2p_dep_pos), coef(m2p))
+    
+        m2p_dep_pos_kw = fit(LinearModel, Xmissingcell, ymissingcell, true;
+            dropcollinear = false, method=:stable)
```
nalimilan, line 451: Fix indentation.

test/runtests.jl
```diff
@@ -275,6 +275,187 @@ end
...
+        @test isapprox(r²(lm1), 0.9990466748057584)
+        @test adjr²(lm1) == adjr2(lm1)
+        @test isapprox(adjr²(lm1), 0.998808343507198)
+        @test isapprox(aic(lm1), -36.409684288095946)
+        @test isapprox(aicc(lm1), -24.409684288095946)
+        @test isapprox(bic(lm1), -37.03440588041178)
+        lm2 = fit(LinearModel, hcat(ones(6), 10form.Carb), form.OptDen, true)
+        @test isa(lm2.pp.chol, CholeskyPivoted)
+        @test lm2.pp.chol.piv == [2, 1]
+        @test isapprox(coef(lm1), coef(lm2) .* [1., 10.])
```
nalimilan, line 302: This doesn't test many methods (notably `stderror`), and we have a more complete test below. Maybe drop it? Same for lm3` and `lm4` below.

test/runtests.jl
```diff
@@ -904,11 +1085,12 @@ end
     m2 = lm(x, y, dropcollinear=false)
 
     p1 = predict(m1, x, interval=:confidence)
-    p2 = predict(m2, x, interval=:confidence)
+    #predict uses chol hence removed
```
nalimilan, line 1088: This should be kept for now as Cholesky is still the default. But add tests to check that these throw an error when QR is used.

nalimilan (commented):
Thanks! Looks quite good.

Regarding tests, can you ensure that we use exactly the same tests for Cholesky and QR, so that the organization of the file is clear and we can easily keep tests for both methods in sync? The names of testsets could mention whether they apply to QR or Cholesky.

test/runtests.jl
mousum-github: Thanks. I will revise it.

test/runtests.jl
mousum-github: Thanks. I will revise it.

test/runtests.jl
mousum-github: Thanks. I will check it.

test/runtests.jl
```diff
@@ -42,8 +42,8 @@ linreg(x::AbstractVecOrMat, y::AbstractVector) = qr!(simplemm(x)) \ y
     @test isapprox(aicc(lm1), -24.409684288095946)
     @test isapprox(bic(lm1), -37.03440588041178)
     lm2 = fit(LinearModel, hcat(ones(6), 10form.Carb), form.OptDen, true)
-    @test isa(lm2.pp.chol, CholeskyPivoted)
-    @test lm2.pp.chol.piv == [2, 1]
+#    @test isa(lm2.pp.chol, CholeskyPivoted)
+#    @test lm2.pp.chol.piv == [2, 1]
```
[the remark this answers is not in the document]
mousum-github: I will revise it.

src/linpred.jl
```diff
@@ -227,7 +271,34 @@ end
 LinearAlgebra.cholesky(p::SparsePredChol{T}) where {T} = copy(p.chol)
 LinearAlgebra.cholesky!(p::SparsePredChol{T}) where {T} = p.chol
 
+function invqr(x::DensePredQR{T,<: QRCompactWY}) where T
+    Q,R = x.qr
+    #Rinv = inv(R) - which one is better inv(R) or R\I?
```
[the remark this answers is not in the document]
mousum-github: Thanks. I will update the code accordingly.

src/linpred.jl
```diff
@@ -35,35 +35,49 @@ end
 """
     DensePredQR
 
-A `LinPred` type with a dense, unpivoted QR decomposition of `X`
+A `LinPred` type with a dense QR decomposition (both pivoted and unpivoted) of `X`
```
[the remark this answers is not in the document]
mousum-github: okay

src/linpred.jl
```diff
@@ -124,7 +168,7 @@ function cholesky(p::DensePredChol{T}) where T<:FP
     c = p.chol
     Cholesky(copy(cholfactors(c)), c.uplo, c.info)
 end
-cholesky!(p::DensePredQR{T}) where {T<:FP} = Cholesky{T,typeof(p.X)}(p.qr.R, 'U', 0)
+cholesky!(p::DensePredQR{T}) where {T<:FP} = p.chol
```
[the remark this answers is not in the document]
mousum-github: okay

src/GLM.jl
```diff
@@ -85,8 +85,10 @@ module GLM
 
     @static if VERSION < v"1.8.0-DEV.1139"
         pivoted_cholesky!(A; kwargs...) = cholesky!(A, Val(true); kwargs...)
+        pivoted_qr(A; kwargs...) = qr(A, Val(true); kwargs...)
```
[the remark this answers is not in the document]
mousum-github: We are passing `X'X`  matrix in  `cholesky` whereas in `qr`, we are passing `X` matrix.  So, if we want to use `qr!` then we need to pass a copy of `X` matrix whenever applicable. I have changed the code accordingly.

mousum-github:
Hi @nalimilan, @bkamins and others - looking for your suggestions.

nalimilan (commented):
Thanks (and sorry for the force rebasing). Looks almost ready now, though I have a few more minor comments.

src/linpred.jl
```diff
@@ -163,7 +163,7 @@ function cholesky(p::DensePredChol{T}) where T<:FP
     c = p.chol
     Cholesky(copy(cholfactors(c)), c.uplo, c.info)
 end
-cholesky!(p::DensePredQR{T}) where {T<:FP} = p.chol
+cholesky!(p::DensePredQR{T}) where {T<:FP} = Cholesky{T,typeof(p.X)}(p.qr.R, 'U', 0)
```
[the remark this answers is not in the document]
mousum-github: Thanks for your suggestions.

src/lm.jl
```diff
@@ -296,18 +317,32 @@ function StatsModels.predict!(res::Union{AbstractVector,
...
+        if mm.pp isa DensePredChol
+            chol = cholesky!(mm.pp)
+            if chol isa CholeskyPivoted
+                ip = invperm(chol.p)
+                R = chol.U[ip, ip]
+            else
+                R = chol.U
+            end
+        elseif mm.pp isa DensePredQR
+            qr = mm.pp.qr
```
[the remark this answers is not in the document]
mousum-github: Thanks for your suggestions. We have updated some based on your suggestions. ALthough, we have not created created a function yet

src/linpred.jl
```diff
@@ -115,6 +153,7 @@ function DensePredChol(X::AbstractMatrix, pivot::Bool)
 end
 
 cholpred(X::AbstractMatrix, pivot::Bool=false) = DensePredChol(X, pivot)
+qrpred(X::AbstractMatrix, pivot::Bool=false) = DensePredQR(float(X), pivot)
```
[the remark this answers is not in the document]
mousum-github: We have updated this.

test/runtests.jl
```diff
@@ -103,7 +158,48 @@ end
...
+    @test loglikelihood(lm_model) ≈ loglikelihood(lm_qr_model)
+    @test nullloglikelihood(lm_model) ≈ nullloglikelihood(lm_qr_model)
+    @test residuals(lm_model) ≈ residuals(lm_qr_model)
+    @test aic(lm_model) ≈ aic(lm_qr_model)
+    @test aicc(lm_model) ≈ aicc(lm_qr_model)
+    @test bic(lm_model) ≈ bic(lm_qr_model)
+    @test GLM.dispersion(lm_model) ≈ GLM.dispersion(lm_qr_model)
+end
+
+@testset "Linear model with QR dropcollinearity" begin
```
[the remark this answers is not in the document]
mousum-github: Thanks for pointing that out. We have added a test case.

bkamins:
Can you please check why tests fail on nightly?

src/lm.jl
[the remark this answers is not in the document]
mousum-github: Bypassing these calculations using the `vcov` function.

src/lm.jl
```diff
@@ -296,18 +317,32 @@ function StatsModels.predict!(res::Union{AbstractVector,
...
+                R = qr.R
+            end
         end
+
         dev = deviance(mm)
         dofr = dof_residual(mm)
         residvar = fill(dev/dofr, size(newx, 2), 1)
         ret = dropdims((newx/R).^2 * residvar, dims=2)
+        # We can calculate `ret` compactfully with the following, but it might be less efficient 
+        # ret = diag(newx*vcov(mm)*newx')
```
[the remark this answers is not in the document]
mousum-github: Tested the performance with a model matrix of size 1000*25. `@benchmark` shows equivalent performance.

Previous:
image

Using the `vcov` function:
image

mousum-github:
> 

It is failing the following test
`@test_broken glm(@formula(y ~ x1 + x2), df, Normal(), IdentityLink())`
in `Saturated linear model` test case

test/runtests.jl
```diff
@@ -1383,7 +1777,7 @@ end
 
 @testset "Issue 153" begin
     X = [ones(10) randn(10)]
-    Test.@inferred cholesky(GLM.DensePredQR{Float64}(X))
+    Test.@inferred cholesky(GLM.DensePredQR(X))
```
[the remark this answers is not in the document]
harsharora21: @nalimilan GLM.DensePredQR{Float64} is no longer a function. It is only a constructor now. It was superseded by GLM.DensePredQR, hence the test change.

mousum-github:
Initially, we thought to raise two PRs to incorporate the QR decomposition method in `lm` and `glm`. While preparing the QR method in `glm` after the QR method in `lm`, we found that both share almost 80% of the new code - hence a single PR.

docs/src/examples.md
```diff
@@ -86,6 +86,24 @@ julia> round.(vcov(ols); digits=5)
   0.38889  -0.16667
  -0.16667   0.08333
 ```
+By default, the `lm` method uses the `Cholesky` factorization which is known as fast but numerically unstable, especially for ill-conditioned design matrices. You can use the `method` keyword argument to apply a more stable `QR` factorization method.
+
+```jldoctest
+julia> data = DataFrame(X=[1,2,3], Y=[2,4,7]);
```
bkamins, line 92: maybe choose an example in which Cholesky indeed does not work well?

src/linpred.jl
```diff
@@ -167,7 +247,7 @@ function delbeta!(p::DensePredChol{T,<:CholeskyPivoted}, r::Vector{T}, wt::Vecto
     mul!(delbeta, transpose(p.scratchm1), r)
     # calculate delbeta = (X'WX)\X'Wr
     rnk = rank(p.chol)
-    if rnk == length(delbeta)
+    if rnk === length(delbeta)
```
bkamins, line 250: why this change? It should not matter, but I guess we want to check if values are equal, and do not care that their types are also equal.

docs/src/examples.md
mousum-github: Sure, I will add one.

src/linpred.jl
mousum-github: Got it. Thanks for your suggestion.

ViralBShah:
@mousum-github Why are the tests failing on master?

ayushpatnaikgit:
> @mousum-github Why are the tests failing on master?

@ViralBShah do you mean why the tests are failing on the nightly builds?

mousum-github:
> @mousum-github Why are the tests failing on master?

#### The following is the test which is failing in Julia nighly build
```
julia> using DataFrames, GLM, Test

julia> df = DataFrame(x1=["a", "b", "c"], x2=["a", "b", "c"], y=[1, 2, 3]);

julia> glm(@formula(y ~ x1 + x2), df, Normal(), IdentityLink())
GeneralizedLinearModel

y ~ 1 + x1 + x2

Coefficients:
──────────────────────────────────────────────────────────────────────
             Coef.  Std. Error       z  Pr(>|z|)  Lower 95%  Upper 95%
──────────────────────────────────────────────────────────────────────
(Intercept)    1.0         Inf    0.00    1.0000       -Inf        Inf
x1: b          1.0         Inf    0.00    1.0000       -Inf        Inf
x1: c          2.0         Inf    0.00    1.0000       -Inf        Inf
x2: b          0.0        NaN   NaN       NaN          NaN        NaN
x2: c          0.0        NaN   NaN       NaN          NaN        NaN
──────────────────────────────────────────────────────────────────────


julia> @test_broken glm(@formula(y ~ x1 + x2), df, Normal(), IdentityLink())
Error During Test at REPL[20]:1
  Expression evaluated to non-Boolean
  Expression: glm(#= REPL[20]:1 =# @formula(y ~ x1 + x2), df, Normal(), IdentityLink())
       Value: GeneralizedLinearModel

y ~ 1 + x1 + x2

Coefficients:
──────────────────────────────────────────────────────────────────────
             Coef.  Std. Error       z  Pr(>|z|)  Lower 95%  Upper 95%
──────────────────────────────────────────────────────────────────────
(Intercept)    1.0         Inf    0.00    1.0000       -Inf        Inf
x1: b          1.0         Inf    0.00    1.0000       -Inf        Inf
x1: c          2.0         Inf    0.00    1.0000       -Inf        Inf
x2: b          0.0        NaN   NaN       NaN          NaN        NaN
x2: c          0.0        NaN   NaN       NaN          NaN        NaN
──────────────────────────────────────────────────────────────────────
```

#### And the following is the output in the stable build
```
julia> using DataFrames, GLM, Test

julia> df = DataFrame(x1=["a", "b", "c"], x2=["a", "b", "c"], y=[1, 2, 3]);

julia> glm(@formula(y ~ x1 + x2), df, Normal(), IdentityLink())
GeneralizedLinearModel

y ~ 1 + x1 + x2

Coefficients:
──────────────────────────────────────────────────────────────────────
             Coef.  Std. Error       z  Pr(>|z|)  Lower 95%  Upper 95%
──────────────────────────────────────────────────────────────────────
(Intercept)    1.0         Inf    0.00    1.0000       -Inf        Inf
x1: b          1.0         Inf    0.00    1.0000       -Inf        Inf
x1: c          2.0         Inf    0.00    1.0000       -Inf        Inf
x2: b          0.0        NaN   NaN       NaN          NaN        NaN
x2: c          0.0        NaN   NaN       NaN          NaN        NaN
──────────────────────────────────────────────────────────────────────


julia> @test_broken glm(@formula(y ~ x1 + x2), df, Normal(), IdentityLink())
Test Broken
  Expression: glm(#= REPL[10]:1 =# @formula(y ~ x1 + x2), df, Normal(), IdentityLink())
```
The test is performed under ` Test Saturated linear model with Cholesky`

bkamins:
Section 6-3b from Wooldridge 7e is a good example when standard estimation fails. Maybe we could include it in tests to make sure it is resolved:
```
lm(@formula(rdintens ~ sales + sales^2), rdchem)
```
(dataset is `rdchem`)

mousum-github:
Thanks, @bkamins. The following are the outputs from R and Julia: -

Output from R:
```
Call:
lm(formula = rdintens ~ sales + I(sales^2), data = rdchem)

Residuals:
            Min              1Q          Median              3Q             Max 
-2.141830034005 -1.362962168805 -0.225670654047  1.068750711730  5.580801329956 

Coefficients:
                       Estimate          Std. Error  t value  Pr(>|t|)    
(Intercept)  2.612512084995e+00  4.294417551587e-01  6.08351 1.267e-06 ***
sales        3.005713005396e-04  1.392952806164e-04  2.15780  0.039362 *  
I(sales^2)  -6.945939423917e-09  3.726138613681e-09 -1.86411  0.072458 .  
```

Outputs from Julia:
```
julia> mdl = lm(@formula(rdintens ~ sales + sales^2), rdchem; method=:cholesky)
LinearModel

rdintens ~ 1 + sales + :(sales ^ 2)

Coefficients:
───────────────────────────────────────────────────────────────────────────────────────
                    Coef.     Std. Error       t  Pr(>|t|)      Lower 95%     Upper 95%
───────────────────────────────────────────────────────────────────────────────────────
(Intercept)   0.0          NaN            NaN       NaN     NaN            NaN
sales         0.000852509    0.000156784    5.44    <1e-05    0.000532313    0.00117271
sales ^ 2    -1.97385e-8     4.56287e-9    -4.33    0.0002   -2.90571e-8    -1.04199e-8
───────────────────────────────────────────────────────────────────────────────────────
```
and 
```
julia> mdl = lm(@formula(rdintens ~ sales + sales^2), rdchem; method=:qr)
LinearModel

rdintens ~ 1 + sales + :(sales ^ 2)

Coefficients:
─────────────────────────────────────────────────────────────────────────────────
                    Coef.   Std. Error      t  Pr(>|t|)    Lower 95%    Upper 95%
─────────────────────────────────────────────────────────────────────────────────
(Intercept)   2.61251      0.429442      6.08    <1e-05   1.73421     3.49082
sales         0.000300571  0.000139295   2.16    0.0394   1.56805e-5  0.000585462
sales ^ 2    -6.94594e-9   3.72614e-9   -1.86    0.0725  -1.45667e-8  6.7487e-10
─────────────────────────────────────────────────────────────────────────────────
```

mousum-github:
I will add a test case accordingly.

bkamins:
Nice, thank you!

bkamins:
In fact - I was thinking the following:

if we use default decomposition and in the output we get any value that is `!isfinite` and in the input data all values are `isfinite` maybe we should automatically try doing `:qr` as the second step? The rationale is that the first `!isfinite` test should be quite cheap to do (or maybe there is some other better way to detect that the default decomposition was not accurate enough?).

mousum-github:
In that case, I will prefer to have something like the following in `lm` (and `glm`).

```
if method === :auto
        decomobj = cholpred(X, dropcollinear)
        if rank(decomobj.col) == size(X, 2)
            fit!(LinearModel(LmResp(y, wts), decomobj, nothing))
        else
            fit!(LinearModel(LmResp(y, wts), qrpred(X, dropcollinear), nothing))
            
    if method === :cholesky
        fit!(LinearModel(LmResp(y, wts), cholpred(X, dropcollinear), nothing))
    elseif method === :qr
        fit!(LinearModel(LmResp(y, wts), qrpred(X, dropcollinear), nothing))
    else
        throw(ArgumentError("The only supported values for keyword argument `method` are `auto`, `:cholesky` and `:qr`."))
    end
```

bkamins:
Yes - makes sense.

ViralBShah:
Would be nice to see this one get merged.

docs/src/examples.md
```diff
@@ -86,6 +86,45 @@ julia> round.(vcov(ols); digits=5)
   0.38889  -0.16667
  -0.16667   0.08333
 ```
+By default, the `lm` method uses the `Cholesky` factorization which is known as fast but numerically unstable, especially for ill-conditioned design matrices. You can use the `method` keyword argument to apply a more stable `QR` factorization method.
```
bkamins, line 89: everywhere in docs - should code `Cholesky` be `:cholesky`. The same for `QR` and `:qr`. In sentences where Cholesky an QR are used in mathematical sense I think they do not need to be backquoted.

bkamins: also - is this sentence still true? We discussed that automatic switching to `:qr` would be performed if `:cholesky` failed.

docs/src/examples.md
```diff
@@ -86,6 +86,45 @@ julia> round.(vcov(ols); digits=5)
...
+The following example shows that `QR` decomposition works better for ill-conditioned design matrix. The linear model with the `Cholesky` decomposition method is unable to estimate parameters correctly whereas the linear model with the `QR` decomposition does.
+
+```jldoctest
+julia> y = [1, 2, 3, 4, 5];
+
+julia> x = [1.0E-12, 2.0E-12, 3.0E-12, 4.0E-12, 5.0E-12];
+
+julia> nasty = DataFrame(y = y, x = x);
+
+julia> mdl1 = lm(@formula(y ~ x), nasty; method=:cholesky);
```
bkamins, line 118: then we should show here the default behavior (which should produce a correct result now)

src/glmfit.jl
```diff
@@ -575,7 +578,15 @@ function fit(::Type{M},
...
 
     rr = GlmResp(y, d, l, offset, wts)
-    res = M(rr, cholpred(X, dropcollinear), nothing, false)
+
+    if method === :cholesky
+        res = M(rr, cholpred(X, dropcollinear), nothing, false)
+    elseif method === :qr
+        res = M(rr, qrpred(X, dropcollinear), nothing, false)
+    else
+        throw(ArgumentError("The only supported values for keyword argument `method` are `:cholesky` and `:qr`."))
```
bkamins, line 587: we discussed to also allow `:auto`

src/linpred.jl
```diff
@@ -35,45 +35,125 @@ end
...
 
 Evaluate and return `p.delbeta` the increment to the coefficient vector from residual `r`
 """
 function delbeta! end
 
-function delbeta!(p::DensePredQR{T}, r::Vector{T}) where T<:BlasReal
+function delbeta!(p::DensePredQR{T,<:QRCompactWY}, r::Vector{T}) where T<:BlasReal
+    rnk = rank(p.qr.R)
+    rnk == length(p.delbeta) || 
+        throw(error("One or more columns in the design matrix are linearly dependent on others"))
```
bkamins, line 93: I would propose a more specific error type thrown.

src/linpred.jl
```diff
@@ -35,45 +35,125 @@ end
...
     p.delbeta = p.qr\r
+    mul!(p.scratchm1, Diagonal(ones(size(r))), p.X)
+    return p
+end
+
+function delbeta!(p::DensePredQR{T,<:QRCompactWY}, r::Vector{T}, wt::Vector{T}) where T<:BlasReal
+    rnk = rank(p.qr.R)
+    rnk == length(p.delbeta) || 
+        throw(error("One or more columns in the design matrix are linearly dependent on others"))
```
bkamins, line 102: same here

mousum-github:
Thanks, @bkamins - I will go through this today.

docs/src/examples.md
palday: @bkamins if we're tying this into [the 2.0 release](https://github.com/JuliaStats/GLM.jl/issues/500) (which it looks that way), then we can just default to QR.

bkamins: As you prefer. I think that most users would prefer to have slower but reliable option as a default.

bkamins: Cholesky fails for numerical reasons often (and this is unrelated to statistical error):
```
julia> x = [1, 2, 3]
3-element Vector{Int64}:
 1
 2
 3

julia> y = [1, 2, 4]
3-element Vector{Int64}:
 1
 2
 4

julia> lm(@formula(y~x), (; x, y))
StatsModels.TableRegressionModel{LinearModel{GLM.LmResp{Vector{Float64}}, GLM.DensePredChol{Float64, LinearAlgebra.CholeskyPivoted{Float64, Matrix{Float64}, Vector{Int64}}}}, Matrix{Float64}}

y ~ 1 + x

Coefficients:
─────────────────────────────────────────────────────────────────────────
                 Coef.  Std. Error      t  Pr(>|t|)  Lower 95%  Upper 95%
─────────────────────────────────────────────────────────────────────────
(Intercept)  -0.666667    0.62361   -1.07    0.4788   -8.59038    7.25704
x             1.5         0.288675   5.20    0.1210   -2.16797    5.16797
─────────────────────────────────────────────────────────────────────────

julia> lm(@formula(y~x), (x=10^9*x, y=y))
StatsModels.TableRegressionModel{LinearModel{GLM.LmResp{Vector{Float64}}, GLM.DensePredChol{Float64, LinearAlgebra.CholeskyPivoted{Float64, Matrix{Float64}, Vector{Int64}}}}, Matrix{Float64}}

y ~ 1 + x

Coefficients:
─────────────────────────────────────────────────────────────────────────────────────
                  Coef.     Std. Error       t  Pr(>|t|)      Lower 95%     Upper 95%
─────────────────────────────────────────────────────────────────────────────────────
(Intercept)  0.0         NaN            NaN       NaN     NaN            NaN
x            1.21429e-9    1.12938e-10   10.75    0.0085    7.28351e-10    1.70022e-9
─────────────────────────────────────────────────────────────────────────────────────
```

andreasnoack: If the scaling of the variables is extreme then indeed the rank estimate of the pivoted cholesky would suggest a reduced rank. I don't think that is the typical reason for rank deficiency, though. I think that is multicollinearity. Notice also that the Cholesky doesn't actually fail in your example so it's a consequence of the chosen tolerance
```
julia> X4 = [ones(3) 1e100*x]
3×2 Matrix{Float64}:
 1.0  1.0e100
 1.0  2.0e100
 1.0  3.0e100

julia> cholesky(X4'X4)
Cholesky{Float64, Matrix{Float64}}
U factor:
2×2 UpperTriangular{Float64, Matrix{Float64}}:
 1.73205  3.4641e100
  ⋅       1.41421e100
```

bkamins: > I don't think that is the typical reason for rank deficiency, though. I think that is multicollinearity. Notice also that the Cholesky doesn't actually fail in your example so it's a consequence of the chosen tolerance

I am in the middle of porting all examples from Wooldridge 7e to GLM.jl. In first 9 chapters I GLM.jl fails to correctly estimate the model I think 3 or 4 times. In each case the reason is the case I have presented (but in all cases on real data - not artificial like my example above).

I was taking perspective of a user (I have never maintained GLM.jl) - and I did not investigate what caused to the problem, but I knew that manually doing QR worked. If the problem is with tolerance then I think we should also fix it.

andreasnoack: Its possible that my gut feeling isn't right and that people use highly variably scaled variables more often that I have thought. It might also be that we shouldn't reduce the rank unless strictly required. Do you have the failing examples from Wooldridge handy? Then I can take a look. I have the 6th edition so I'd think the examples are mostly identical.

andreasnoack: I just realized that you'd already mentioned the example in the main thread. Sorry about that.

I just tried the first one and indeed adjusting the tolerance fixes the issue.

Current:
```julia
julia> lm(@formula(rdintens ~ sales + salessq), rdchem_df)
LinearModel

rdintens ~ 1 + sales + salessq

Coefficients:
───────────────────────────────────────────────────────────────────────────────────────
                    Coef.     Std. Error       t  Pr(>|t|)      Lower 95%     Upper 95%
───────────────────────────────────────────────────────────────────────────────────────
(Intercept)   0.0          NaN            NaN       NaN     NaN            NaN
sales         0.000852509    0.000156784    5.44    <1e-05    0.000532313    0.00117271
salessq      -1.97385e-8     4.56287e-9    -4.33    0.0002   -2.90571e-8    -1.04199e-8
───────────────────────────────────────────────────────────────────────────────────────
```
and with `tol=0` in https://github.com/JuliaStats/GLM.jl/blob/57669ef89de3588ffdbf757d82fd28d5c71f5177/src/linpred.jl#L107
```julia
julia> lm(@formula(rdintens ~ sales + salessq), rdchem_df)
LinearModel

rdintens ~ 1 + sales + salessq

Coefficients:
─────────────────────────────────────────────────────────────────────────────────
                    Coef.   Std. Error      t  Pr(>|t|)    Lower 95%    Upper 95%
─────────────────────────────────────────────────────────────────────────────────
(Intercept)   2.61251      0.429442      6.08    <1e-05   1.73421     3.49082
sales         0.000300571  0.000139295   2.16    0.0394   1.56805e-5  0.000585462
salessq      -6.94594e-9   3.72614e-9   -1.86    0.0725  -1.45667e-8  6.7487e-10
─────────────────────────────────────────────────────────────────────────────────
```
which matches the result from R. I'll file a separate issue about that but I don't think we should change any default or introduce `:auto` in this PR.

bkamins: OK - as I have said - I have not tested it in detail (I am reporting from user's perspective 😄).
@mousum-github - can you then please give examples when Cholesky fails in practice and QR does not fail?

@andreasnoack - regarding tolerance: yes, in particular it would be worth to dicsuss when setting such tolerance would lead to the reverse problems (i.e. not detectin multicollinearity when it is present; these issues are unfortunately tricky).

andreasnoack: I've opened https://github.com/JuliaStats/GLM.jl/pull/518 to see the effect of changing the tolerance and we can't discuss the details there but my current thinking is that we should probably only drop variables when it's numerically necessary. It's indeed tricky, though.

docs/src/examples.md
mousum-github: The problem is that the estimates (small numbers) differ in different systems.  Initially, we added the following: -
```
julia> y = [1, 2, 3, 4, 5];
julia> x = [1.0E-12, 2.0E-12, 3.0E-12, 4.0E-12, 5.0E-12];
julia> nasty = DataFrame(y = y, x = x);
julia> lm(@formula(y ~ x), nasty; method=:cholesky)
LinearModel
y ~ 1 + x
Coefficients:
──────────────────────────────────────────────────────────────────────
             Coef.  Std. Error       t  Pr(>|t|)  Lower 95%  Upper 95%
──────────────────────────────────────────────────────────────────────
(Intercept)    3.0    0.707107    4.24    0.0132    1.03676    4.96324
x              0.0  NaN         NaN       NaN     NaN        NaN
──────────────────────────────────────────────────────────────────────
julia> lm(@formula(y ~ x), nasty; method=:qr)
LinearModel
y ~ 1 + x
Coefficients:
──────────────────────────────────────────────────────────────────────────────────────────────
                   Coef.   Std. Error                    t  Pr(>|t|)    Lower 95%    Upper 95%
──────────────────────────────────────────────────────────────────────────────────────────────
(Intercept)  7.94411e-16  1.2026e-15                  0.66    0.5561  -3.0328e-15  4.62162e-15
x            1.0e12       0.000362597  2757880273211543.50    <1e-45   1.0e12      1.0e12
──────────────────────────────────────────────────────────────────────────────────────────────
```
but the doc test failed.

Please let us know if we can handle it better way.

bkamins: Yes, I know. I would show them anyway, but just disable doctesting the results.

src/linpred.jl
mousum-github: I agree. I was thinking to use `SingularException`. Since QR decomposition does not throw this error, I was a little hesitant to use it. Let me check this again,

bkamins:
@andreasnoack - could you please summarize here the conclusions from your investigation in #518? What are the conclusions for the design we should make (and I understand we have a tension between two kinds of errors we can make).

Do you think we should default to `:cholesky` and have `:qr` as opt in (without `:auto` as the default) as it would be breaking? Then maybe we should add a warning in case when `:cholesky` detects multicolinearity that user might want to check `:qr`? I am not 100% sure what is best.

src/glmfit.jl
mousum-github: The changes are almost done. Hope to update it by tomorrow.

docs/src/examples.md
ayushpatnaikgit: So for this doctring, do we just change three back ticks followed by `jldoctest` to three back ticks followed by `julia`?

bkamins: yes, just not run doctest on this line. (I know it is not the cleanest, but the issue is that we are using linear algebra package that does not produce consistent results on different machines)

src/glmfit.jl
mousum-github: And the following with `:qr`
```
julia> [test(eps, mul) for eps in [1/10^i for i in 0:10], mul in [10^i for i in 0:9]]
11×10 Matrix{Bool}:
 0  0  0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  0  0  0
```

bkamins: yes, but the question is when would you start getting ones (i.e. at some point columns indeed become lineary dependent).

ViralBShah:
I think keeping `cholesky` default and warning when falling back to `qr` makes sense. 

Simultaneously for changing the default, we should announce it as upcoming in the README for a whole release cycle (and a few months) and change it in the next release.

andreasnoack:
> could you please summarize here the conclusions from your investigation in https://github.com/JuliaStats/GLM.jl/pull/518?

Let me try. The mechanism for excluding variable when there is *numerical* collinearity introduced in https://github.com/JuliaStats/GLM.jl/pull/207 depends on a numerical rank determination (currently) based on the pivoted Cholesky decomposition. However, because of the famous squaring-of-the-condition-number-effect of `X'X`, there isn't much space between reduced rank because of bad scaling and reduced rank because of perfect multicollinearity. That is why the Wooldridge example can't currently be reproduced with GLM as discussed in https://github.com/JuliaStats/GLM.jl/pull/507#discussion_r1145045405 https://github.com/JuliaStats/GLM.jl/pull/507#discussion_r1147246862. So there are the following options

1. Continue to use Cholesky and compute the numerical rank based on `tol=-1`. The pros are that the current multicollinearity detection examples will continue to work, it will be fast and memory efficient. The contra point is that many examples that will work in other software will drop variables in GLM.
2. Continue to use Cholesky but set `tol=0`. This will be fast and memory efficient and only drop variables when strictly required. I.e. when `X'X` goes singular or slightly indefinite. Problems with multicollinearity will show up as very large confidence bounds and therefore require manual removal of explanatory variables. I like this approach but others might not.
3. Use pivoted QR when `dropcollinear=true` but otherwise use Cholesky or PivotedCholesky with `tol=0`. That would ensure that the Wooldridge like examples work while also detecting perfect multicollinearity. The contra points here would be the extra cost of computing the QR both in flops and memory and the added complexity in switching algorithm based on the `dropcollinear` argument.
4. Always use QR and use the pivoted if `dropcollinear=true`. This will have the same benefit as 3. regarding the detection of multicollinear variables but you'd pay the price of slower computations.
5. Try catch on the Cholesky and fall back on QR. However, I think we know when things might fail so I'd think the options above are preferable.

As I argued in the thread above, I think the numerical error is often dominated by the statistical error and therefore tend to prefer the Cholesky because it's so much more faster. I hadn't realized the effect of the numerical rank determination that @bkamins mentioned, though.

bkamins:
Thank you for a fantastic summary. I have now checked the docstring of `dropcollinear`:

> dropcollinear controls whether or not lm accepts a model matrix which is less-than-full rank. If true (the default),      
> only the first of each set of linearly-dependent columns is used. The coefficient for redundant linearly dependent        
> columns is 0.0 and all associated statistics are set to NaN.

If I read it correctly we should drop columns when `rank(X) < size(X, 2)`. The problem is that with `tol=-1` in Cholesky this is not ensured. Are you aware if if we set `tol=0` we get such equivalence?

docs/src/examples.md
nalimilan: You can also filter out some values from the doctest, see https://documenter.juliadocs.org/stable/man/doctests/#Filtering-Doctests. That ensure that at least the code runs. Not sure how easy it would be to use here, maybe just something like 
```
```jldoctest; filter = [r"^\(Intercept\).*$", r"^x.*$"]
```
would do?

nalimilan:
How about merging the PR without changing the default nor adding `:auto`, and discuss that issue later? We don't need to make a decision about the default to add support for QR as an option.

test/runtests.jl
```diff
@@ -180,6 +180,25 @@ end
     @test GLM.dispersion(lm_model) ≈ GLM.dispersion(lm_qr_model)
 end
 
+@testset "Linear model with Cholesky dropcollinearity" begin
```
nalimilan, line 183: Since this testset is identical to the one below for QR, why not use a single block, with a `for method in (:qr, :cholesky)`? Same for other sets.

bkamins:
> We don't need to make a decision about the default to add support for QR as an option.

I think we can do this, but we would need to document how QR differs from Cholesky. The point is that in examples we have it does differ because different rule for detection of multi-collinearity is used (i.e. Cholesky works without a problem, but has set `tol=-1` which makes it decide that near-multicolinearity means dropping columns).

What I would want to avoid is leaving an impression that the difference is between QR and Cholesky as if in Cholesky we would add `tol=0` option we would also get the same results.

For these reasons I asked @mousum-github if we have an example where indeed QR does produce a qualitatively different result than Cholesky on numerical level (not at rank determination level). I think there will be such cases, but I just wanted to make sure we understand clearly what the options are, what they mean and how to interpret them.

For example in the case from the changed docs we get:

```
julia> using GLM

julia> y = [1, 2, 3, 4, 5];

julia> x = [1.0E-12, 2.0E-12, 3.0E-12, 4.0E-12, 5.0E-12];

julia> mdl1 = lm(@formula(y ~ x), (; x, y); dropcollinear=false)
StatsModels.TableRegressionModel{LinearModel{GLM.LmResp{Vector{Float64}}, GLM.DensePredChol{Float64, LinearAlgebra.Cholesky{Float64, Matrix{Float64}}}}, Matrix{Float64}}

y ~ 1 + x

Coefficients:
────────────────────────────────────────────────────────────────────────────────────────────────
                    Coef.   Std. Error                    t  Pr(>|t|)     Lower 95%    Upper 95%
────────────────────────────────────────────────────────────────────────────────────────────────
(Intercept)  -7.94411e-16  4.30465e-16                -1.85    0.1622  -2.16434e-15  5.75521e-16
x             1.0e12       0.00012979   7704748539664896.00    <1e-99   1.0e12       1.0e12
────────────────────────────────────────────────────────────────────────────────────────────────
```
and the result is different, but I could not say that worse than the one for QR reported in the docs.

nalimilan:
I see. The most consistent approach would probably be to use QR by default (which implies `tol=0` IIUC), use `tol=0` by default for Cholesky when `method=:cholesky` is passed, and add an argument to set a different tolerance (or maybe that value can be passed to `dropcollinear`?). Or support two `:cholesky` variants, one which sets `tol=0` and one which sets `tol=-1`. Not sure that's user-friendly though.

mousum-github:
An example where QR method performs better than Choleskly in estimation.
```
julia> Random.seed!(12345);

julia> X = rand(10, 2);

julia> Q,R = qr(X);

julia> R[1,1] = 1
1

julia> R[2,2] = 1.0E-7
1.0e-7

julia> X = Q*R;

julia> beta = [5, 10]
2-element Vector{Int64}:
  5
 10

julia> y = X*beta;

julia> md11 = lm(X, y; method=:qr, dropcollinear=false)
LinearModel

Coefficients:
───────────────────────────────────────────────────────────────────
    Coef.  Std. Error             t  Pr(>|t|)  Lower 95%  Upper 95%
───────────────────────────────────────────────────────────────────
x1    5.0  2.53054e-8  197586428.62    <1e-63        5.0        5.0
x2   10.0  1.59395e-8  627370993.89    <1e-99       10.0       10.0
───────────────────────────────────────────────────────────────────


julia> md21 = lm(X, y; method=:cholesky, dropcollinear=false)
LinearModel

Coefficients:
────────────────────────────────────────────────────────────────
       Coef.  Std. Error       t  Pr(>|t|)  Lower 95%  Upper 95%
────────────────────────────────────────────────────────────────
x1   5.28865   0.103248    51.22    <1e-10    5.05056    5.52674
x2  10.1818    0.0650348  156.56    <1e-14   10.0318    10.3318
────────────────────────────────────────────────────────────────
```
Whenever we set 
```
R[2,2] = 1.0E-04
```
both generate the same estimates.
Whenever we set 
```
R[2,2] = 1.0E-08
```
Cholesky throws PosDefException.
Whenever we set 
```
R[2,2] = 1.0E-016
```
Multicollinearity is detected in QR also.

bkamins:
OK. Thank you! So - @mousum-github - what do you feel would be a good option given the discussion we had?

ayushpatnaikgit:
@bkamins I have sent you an invitation for write access to xKDR/GLM.jl
In case you want to introduce changes to the PR directly. There was no other way, as the PR is made by an organisation.

docs/src/examples.md
mousum-github: Even after using filters, I am unable to pass the doctest.  One of the reasons is, not only estimates are different, but the number of `-` (or the position of `Coeff.`) is also changed depending on estimates.
 I have changed the example.

test/runtests.jl
mousum-github: Updated most of the test cases.

mousum-github:
Hi @bkamins,
In my opinion let's have `:cholesky` as default since the Cholesky decomposition method is not as bad as we usually think. I started with Julia GLM creating hundreds of datasets from different distributions and link functions to test and be familiar with this module. I have tested with very large datasets, especially to test speed. Most of the time the Cholesky and R's QR gave the same results. In a very few cases where results differ, because of singularity detection. Most of the issues raised in GLM about the estimates are because of singularity detection.
Additionally, as we discussed, we can provide (now or later) the `:auto` option to switch from `:cholesky` to `:qr` automatically whenever a singularity is detected by the Cholesky decomposition.

bkamins:
I think it is OK. We just need to document that `:cholesky` eagerly drops near-collinear columns so if a user has such case and does not want them to be dropped `:qr` should be used.

bkamins:
@mousum-github - I can review this PR, but please note that I am not a maintainer of GLM.jl, so someone from the team working with this package should approve and merge this PR.

docs/src/examples.md
```diff
@@ -86,6 +86,63 @@ julia> round.(vcov(ols); digits=5)
   0.38889  -0.16667
  -0.16667   0.08333
 ```
+By default, the `lm` method uses the Cholesky factorization which is known as fast but numerically unstable, especially for ill-conditioned design matrices. You can use the `method` keyword argument to apply a more stable QR factorization method.
```
bkamins, line 89: I think it would be more accurate to also state that with Cholesky factorization near-collinear columns are dropped (as this is the most common issue, the case where Cholesky is not accurate enough also can happen but is much rarer).

docs/src/examples.md
```diff
@@ -86,6 +86,63 @@ julia> round.(vcov(ols); digits=5)
...
+
+Coefficients:
+───────────────────────────────────────────────────────────────────
+    Coef.  Std. Error             t  Pr(>|t|)  Lower 95%  Upper 95%
+───────────────────────────────────────────────────────────────────
+x1    5.0  2.53054e-8  197586428.62    <1e-63        5.0        5.0
+x2   10.0  1.59395e-8  627370993.89    <1e-99       10.0       10.0
+───────────────────────────────────────────────────────────────────
+
+julia> md2 = lm(X, y; method=:cholesky, dropcollinear=false)
```
bkamins, line 135: do you understand why when `dropcollinear=false` is dropped `:cholesky` produces a different result? Does `:qr` produce the same or different result?

src/linpred.jl
```diff
@@ -35,45 +35,125 @@ end
...
+
+    function DensePredQR(X::AbstractMatrix, beta0::AbstractVector, pivot::Bool=false)
         n, p = size(X)
         length(beta0) == p || throw(DimensionMismatch("length(β0) ≠ size(X,2)"))
-        new{T}(X, beta0, zeros(T,p), zeros(T,p), qr(X))
+        T = typeof(float(zero(eltype(X))))
+        Q = pivot ? QRPivoted : QRCompactWY
+        fX = float(X)
+        cfX = fX === X ? copy(fX) : fX
+        F = pivot ? pivoted_qr!(copy(cfX)) : qr!(cfX)
```
bkamins, line 64: it seems we perform `copy` twice in some cases, where one copy would be enough - right?

src/linpred.jl
```diff
@@ -35,45 +35,125 @@ end
...
+        new{T,Q}(Matrix{T}(X),
+            Vector{T}(beta0),
+            zeros(T, p),
+            zeros(T, p),
+            F,
+            similar(X, T))
     end
-    function DensePredQR{T}(X::Matrix{T}) where T
+    function DensePredQR(X::AbstractMatrix, pivot::Bool=false)
```
bkamins, line 72:
```suggestion

    function DensePredQR(X::AbstractMatrix, pivot::Bool=false)
```

src/linpred.jl
```diff
@@ -35,45 +35,125 @@ end
...
-
 """
     delbeta!(p::LinPred, r::Vector)
 
 Evaluate and return `p.delbeta` the increment to the coefficient vector from residual `r`
 """
 function delbeta! end
 
-function delbeta!(p::DensePredQR{T}, r::Vector{T}) where T<:BlasReal
+function delbeta!(p::DensePredQR{T,<:QRCompactWY}, r::Vector{T}) where T<:BlasReal
```
bkamins, line 92: are we sure that all `delbeta!` methods are tested (this repo does not coverage tests, so that is why I am asking).

mousum-github: We have replicated almost all test cases for QR method. And I can recall we have changed something in all 4 `delbeta!` functions while testing. But yes, it does not say 100% code coverage. @ayushpatnaikgit let's have a discuss on this.

src/linpred.jl
mousum-github: Yes, I missed it. Updating the code.

docs/src/examples.md
mousum-github:
```
julia> using LinearAlgebra, GLM

julia> X = [-0.4011512997627107 0.6368622664511552;
                   -0.0808472925693535 0.12835204623364604;
                   -0.16931095045225217 0.2687956795496601;
                   -0.4110745650568839 0.6526163576003452;
                   -0.4035951747670475 0.6407421349445884;
                   -0.4649907741370211 0.7382129928076485;
                   -0.15772708898883683 0.25040532268222715;
                   -0.38144358562952446 0.6055745630707645;
                   -0.1012787681395544 0.16078875117643368;
                   -0.2741403589052255 0.4352214984054432];

julia> y = X * [5, 10];

julia> md1 = lm(X, y; method=:qr, dropcollinear=false)
LinearModel

Coefficients:
───────────────────────────────────────────────────────────────────
    Coef.  Std. Error             t  Pr(>|t|)  Lower 95%  Upper 95%
───────────────────────────────────────────────────────────────────
x1    5.0  2.53054e-8  197586428.62    <1e-63        5.0        5.0
x2   10.0  1.59395e-8  627370993.89    <1e-99       10.0       10.0
───────────────────────────────────────────────────────────────────


julia> md1 = lm(X, y; method=:qr, dropcollinear=true)
LinearModel

Coefficients:
────────────────────────────────────────────────────────────────────
    Coef.  Std. Error              t  Pr(>|t|)  Lower 95%  Upper 95%
────────────────────────────────────────────────────────────────────
x1    5.0  7.70812e-9   648666882.71    <1e-99        5.0        5.0
x2   10.0  4.85524e-9  2059629237.94    <1e-99       10.0       10.0
────────────────────────────────────────────────────────────────────


julia> md1 = lm(X, y; method=:cholesky, dropcollinear=false)
LinearModel

Coefficients:
────────────────────────────────────────────────────────────────
       Coef.  Std. Error       t  Pr(>|t|)  Lower 95%  Upper 95%
────────────────────────────────────────────────────────────────
x1   5.28865   0.103248    51.22    <1e-10    5.05056    5.52674
x2  10.1818    0.0650348  156.56    <1e-14   10.0318    10.3318
────────────────────────────────────────────────────────────────


julia> md1 = lm(X, y; method=:cholesky, dropcollinear=true)
LinearModel

Coefficients:
───────────────────────────────────────────────────────────────
       Coef.  Std. Error      t  Pr(>|t|)  Lower 95%  Upper 95%
───────────────────────────────────────────────────────────────
x1   5.64706    0.23454   24.08    <1e-08    5.10621    6.18791
x2  10.4076     0.147734  70.45    <1e-11   10.0669    10.7482
───────────────────────────────────────────────────────────────
```

bkamins: This is what I expected. So we get a different result in all four cases. Do you understand why?

Or maybe one of the maintainers of GLM.jl knows? I would expect that if we do not drop anything then `dropcollinear` setting should not have any impact on results.

CC @nalimilan @andreasnoack

mousum-github: Unlike Cholesky, estimates are not changed for QR due to pivoting. Although, there are (not very significant) mismatch in std error. 

 The difference in factorization ( and hence the difference in inverse of X'X) with Cholesky might be the reason the difference in estimates.
```
julia> GLM.inverse(md11.pp)
2×2 Matrix{Float64}:
 2.52043e14  1.58759e14
 1.58759e14  1.0e14

julia> GLM.inverse(md12.pp)
2×2 Matrix{Float64}:
 2.52043e14  1.58759e14
 1.58759e14  1.0e14

julia> GLM.inverse(md21.pp)
2×2 Matrix{Float64}:
 2.57977e14  1.62497e14
 1.62497e14  1.02355e14

julia> GLM.inverse(md22.pp)
2×2 Matrix{Float64}:
 2.64918e14  1.66868e14
 1.66868e14  1.05108e14

julia> GLM.inverse(md11.pp) ≈ GLM.inverse(md12.pp)
true

julia> GLM.inverse(md21.pp) ≈ GLM.inverse(md22.pp)
false
```

mousum-github: More over, at least in this example, Cholesky perform better (more close to original, and less standard error) than pivoting Cholesky.

bkamins: Yes, but why do we have differences in standard errors under `:qr`? This is what I do not understand. (we should resolve this in this PR I think, or at least understadnd the reason)

Also why do we have a different factorization in `:cholesky` if we in the end keep the variables (this is a question unrelated to this PR if we do not know).

bkamins: Is the difference in both cases beause we do/do not to pivoting respectively?

mousum-github: > Yes, but why do we have differences in standard errors under `:qr`? This is what I do not understand. (we should resolve this in this PR I think, or at least understadnd the reason)

The reason is the differences in corresponding estimated dispersion/deviance and these are calculated from residuals 
 ```
julia> GLM.dispersion(md11)
1.5939531933641094e-15

julia> GLM.dispersion(md12)
4.855242786854203e-16

julia> deviance(md11)
2.0325494261085132e-29

julia> deviance(md12)
1.8858706015439813e-30

julia> residuals(md11)
10-element Vector{Float64}:
 -1.7763568394002505e-15
 -3.3306690738754696e-16
 -1.5543122344752192e-15
 -1.7763568394002505e-15
 -1.7763568394002505e-15
 -1.7763568394002505e-15
 -1.1102230246251565e-15
 -1.7763568394002505e-15
 -8.881784197001252e-16
  0.0

julia> residuals(md12)
10-element Vector{Float64}:
  0.0
 -1.1102230246251565e-16
 -2.220446049250313e-16
 -8.881784197001252e-16
  0.0
  0.0
  2.220446049250313e-16
  8.881784197001252e-16
  0.0
  4.440892098500626e-16
```

Comment (bkamins) [reply docs/src/examples.md:135]: But why do we have different residuals if we have the same estimates of coefficients?

Comment (mousum-github) [reply docs/src/examples.md:135]: Actually,
```
julia> coef(md11)
2-element Vector{Float64}:
 4.999999976882807
 9.999999985438782

julia> coef(md12)
2-element Vector{Float64}:
  5.0000000040879975
 10.000000002574977

julia> isapprox(coef(md11), coef(md12))
true

julia> isapprox(coef(md11), coef(md12); atol=1.0e-16)
false

julia> isapprox(coef(md11), coef(md12); atol=1.0e-8)
false

julia> isapprox(coef(md11), coef(md12); atol=1.0e-7)
true
```

File: src/linpred.jl (hunk 29 above)
Comment (andreasnoack) [reply src/linpred.jl:92]: Not sure why the coverage isn't being reported here in the PR but it can be seen at https://app.codecov.io/gh/JuliaStats/GLM.jl/pull/507/blob/src/linpred.jl and indeed it looks like the `delbeta!` methods are covered. It looks like there is an uncovered constructor, though, https://app.codecov.io/gh/JuliaStats/GLM.jl/pull/507/blob/src/linpred.jl#L57

File: docs/src/examples.md (hunk 25 above)
Comment (mousum-github) [reply docs/src/examples.md]: I have updated the example. Now there are two examples, aggressive multicollinearity detection and unstable estimates.

File: src/linpred.jl (hunk 29 above)
Comment (andreasnoack) [reply src/linpred.jl:95]: It looks we now get the coverage info for the PR in-line

File: docs/src/examples.md (hunk 26 above)
Comment (bkamins) [reply docs/src/examples.md:134]: I am digging into it since I wanted to know the exact reason. I think it is OK that they differ given the reason.

File: src/linpred.jl [hunk 30]
```diff
@@ -35,45 +35,128 @@ end
...
     X::Matrix{T}                  # model matrix
     beta0::Vector{T}              # base coefficient vector
     delbeta::Vector{T}            # coefficient increment
     scratchbeta::Vector{T}
-    qr::QRCompactWY{T}
-    function DensePredQR{T}(X::Matrix{T}, beta0::Vector{T}) where T
+    qr::Q
+    scratchm1::Matrix{T}
+
+    function DensePredQR(X::AbstractMatrix, beta0::AbstractVector, pivot::Bool=false)
```
andreasnoack, line 57: It looks like this method isn't used at all so maybe we should just get rid of it.

docs/src/examples.md
mousum-github: Thanks @andreasnoack. we planned to have two examples where 

1. Cholesky detects multicollinearity eagerly like Wooldridge's example
2. QR estimates better in terms of precision 

There are many examples and issues related to Point 1.  After a lot of trial and error, finally able to create a dataset that reflects Point 2 (https://github.com/JuliaStats/GLM.jl/pull/507#issuecomment-1488860570). 

As you mentioned, I also thought to add some noise to `y` then we will have a decent standard error of estimates but estimates will differ from [ 5, 10].  Now I am thinking that we can compare the models based on likelihood values.

In that cases, the updated example will be 

```
julia> y = X*[5, 10] + 1.0e-08rand(10);

julia> mdqr = lm(X, y; method=:qr, dropcollinear=false)
LinearModel

Coefficients:
───────────────────────────────────────────────────────────────
      Coef.  Std. Error       t  Pr(>|t|)  Lower 95%  Upper 95%
───────────────────────────────────────────────────────────────
x1  4.98461   0.0571763   87.18    <1e-12    4.85276    5.11646
x2  9.99031   0.0360146  277.40    <1e-16    9.90726   10.0734
───────────────────────────────────────────────────────────────

julia> loglikelihood(mdqr)
181.34558612704268

julia> mdclol = lm(X, y; method=:cholesky, dropcollinear=false)
LinearModel

Coefficients:
───────────────────────────────────────────────────────────────
      Coef.  Std. Error       t  Pr(>|t|)  Lower 95%  Upper 95%
───────────────────────────────────────────────────────────────
x1  4.71135   0.113578    41.48    <1e-09    4.44944    4.97326
x2  9.81818   0.0715412  137.24    <1e-14    9.65321    9.98316
───────────────────────────────────────────────────────────────

julia> loglikelihood(mdchol)
174.59847167533468
```
@bkamins - looking for your opinion. 

Also, If we want to add Wooldridge's example, is there any better way to access the dataset other than putting the `csv` file in the data folder? - Please suggest.

src/linpred.jl
mousum-github: Thanks @andreasnoack, I will be removing the constructor.

mousum-github: Updated the code and pushed.

docs/src/examples.md
```diff
@@ -155,39 +155,50 @@ x2  10.1112    0.053489   189.03    <1e-15    9.98781   10.2345
...
-```
-julia> y = [1, 2, 3, 4, 5];
+Since the Cholesky method aggressively detect multicollinearity, if you ever encounter multicollinearity in any GLM model with Cholesky, it is worth trying the same model with QR decomposition. The following example is taken from `Introductory Econometrics: A Modern Approach, 7e" by Jeffrey M. Wooldridge`. The dataset is used studying the relationship between firm size—often measured by annual sales—and spending on research and development (R&D). The following shows that for the given model, the Cholesky method detects multicollinearity in the design matrix and hence can't estimate properly whereas the QR method estimates correctly.
```
bkamins, line 158: This is with `dropcollinear=true` (i.e. when you turn on autodetection of multi-collinearity. If you would pass `dropcollinear=false` Cholesky gives a correct result.

andreasnoack: @bkamins do you think we are good to go here once this change is in? I don't have further comments.

bkamins: Yes. My point is from the user's perspective (I do not know all the design details of GLM.jl internals, but based on my checks the iimplementation is correct). And as a user it is important for me to know that there are two important keyword arguments:
* `method` (`:cholesky` by default);
* `dropcollinear` (`true` by default).

Then the default values can lead to problems in near multi-colinearity case and depending on nature of your data you can set `method=:qr` or `dropcollinear=false` or both (and I guess it is hard to give hard rules when to do what).

andreasnoack: Just to be sure. Are you proposing that the paragraph here explicitly mentions the `dropcollinear` argument? At this point, maybe we should get the current version in and then delegate further docstring improvements to a separate PR?

bkamins: I am OK to have documentation improvements in a separate PR. My point is that ultimately we should be explicit how things work as without a proper explanation people might think that `:cholesky` fails in some cases, while it does work as expected then (just does aggresive near-collinearity detection, which can be disabled by `dropcollinear=false`). Of course there are cases when indeed `:cholesky` fails and only `:qr` works, but as it was discussed above they are much rarer than cases when `:cholesky` works, but the rule for collinearity detection we use is triggered.

bkamins: In particular `dropcollinear` kwarg means a bit different thing when we choose `:cholesky` and when we choose `:qr`.

mousum-github: from our previous example
https://github.com/JuliaStats/GLM.jl/pull/507#issuecomment-1488860570

```
julia> Random.seed!(12345); X = rand(10, 2); Q,R = qr(X); R[1,1] = 1; beta = [5, 10];

julia> Random.seed!(12345); R[2,2] = 1.0E-8; X = Q*R; y = X*beta + 10.0E-08rand(10);

julia> md11 = lm(X, y; method=:qr, dropcollinear=false)
LinearModel

Coefficients:
──────────────────────────────────────────────────────────────
       Coef.  Std. Error     t  Pr(>|t|)  Lower 95%  Upper 95%
──────────────────────────────────────────────────────────────
x1   5.12206     3.05769  1.68    0.1324   -1.92899    12.1731
x2  10.0769      1.926    5.23    0.0008    5.63552    14.5183
──────────────────────────────────────────────────────────────

julia> md12 = lm(X, y; method=:qr, dropcollinear=true)
LinearModel

Coefficients:
──────────────────────────────────────────────────────────────
       Coef.  Std. Error     t  Pr(>|t|)  Lower 95%  Upper 95%
──────────────────────────────────────────────────────────────
x1   5.12206     3.05769  1.68    0.1324   -1.92899    12.1731
x2  10.0769      1.926    5.23    0.0008    5.63552    14.5183
──────────────────────────────────────────────────────────────

julia> md22 = lm(X, y; method=:cholesky, dropcollinear=true)
LinearModel

Coefficients:
───────────────────────────────────────────────────────────────────────
      Coef.    Std. Error             t  Pr(>|t|)  Lower 95%  Upper 95%
───────────────────────────────────────────────────────────────────────
x1  0.0      NaN                 NaN       NaN     NaN        NaN
x2  6.85056    1.32933e-8  515340074.98    <1e-99    6.85056    6.85056
───────────────────────────────────────────────────────────────────────

julia> md21 = lm(X, y; method=:cholesky, dropcollinear=false)
ERROR: PosDefException: matrix is not positive definite; Cholesky factorization failed.
```
- setting R[2,2] = 1.0E-08 till R[2,2] = 1.0E-14, both Cholesky (dropcollinear = true or false) detect multicollinearity whereas both QR do not detect.
- setting R[2,2] = 1.0E-07, Cholesky with dropcollinear = true detects multicollonearity, but Cholesky with dropcollinear = false does not detect like QR.
- setting R[2,2] = 1.0E-15, both QR (dropcollinear = true or false)  detect multicollinearity.

mousum-github: @bkamins - does it sound better, if I change the paragraph to the following?

``
Since the Cholesky, especially with `dropcollinear = true`, method aggressively detects multicollinearity, if you ever encounter multicollinearity in any GLM model with Cholesky, it is worth trying the same model with QR decomposition. The following example is taken from `Introductory Econometrics: A Modern Approach, 7e" by Jeffrey M. Wooldridge`. The dataset is used to study the relationship between firm size—often measured by annual sales—and spending on research and development (R&D). The following shows that for the given model, the Cholesky method detects multicollinearity in the design matrix and hence can't estimate all parameters whereas the QR method can.
``

bkamins: I would propose to change the last sentence to:

> the Cholesky method detects multicollinearity in the design matrix with dropcollinear=true and hence does not estimate all parameters as opposed to QR.

The point is that with `dropcollinear=false` Cholesky would produce the expected estimates so we cannot say that it "can't estimate all parameters" in general (as in this case it could, but with different settings)

bkamins: > from our previous example

Yes - I agree. That is why I say that the `method/dropcollinear` has four combinations and different cases are in general possible (especially because `dropcollinear` turns on/off pivoting which affects the results in some cases)

mousum-github: Thanks @bkamins, I have updated the text as you suggested.

docs/src/examples.md
```diff
@@ -155,7 +155,7 @@ x2  10.1112    0.053489   189.03    <1e-15    9.98781   10.2345
...
+Since the Cholesky, especially with `dropcollinear = true`, method aggressively detects multicollinearity, if you ever encounter multicollinearity in any GLM model with Cholesky, it is worth trying the same model with QR decomposition. The following example is taken from `Introductory Econometrics: A Modern Approach, 7e" by Jeffrey M. Wooldridge`. The dataset is used to study the relationship between firm size—often measured by annual sales—and spending on research and development (R&D). The following shows that for the given model, the Cholesky method detects multicollinearity in the design matrix with dropcollinear=true and hence does not estimate all parameters as opposed to QR.
```
bkamins, line 158:
```suggestion
Since the Cholesky method with `dropcollinear = true` aggressively detects multicollinearity,
if you ever encounter multicollinearity in any GLM model with Cholesky,
it is worth trying the same model with QR decomposition.
The following example is taken from `Introductory Econometrics: A Modern Approach, 7e" by Jeffrey M. Wooldridge`.
The dataset is used to study the relationship between firm size—often measured by annual sales—and spending on
research and development (R&D).
The following shows that for the given model,
the Cholesky method detects multicollinearity in the design matrix with `dropcollinear=true`
and hence does not estimate all parameters as opposed to QR.
```

bkamins (approved):

ararslan:
For whatever it's worth, I think the use of `method` to choose the solution method is more appropriate than using `method` to choose where a computation is performed. The `method` keyword here in GLM.jl is the same as in R's `lm`. I don't know whether that was the motivator behind the name but it is pleasantly discoverable for folks coming from R.

PR #512, opened 2023-01-24
add compat for statsmodels 0.7

kleinschmidt:
We will tag 0.7 very soon (~~probably today~~ ahahaha good one...) with changes to `FunctionTerm` that are breaking but do not directly affect anything that GLM itself uses.  This PR should not be merged until 0.7 is tagged and CI can be run to ensure that nothing breaks in GLM, so I'll start this in draft mode and flip it out when it's time to check CI.

Update: 0.7 is now released.  I've added bumped the patch version here and also fixed a bug where we were defining our own `formula` function instead of creating a method for `StatsModels.formula`.

Update to teh update: based on #500, I re-based this on a commit before #339 was merged, targeting a "backport" branch called `v1`.

### DIFF
--- a/.github/workflows/CI-future.yml
+++ b/.github/workflows/CI-future.yml
@@ -6,9 +6,11 @@ on:
   pull_request:
     branches:
       - master
+      - v1
   push:
     branches:
       - master
+      - v1
     tags: '*'
 jobs:
   test:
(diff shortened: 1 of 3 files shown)


kleinschmidt:
Okay, I've double checked at least the ubuntu runs to make sure that statsmodels 0.7 was used and all tests are passing so I think this is good to go!

palday (approved):

kleinschmidt:
nightly failures are due to the "evaluated to nonboolean" change for `@test_broken`

PR #514, opened 2023-02-10
Set minimum Julia version to 1.6

ararslan:
Currently the minimum version is 1.0, which is old and unsupported and makes development and addition of new features more difficult. I suspect this won't be too controversial.

### DIFF
--- a/.github/workflows/CI-stable.yml
+++ b/.github/workflows/CI-stable.yml
@@ -19,7 +19,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        version: ['1.0', '1']
+        version: ['1.6', '1']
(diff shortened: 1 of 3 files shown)


bkamins (approved):
maybe add nightly to CI?

ararslan:
Nightly is already in CI as the "CI-future" jobs

bkamins:
ah - OK. In DataFrames.jl we have a bit different structure of CI file and I looked only at diff.

ararslan:
Yeah the CI structure for this package is kind of strange, could be worth standardizing it at some point.

ararslan:
I fixed the main CI failure on nightly, the nightly failure on Windows is unrelated and related to building TimeZones.

PR #529, opened 2023-05-02
bump compat for StatsBase 0.34

palday:


ararslan:
Looks like CI ran with StatsBase 0.33 because something else is upper-bounding it.

ararslan:
Ah, it's StatsModels. So we'll need https://github.com/JuliaStats/StatsModels.jl/pull/291

kleinschmidt:
I think nightly failures are the same as we see in StatsModels: 

```
saturated linear model: Error During Test at /home/runner/work/GLM.jl/GLM.jl/test/runtests.jl:207
  Expression evaluated to non-Boolean
```

although looking at the tests, I don't see why (in my experience this is caused by `@test_broken` normally accepting non-boolean values in 1.8 or earlier)

ararslan:
We could backport https://github.com/JuliaStats/GLM.jl/pull/514 to the `v1` branch. That PR fixed that issue.

Issue #534, opened 2023-05-15
Feature request: Support for reuse of pinv computation for multiple Y values in linear models


yoninazarathy:
For `LinearModel`, it may be useful to have some sort of API which allows one to reapply the model to multiple right hand sides repeatedly (without redoing the internal QR or Cholesky decomposition or other pinv computation again).

This can help performance when using the package for simulation with a fixed design matrix and multiple right hand sides. 

My current workaround for this is of the form:

```
model = fit(LinearModel, formula ... , data, ...)
model_design_matrix = model.mm.m
model_pseudo_inv_matrix = pinv(model_design_matrix) #from LinearAlgebra

#now when looping on many response vectors y:
coef = model_pseudo_inv_matrix *y
...
```
With this one can clearly just carry out the complete `LinearModel` fit just using elementary `LinearAlgebra` operations, still the support that `GLM.jl` gives to data frames, contrasts, and model queries is useful.

andreasnoack:
I think this is almost already supported. With https://github.com/JuliaStats/GLM.jl/pull/535, I can write
```julia
julia> X = [ones(10) randn(10)];

julia> y1 = X*ones(2) + randn(10)*0.1;

julia> y2 = X*ones(2) + randn(10)*0.1;

julia> myp = GLM.DensePredChol(X, false);

julia> myr1 = GLM.LmResp(y1);

julia> myr2 = GLM.LmResp(y2);

julia> fit!(LinearModel(myr1, myp, nothing))
LinearModel

Coefficients:
───────────────────────────────────────────────────────────────
       Coef.  Std. Error      t  Pr(>|t|)  Lower 95%  Upper 95%
───────────────────────────────────────────────────────────────
x1  0.976384   0.0449526  21.72    <1e-07   0.872723    1.08004
x2  0.977983   0.0427468  22.88    <1e-07   0.879409    1.07656
───────────────────────────────────────────────────────────────


julia> fit!(LinearModel(myr2, myp, nothing))
LinearModel

Coefficients:
───────────────────────────────────────────────────────────────
       Coef.  Std. Error      t  Pr(>|t|)  Lower 95%  Upper 95%
───────────────────────────────────────────────────────────────
x1  0.995871   0.0278513  35.76    <1e-09   0.931646    1.0601
x2  0.965657   0.0264846  36.46    <1e-09   0.904583    1.02673
───────────────────────────────────────────────────────────────
```

yoninazarathy:
Thank you.

PR #535, opened 2023-05-15
Remove installbeta! and make linear model fit insensitive to beta0 state

andreasnoack:
When looking at https://github.com/JuliaStats/GLM.jl/issues/534 I noticed that `fit!(::LinearModel)` depended on the state of `beta0` stored in the linear predictor. The reason was that the solution stored in `delbeta` was transferred to `beta0` by the `installbeta!` method which updated the current value. This is important for GLMs but not LMs. Furthermore, the operations are so simple to do with broadcasting that I think we should rather just do that. (I think `installbeta!` might predate broadcasting in Julia.)

### DIFF
--- a/docs/src/api.md
+++ b/docs/src/api.md
@@ -64,7 +64,6 @@ fit
 StatsBase.deviance
 GLM.dispersion
 GLM.ftest
-GLM.installbeta!
 StatsBase.nobs
 StatsBase.nulldeviance
 StatsBase.predict
--- a/src/deprecated.jl
+++ b/src/deprecated.jl
@@ -19,3 +19,6 @@ function Base.getproperty(mm::LinPredModel, f::Symbol)
         return getfield(mm, f)
     end
 end
+
+@deprecate installbeta!(p) (p.beta0 .= p.delbeta) false
+@deprecate installbeta!(p, f) (p.beta0 .+= p.delbeta .* f) false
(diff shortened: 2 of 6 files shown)


andreasnoack:
To give an example of the issue with the current behavior
```julia
julia> X = [ones(10) randn(10)];

julia> y = X*ones(2) + randn(10)*0.1;

julia> myp = GLM.DensePredChol(X, false);

julia> myr = GLM.LmResp(y);

julia> coef(fit!(LinearModel(myr, myp, nothing)))
2-element Vector{Float64}:
 0.966074297502665
 1.0104887004807501

julia> coef(fit!(LinearModel(myr, myp, nothing)))
2-element Vector{Float64}:
 1.93214859500533
 2.0209774009615002

julia> coef(fit!(LinearModel(myr, myp, nothing)))
2-element Vector{Float64}:
 2.898222892507995
 3.03146610144225

julia> coef(fit!(LinearModel(myr, myp, nothing)))
2-element Vector{Float64}:
 3.86429719001066
 4.0419548019230005
```

docs/src/api.md
```diff
@@ -64,7 +64,6 @@ fit
 StatsBase.deviance
 GLM.dispersion
 GLM.ftest
-GLM.installbeta!
```
ararslan, line 67: If this was documented as part of the API, should we add a simple deprecation like
```julia
@deprecate installbeta!(p) (p.beta0 .= p.delbeta) false
@deprecate installbeta!(p, f) (p.beta0 .= p.delbeta .* f) false
```
?

andreasnoack: I thought it might be okay since we are preparing for a breaking change anyway and I was not envisioning that anybody would be using it but [I was wrong](https://juliahub.com/ui/Search?q=installbeta&type=code). All add some deprecations.

ararslan: 😬

ararslan (approved):
Windows CI failure is unrelated

PR #536, opened 2023-06-07
set version to 2.0.0-DEV

palday:
There are changes currently on `master` that we have determined are Technically Breaking. Setting the version to 2.0.0-DEV indicates the next version that will be released from master and that we're not yet in the state we want for that release.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "GLM"
 uuid = "38e38edf-8417-5370-95a0-9cbb8c7f171a"
-version = "1.8.1"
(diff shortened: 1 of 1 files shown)


kleinschmidt (approved):
I think that communicating this is important somehow (I've made a mistake before because I didn't understand that master was meant to be 2.0).  I'm not 100% sure this is the way to do it but it does have the advantage that when someone goes to bump the version in PR, they will notice that it's 2.0.0-DEV, so I don't see a better/safer way.

nalimilan (approved):

PR #499, opened 2022-09-20
add residuals method for GLM

palday:
~I haven't had the time to figure out how the denominator is computed for Pearson residuals and so haven't yet implemented them.~

I set the default residual type to `deviance` to match R's behavior, though that might be surprising for GLMs with normal distribution and identity link.

Note: once again I've noticed that `-0` in outputs creates a problem. Tests fail for me on Julia 1.8 with Apple Silicon because of sign-swapping on zero.

PR #540, opened 2023-06-12
Residuals in GLM

mousum-github:
I have added most of the test cases as discussed in #499.

Also, I have changed the residuals function for `lm`. Now the residuals function for `lm` is similar to residuals function in 'glm'.
Although, I am not sure about applicability of `:working` type residuals in `lm`. `R` supports `:working` type residuals in `lm` and which is same as `:response` type.


ViralBShah:
This PR seems to be picking up some older commits and needs cleaning up.

mousum-github:
Thanks, Viral.
I will check it this week.

mousum-github:
The PR is now cleaned up and merged with the latest master branch.

src/GLM.jl
```diff
@@ -10,6 +10,7 @@ module GLM
     import Base: (\), convert, show, size
     import LinearAlgebra: cholesky, cholesky!
     import Statistics: cor
+    using StatsAPI
```
bkamins, line 13: it should be in `using` part

src/linpred.jl
```diff
@@ -360,9 +360,21 @@ response(obj::LinPredModel) = obj.rr.y
 
 fitted(m::LinPredModel) = m.rr.mu
 predict(mm::LinPredModel) = fitted(mm)
-residuals(obj::LinPredModel) = residuals(obj.rr)
 
-function formula(obj::LinPredModel)
+function residuals(model::LinPredModel; type=:deviance)
```
bkamins, line 364: I do not know the details of the implementation, but is the `residuals` method properly documented (as it is not documented here). I.e. that it supports `type` kwarg and what is the behavior depending on the value of the kwarg.

jerlich:
what's the status of this PR? Can it get a bit of love?

src/linpred.jl
jerlich: It is documented on line 847 of glmfit.jl

src/linpred.jl
```diff
@@ -360,9 +360,21 @@ response(obj::LinPredModel) = obj.rr.y
...
+    type in _RESIDUAL_TYPES ||
+        throw(ArgumentError("Unsupported type `$(type)``; supported types are" *
```
nalimilan:
```suggestion
        throw(ArgumentError("Unsupported type `$(type)`; supported types are " *
```

PR #552, opened 2024-01-07
Let future test on beta instead of nightly

andreasnoack:
Early in the cycle, the failures on nightly are not that useful to us. We only really need to pay attention once Julia enters the beta stage.


andreasnoack:
Hm. I thought there was a general "beta channel" but it looks like it has to be a specific version so closing this one.

Issue #558, opened 2024-05-05
GLM fails to match certified values for the Filip dataset 


mousum-github:
[filip.csv](https://github.com/JuliaStats/GLM.jl/files/15214574/filip.csv)

`In response to industrial concerns about the numerical accuracy of computations from statistical software, the Statistical Engineering and Mathematical and Computational Sciences Divisions of NIST's Information Technology Laboratory are providing datasets with certified values for various statistical methods`.
One of them for linear regression is the Filip dataset.
https://www.itl.nist.gov/div898/strd/lls/data/Filip.shtml

For this dataset, the `lm` function with the `QR` decomposition method fails to estimate all parameters
```
LinearModel

y ~ 1 + x + :(x ^ 2) + :(x ^ 3) + :(x ^ 4) + :(x ^ 5) + :(x ^ 6) + :(x ^ 7) + :(x ^ 8) + :(x ^ 9) + :(x ^ 10)

Coefficients:
────────────────────────────────────────────────────────────────────────────────────────
                    Coef.     Std. Error       t  Pr(>|t|)      Lower 95%      Upper 95%
────────────────────────────────────────────────────────────────────────────────────────
(Intercept)   8.13378        9.54815        0.85    0.3971  -10.9001        27.1677
x             0.0          NaN            NaN       NaN     NaN            NaN
x ^ 2        -7.14418       14.9533        -0.48    0.6343  -36.953         22.6647
x ^ 3        -4.53337       14.5308        -0.31    0.7560  -33.5           24.4333
x ^ 4        -0.881151       6.84421       -0.13    0.8979  -14.5248        12.7625
x ^ 5         0.135741       1.93585        0.07    0.9443   -3.7233         3.99479
x ^ 6         0.0989815      0.351359       0.28    0.7790   -0.601441       0.799403
x ^ 7         0.0207993      0.0413984      0.50    0.6169   -0.0617269      0.103326
x ^ 8         0.0022362      0.00307053     0.73    0.4688   -0.0038848      0.0083572
x ^ 9         0.000124681    0.000130509    0.96    0.3426   -0.000135484    0.000384846
x ^ 10.0      2.86391e-6     2.42701e-6     1.18    0.2419   -1.97424e-6     7.70207e-6
```
whereas the `lm` function with the `Cholesky` decomposition method fails to estimate 6 parameters
```
LinearModel

[...]

PR #559, opened 2024-05-07
Fix QR based fitting

andreasnoack:
Avoid erroring out for low rank design matrices when dropcollinear=false.
Avoid unnecessary triangular solves. Avoid indexing in the Q.
Avoid slicing R matrix in a way that triggers a minimum norm solution.
Remove unnecessary temporaries.

https://github.com/JuliaStats/GLM.jl/issues/558


src/linpred.jl
```diff
@@ -63,23 +63,19 @@ function delbeta! end
...
     rnk = rank(p.qr.R)
-    rnk == length(p.delbeta) || throw(RankDeficientException(rnk))
-    p.delbeta = p.qr\r
-    mul!(p.scratchm1, Diagonal(ones(size(r))), p.X)
+    p.delbeta = p.qr \ r
     return p
 end
 
 function delbeta!(p::DensePredQR{T,<:QRCompactWY}, r::Vector{T}, wt::Vector{T}) where T<:BlasReal
     rnk = rank(p.qr.R)
```
devmotion, line 71: `rnk` is not used anymore:
```suggestion
```

src/linpred.jl
```diff
@@ -88,44 +84,32 @@ function delbeta!(p::DensePredQR{T,<:QRPivoted}, r::Vector{T}) where T<:BlasReal
     if rnk == length(p.delbeta)
         p.delbeta = p.qr\r
```
devmotion, line 85: Maybe
```suggestion
        ldiv!(p.delbeta, p.qr, r)
```
?

src/linpred.jl
```diff
@@ -88,44 +84,32 @@ function delbeta!(p::DensePredQR{T,<:QRPivoted}, r::Vector{T}) where T<:BlasReal
...
-        R = @view qnr.R[1:rnk, 1:rnk]
-        Rinv = inv(R)
-        piv = qnr.p
-        permute!(delbeta, piv)
-        for k=(rnk+1):length(delbeta)
-            delbeta[k] = -zero(T)
-        end
-        p.delbeta[1:rnk] = Rinv * Rinv' * view(delbeta, 1:rnk)
-        invpermute!(delbeta, piv)
+    r̃ = sqrtW * r
```
devmotion, line 101: It seems the same approach as above could applied here?

src/linpred.jl
```diff
@@ -88,44 +84,32 @@ function delbeta!(p::DensePredQR{T,<:QRPivoted}, r::Vector{T}) where T<:BlasReal
...
+
+    p.qr = pivoted_qr!(copy(p.scratchm1))
```
devmotion, line 103: Is `scratchm1` reused in another function?

src/linpred.jl
```diff
@@ -88,44 +84,32 @@ function delbeta!(p::DensePredQR{T,<:QRPivoted}, r::Vector{T}) where T<:BlasReal
...
+    rnk = rank(p.qr.R) # FIXME! Don't use svd for this
+    R = UpperTriangular(view(parent(p.qr.R), 1:rnk, 1:rnk))
+    permute!(p.delbeta, p.qr.p)
+    for k = (rnk + 1):length(p.delbeta)
+        p.delbeta[k] = -zero(T)
```
devmotion, line 108: It seems that's copied from the current implementation - but why `-zero(T)` and not `zero(T)`?

src/linpred.jl
```diff
@@ -88,44 +84,32 @@ function delbeta!(p::DensePredQR{T,<:QRPivoted}, r::Vector{T}) where T<:BlasReal
...
     end
+    p.delbeta[1:rnk] = R \ (p.qr.Q'*r̃)[1:rnk]
```
devmotion, line 110: Maybe
```suggestion
    mul!(view(p.delbeta, 1:rnk), view(p.qr.Q, :, 1:rnk)', r̃)
    ldiv!(R, view(p.delbeta, 1:rnk))
```
?

src/linpred.jl
andreasnoack: Done

src/linpred.jl
andreasnoack: I don't think so, so I'll drop the `copy`

src/linpred.jl
andreasnoack: I'm not sure what the point is. Let's drop the minus.

src/linpred.jl
andreasnoack: We can't make a view into `Q`, see above, but we can make a view into the result.

andreasnoack:
@devmotion I finally found a way to download the test dataset reliably so this one should be good to go.

test/runtests.jl
```diff
@@ -1990,27 +1991,45 @@ end
...
+    @test gvif(gm2; scale=true) ≈ [1.486330, 2.301648, 1.502666] atol=1e-4
+
     # the VIF definition depends on modelmatrix, vcov and stderror returning valid
     # values. It doesn't care about links, offsets, etc. as long as the model matrix,
     # vcov matrix and stderrors are well defined.
 end
+
+@testset "NIST - Filip. Issue 558" begin
+    # Since "https://www.itl.nist.gov/div898/strd/lls/data/LINKS/DATA/Filip.dat" seems to block download, we'll use a "mirror"
+    fn = Downloads.download("https://gist.githubusercontent.com/andreasnoack/9dadfb922fc45ca52bc7b2659b7c5b67/raw/f1dcdd2179b606ecbb7e23316da1ede971e7ffe8/Filip.dat")
```
devmotion, line 2021: There's https://github.com/joshday/NISTStatisticalReferenceData.jl/ which however seems to use `Downloads.download` for downloading the datasets (and storing them in `deps/` seems a bit suboptimal).

devmotion (approved):

test/runtests.jl
andreasnoack: Maybe there will be some better way of doing this in the future. Then we can adjust.

Issue #509, opened 2022-11-16
About `dof_residual` 


gragusa:
I noticed something odd about `dof_residual`. In few words, when the `family` does not have a `dispersion_parameter`, `dof_residual`  is bumped by 1which is inconsistent with other software (R, Stata, etc.) and with statistics more generally :smile:. 

```julia
using DataFrames
using StableRNGs
using RCall

rng = StableRNG(123);

x1 = rand(rng, 25);
x2 = ifelse.(randn(rng, 25) .> 0, 1, 0);
y = ifelse.(0.004 .- 0.01 .* x1 .+ 1.5 .* x2 .+ randn(rng, 25) .> 0, 1, 0);
df = DataFrame(y=y, x1=x1, x2=x2);

## Julia
l = glm(@formula(y~x1+x2), df, Binomial());

## R
@rput df;
R"rl <- glm(y~x1+x2, df, family='binomial')";
@rget rl;

println("Julia dof_residual:", dof_residual(l))
println("R dof_residual: ", rl[:df_residual])

```
which gives 
```julia
Julia dof_residual: 23
R dof_residual: 22
```

The families without a `dispersion_parameter` are `Bernoulli, Binomial, Poisson`:
```
dispersion_parameter(D) = true
dispersion_parameter(::Union{Bernoulli, Binomial, Poisson}) = false
```

Fortunately, `dof_residuals` is not used to scale the `vcov` and so this bug does not have important ramifications.  I also noted that we are not testing `dof_residual` for these families.

The fix is relatively simple: change 

```julia
dof_residual(obj::LinPredModel) = nobs(obj) - dof(obj) + 1
```

to

```julia
dof_residual(obj::LinPredModel) = nobs(obj) - dof(obj) + dispersion_parameter(obj.rr.d)
```

nalimilan:
Good catch. Actually this definition seems to have been there forever, though the implementation has been changed recently by https://github.com/JuliaStats/GLM.jl/pull/265. We should probably add a docstring to explain more precisely what this method returns as it's not necessarily obvious for users how the dispersion parameter must be handled (R doesn't say anything).

Cc: @andreasnoack @Nosferican

alfa33333:
What is dof returning? I had a simple linear regression model which in principle, should have only one intercept and one predictor, but dof returns 3., although dof_residuals returns the right value. 

From the discussion, it seems to add a third parameter, the dispersion, but this is not clear from the documentation, especially for linear regression models.

[...]

PR #560, opened 2024-05-08
Change dof_residual definition to not depend on dispersion parameter

andreasnoack:
Fixes https://github.com/JuliaStats/GLM.jl/issues/509

There was actually already a test of a model affected by this but the reference value was wrong. I checked the value in R. 

@nalimilan how do you feel about `dof` after all these years? After preparing this, I think we might be better of from deprecating `dof` (which would have to happen in `StatsAPI`). I tend to think that it brings more confusion than convenience.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -17,7 +17,7 @@ StatsModels = "3eaba693-59b7-5ba5-a881-562e759f1c8d"
 Tables = "bd369af6-aec1-5ad0-b16a-f7cc5008161c"
 
 [compat]
-CSV = "0.7, 0.8"
+CSV = "0.7, 0.8, 0.9, 0.10"
 CategoricalArrays = "0.8, 0.9, 0.10"
 DataFrames = "0.22, 1"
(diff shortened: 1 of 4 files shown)


nalimilan (approved):
Sorry for the delay. Sounds reasonable to me to include this as part of 2.0 given that it's breaking.

Regarding `dof`, you mean we should deprecate it because people confuse it with `dof_residual`?

andreasnoack:
> Regarding `dof`, you mean we should deprecate it because people confuse it with `dof_residual`?

Yeah. The confusion seems to show up repeatedly and I think that `dof_residual` is all you really need.

nalimilan:
I see. But it can be useful in some cases, notably it's used in the definition of `aic`, `bic` and `adjr2`. In R you have to access private fields of model objects to get that information, which is bad.

Maybe we could put a warning in the docstring? Or we could have a single `dof` function with an argument to choose whether you want the consumed or residual DOF, but that would be very breaking.

PR #569, opened 2024-11-19
Update Documenter.yml

andreasnoack:
Use specific versions of actions and enable caching.

~This will probably fail for now. Looks like there are some changes to random numbers in Julia 1.11.~ This now uses a dataset from RDatasets instead of a random dataset. This change avoids the current failure on Julia 1.11 without braking 1.10. I also think it's better to use real datasets for examples. Furthermore, the PR now fixes several warnings currently present in the docs build.

### DIFF
--- a/.github/workflows/Documenter.yml
+++ b/.github/workflows/Documenter.yml
@@ -1,21 +1,33 @@
-name: Documentation
+name: Documenter
+
+permissions:
+  actions: write
+  contents: read
+
 concurrency:
   group: ${{ github.head_ref }}.docs
   cancel-in-progress: true
+
 on:
   push:
-    branches:
-      - 'master'
-    tags: '*'
+    branches: [master]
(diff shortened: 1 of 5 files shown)


.github/workflows/Documenter.yml
```diff
@@ -1,21 +1,24 @@
-name: Documentation
-concurrency:
-  group: ${{ github.head_ref }}.docs
-  cancel-in-progress: true
```
palday, line 4: I would keep concurrency in -- it cancels stale runs if you push again

palday:
@andreasnoack can you set `push_preview=true` in `docs/make.jl`?

andreasnoack:
> can you set push_preview=true in docs/make.jl?

Sure. I've also upgraded to Documenter@1 while visiting this

palday (approved):
thanks!

PR #579, opened 2024-11-21
Use Julia's standard debug loggin mechanism instead of the custom verbose flag.

andreasnoack:
Minor thing but I think it's nicer to use the default logging machinery for this.

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -256,3 +256,17 @@ of a ```LinPred``` type use a default step factor of 1.  The value of
 ```linpred``` is the argument to the ```updatemu``` method for
 ```ModResp``` types.  The ```updatemu``` method returns the updated
 deviance.
+
+## Debugging failed fits
+In the rare cases when a fit of a generalized linear model fails, it can be useful
+to enable more output from the fitting steps. This can be done through
+the Julia logging mechanism by setting `ENV[JULIA_DEBUG] = GLM`. Enabling debug output
+will result in ouput like the following
+```julia
+┌ Debug: Iteration: 1, deviance: 5.129147109764238, diff.dev.:0.05057195315968688
+└ @ GLM ~/.julia/dev/GLM/src/glmfit.jl:418
+┌ Debug: Iteration: 2, deviance: 5.129141077001254, diff.dev.:6.032762984276019e-6
+└ @ GLM ~/.julia/dev/GLM/src/glmfit.jl:418
+┌ Debug: Iteration: 3, deviance: 5.129141077001143, diff.dev.:1.1102230246251565e-13
+└ @ GLM ~/.julia/dev/GLM/src/glmfit.jl:418
+```
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -260,7 +260,7 @@ deviance.
 ## Debugging failed fits
 In the rare cases when a fit of a generalized linear model fails, it can be useful
 to enable more output from the fitting steps. This can be done through
-the Julia logging mechanism by setting `ENV[JULIA_DEBUG] = GLM`. Enabling debug output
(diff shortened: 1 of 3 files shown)


docs/src/index.md
```diff
@@ -256,3 +256,17 @@ of a ```LinPred``` type use a default step factor of 1.  The value of
 ```linpred``` is the argument to the ```updatemu``` method for
 ```ModResp``` types.  The ```updatemu``` method returns the updated
 deviance.
+
+## Debugging failed fits
+In the rare cases when a fit of a generalized linear model fails, it can be useful
+to enable more output from the fitting steps. This can be done through
+the Julia logging mechanism by setting `ENV[JULIA_DEBUG] = GLM`. Enabling debug output
```
ararslan, line 263:
```suggestion
the Julia logging mechanism by setting `ENV["JULIA_DEBUG"] = GLM`. Enabling debug output
```
I can't recall whether the RHS also needs quotes, nor whether this needs to be specified before Julia and/or the package is loaded.

ararslan (commented):
Looks good modulo one suggestion.

Do you think it would be worth doing a proper deprecation for the `verbose` keyword argument?

docs/src/index.md
andreasnoack: This version should be good.

ararslan (approved):
LGTM, thanks! Could possibly be worthwhile to add a deprecation message but I defer to you on that.

andreasnoack:
@ararslan I've added a deprecation warning but I'm wondering if I should make it an `@info` instead of using `depwarn`. I think deprecation warnings have become mostly useless after they aren't enabled by default as nobody sees them anymore. 

I've also adjusted the debug messages to rely on the keyword functionality in the logging messages instead of interpolating the values into the string.

src/glmfit.jl
```diff
@@ -451,7 +450,10 @@ function StatsBase.fit!(m::AbstractGLM;
         Base.depwarn("'convTol' argument is deprecated, use `atol` and `rtol` instead", :fit!)
         rtol = kwargs[:convTol]
     end
-    if !issubset(keys(kwargs), (:maxIter, :minStepFac, :convTol))
+    if haskey(kwargs, :verbose)
+        Base.depwarn("""`verbose` argument is deprecated, use `ENV["JULIA_DEBUG"]=GLM` instead.""", :negbin)
```
ararslan, line 454: I assume you mean
```suggestion
        Base.depwarn("""`verbose` argument is deprecated, use `ENV["JULIA_DEBUG"]=GLM` instead.""", :fit!)
```
?

src/glmfit.jl
```diff
@@ -490,7 +491,10 @@ function StatsBase.fit!(m::AbstractGLM,
...
```
ararslan, line 495: Here too?
```suggestion
        Base.depwarn("""`verbose` argument is deprecated, use `ENV["JULIA_DEBUG"]=GLM` instead.""", :fit!)
```

ararslan (commented):
The wrong function is getting associated with the deprecation in a couple of places but otherwise looks great! Good to merge once the suggested changes are incorporated.

src/glmfit.jl
andreasnoack: indeed

ararslan:
> I've added a deprecation warning but I'm wondering if I should make it an `@info` instead of using `depwarn`. I think deprecation warnings have become mostly useless after they aren't enabled by default as nobody sees them anymore.

IMO it's best to use the functionality that's designed for this purpose, even if it means people are less likely to see it. It's a pet peeve of mine when deprecations are implemented in a way that bypasses the deprecation functionality because it can be difficult to silence when the caller isn't your own code.

> I've also adjusted the debug messages to rely on the keyword functionality in the logging messages instead of interpolating the values into the string.

Very nice!

PR #580, opened 2024-11-25
Disable previews

andreasnoack:
The current implementation seems to check in each preview causing the repo size to grow with each docs run so disabling until a different solution is available.

@palday I don't think it's a good idea to have this enabled with the current implementation. In a different context, I just noticed that the Catalyst.jl repo is ~3GB because of docs previews.

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -16,5 +16,6 @@ makedocs(
 
 deploydocs(
     repo   = "github.com/JuliaStats/GLM.jl.git",
-    push_preview = true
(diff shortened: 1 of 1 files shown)


palday:
The Documenter docs provide a workflow for cleaning up previews: https://documenter.juliadocs.org/dev/man/hosting/#Cleaning-up-gh-pages

You can also safely delete the entire history of the gh-pages branch.

andreasnoack:
Superseded by https://github.com/JuliaStats/GLM.jl/pull/583

PR #588, opened 2025-03-21
fixed docstring for `GLM.inverselink`, third return is (1 - mu) not the variance function

ajinkya-k:
Sorry this should have been a part of #587, but i missed this change


ajinkya-k:
related to #586

andreasnoack (approved):

PR #596, opened 2025-04-09
Make sparse support an extension

andreasnoack:
This will allow users to avoid the dependency on SparseArrays and therefore GPL.


andreasnoack:
@ararslan The PRs are ordered such that this one should go in first, then #598, and finally, #599 . #597 is independent from the other changes.

nalimilan (approved):

PR #599, opened 2025-04-09
Switch to method=:qr as the default

### LINKED ISSUE #573: Defaulting to the slightly slower but more stable QR method?
Created from #500 to have a place to discuss if we want to make this change. This issue is closely related to #426 and #375.

I think I'm in favor of switching to using QR as the default. Thoughts?

cc @nalimilan @palday 

Blocked by https://github.com/JuliaStats/GLM.jl/issues/595

andreasnoack:
Closes #573 
Closes https://github.com/JuliaStats/GLM.jl/issues/426
Closes https://github.com/JuliaStats/GLM.jl/issues/375


test/runtests.jl
```diff
@@ -348,7 +348,7 @@ end
 @testset "Linear model with QR method and NASTY data" begin
     x = [1, 2, 3, 4, 5, 6, 7, 8, 9]
     nasty = DataFrame(X = x, TINY = 1.0E-12*x)
-    mdl = lm(@formula(X ~ TINY), nasty; method=:qr)
+    mdl = lm(@formula(X ~ TINY), nasty)
```
nalimilan, line 351: Have you checked that Cholesky is covered well enough after this PR? Haven't some tests for QR become redundant now that it's the default?

src/GLM.jl
```diff
@@ -99,24 +99,24 @@ module GLM
...
           (however, the exact selection of columns identified as redundant is not guaranteed).
-        - `method::Symbol=:cholesky`: Controls which decomposition method to use.
-          If `method=:cholesky` (the default), then the `Cholesky` decomposition method will be used.
-          If `method=:qr`, then the `QR` decomposition method (which is more stable
-          but slower) will be used.
-        - `wts::Vector=similar(y,0)`: Prior frequency (a.k.a. case) weights of observations.
+        - `method::Symbol`: Controls which decomposition method to use.
+          If `method=:qr` (the default), then the `QR` decomposition method will be used.
+          If `method=:cholesky`, then the `Cholesky` decomposition method (which is faster
+          but less accurate) will be used.
```
nalimilan, line 111: Less robust too?

test/runtests.jl
andreasnoack: Yes. Most of the testsets loop through both methods.

src/GLM.jl
andreasnoack: I thought is was a bit unclear two robust meant in this context so decided that it was better to focus on accuracy.

nalimilan (approved):

src/GLM.jl
```diff
@@ -98,24 +98,24 @@ module GLM
...
```
ararslan, line 110: Could be worth giving some more color on the tradeoffs here, but maybe too much info? Idk
```suggestion
          If `method=:cholesky`, then the `Cholesky` decomposition method will be used.
          The Cholesky decomposition is faster and more computationally efficient than
          QR, but is less numerically stable and thus may fail or produce less accurate
          estimates for some models.
```

andreasnoack:
I had to adjust one more testset that was testing things specific to Cholesky, e.g. that the factorization was a Cholesky. Otherwise no changes were needed so I'll go ahead and merge.

PR #600, opened 2025-04-10
Adjust tolerance when computing rank of R matrix

andreasnoack:
When fitting based on a QRPivoted factorization, the tolerance used for determining the rank of R should reflect the size of the matrix of which the QR factorization was computed. Hence, a custom rtol based on that first dimension of X is passed to rank.

For consistency, all rank determinations are not handled via rank_linpred as the LinPred object has all the relevant information.

This issue was revealed by #599


nalimilan:
Would it make any sense to adjust this tolerance depending on the value of `rtol` passed by the user?

andreasnoack:
Possibly, but it always complicates things a bit when things touch the interface so I'd prefer to keep that as a separate issue. I also think it would be pretty rare that users would like fine tune it.

nalimilan (approved):

PR #603, opened 2025-04-14
Enable JuliaFormatter

nalimilan:
Add a JuliaFormatter style, apply it to the code base, and enable a GitHub workflow to add suggestions to PRs to follow the style. Inspired by discussions at https://github.com/JuliaStats/HypothesisTests.jl/pull/299.

EDIT: the second commit shows what the default style looks like. It's a bit different from the majority of the current code regarding alignment of arguments.


devmotion:
Maybe even if we decide to use the default style, a `.JuliaFormatter.toml` file (in that case empty) should be added to the repo. `JuliaFormatter.format` traverses the file path upwards until it finds (or in the end doesn't find) a `.JuliaFormatter.toml` file: https://domluna.github.io/JuliaFormatter.jl/stable/config/ So in principle a configuration higher up in the file tree could cause the formatter to format GLM with a different style.

PR #487, opened 2022-07-15
Taking weighting seriously

gragusa:
This PR addresses several problems with the current GLM implementation. 

**Current status**
In master, GLM/LM only accepts weights through the keyword `wts`. These weights are implicitly frequency weights. 

**With this PR**
FrequencyWeights, AnalyticWeights, and ProbabilityWeights are possible. The API is the following

```julia
## Frequency Weights
lm(@formula(y~x), df; wts=fweights(df.wts)
## Analytic Weights
lm(@formula(y~x), df; wts=aweights(df.wts)
## ProbabilityWeights
lm(@formula(y~x), df; wts=pweights(df.wts)
```
The old behavior -- passing a vector `wts=df.wts` is deprecated and for the moment, the array os coerced `df.wts` to FrequencyWeights.

To allow dispatching on the weights, `CholPred` takes a parameter `T<:AbstractWeights`. The unweighted LM/GLM has UnitWeights as the parameter for the type.  

This PR also implements `residuals(r::RegressionModel; weighted::Bool=false)` and `modelmatrix(r::RegressionModel; weighted::Bool = false)`. The new signature for these two methods is pending in StatsApi. 

There are many changes that I had to make to make everything work. Tests are passing, but some new feature needs new tests. Before implementing them, I wanted to ensure that the approach taken was liked.

I have also implemented `momentmatrix`, which returns the estimating function of the estimator. I arrived to the conclusion that it does not make sense to have a keyword argument `weighted`. Thus I will amend  [https://github.com/JuliaStats/StatsAPI.jl/pull/16](https://github.com/JuliaStats/StatsAPI.jl/pull/16) to remove such a keyword from the signature. 

*Update*

I think I covered all the suggestions/comments with [this](https://github.com/JuliaStats/GLM.jl/pull/487#discussion_r958907121) exception as I have to think about it. Maybe this can be addressed later. The new standard errors (the one for `ProbabilityWeights`) also work in the rank deficient case (and so does `cooksdistance`).  

Tests are passing and I think they cover everything that I have implemented. Also, added a section in the documentation about using `Weights` and updated `jldoc` with the new signature of `CholeskyPivoted`. 

To do:
- [X] Deal with weighted standard errors with rank deficient designs
- [X] Document the new API
- [X] Improve testing

Closes https://github.com/JuliaStats/GLM.jl/issues/186.

PR #619, opened 2025-12-17
Deprecate passing non-`FrequencyWeights` to `wts`

nalimilan:
This will allow throwing an error in 2.0, so that we don't have to keep allowing this behavior until 3.0. With the new support for other weights types (#487), treating vectors as frequency weights would be dangerous and deprecation warnings are not always printed.

EDIT: note that this PR is against the v1 branch, destined for a 1.10 or 1.9.1 release

Cc: @gragusa

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "GLM"
 uuid = "38e38edf-8417-5370-95a0-9cbb8c7f171a"
-version = "1.9.0"
+version = "1.9.1"
 
 [deps]
 Distributions = "31c24e10-a181-5473-b8eb-7969acd0382f"


nalimilan:
Ah probably you were too fast and read the description right before I edited it. ;-)
> EDIT: note that this PR is against the v1 branch, destined for a 1.10 or 1.9.1 release

To develop a bit: master has already many breaking changes that can only go into 2.0, here I just want to add a deprecation warning to the 1.x series that will be replaced with an error in 2.0.

devmotion:
Ah, thank you! I had missed this crucial part 🙂

src/lm.jl
```diff
@@ -123,31 +123,38 @@ const FIT_LM_DOC = """
 
 """
     fit(LinearModel, formula, data, allowrankdeficient=false;
-       [wts::AbstractVector], dropcollinear::Bool=true)
+       [wts::FrequencyWeights], dropcollinear::Bool=true)
     fit(LinearModel, X::AbstractMatrix, y::AbstractVector;
-        wts::AbstractVector=similar(y, 0), dropcollinear::Bool=true)
+        wts::FrequencyWeights=fweights(Float64[]), dropcollinear::Bool=true)
```
devmotion: Shouldn't this be

```suggestion
        wts::FrequencyWeights=fweights(similar(y, 0)), dropcollinear::Bool=true)
```

to keep it consistent with the current implementation? Hardcoding `Float64` would mean that there's likely unintended promotion to `Float64` for `Float32`-user input

src/lm.jl
```diff
@@ -123,31 +123,38 @@ const FIT_LM_DOC = """
...
 
 Fit a linear model to data.
 
 $FIT_LM_DOC
 """
 function fit(::Type{LinearModel}, X::AbstractMatrix{<:Real}, y::AbstractVector{<:Real},
              allowrankdeficient_dep::Union{Bool,Nothing}=nothing;
-             wts::AbstractVector{<:Real}=similar(y, 0),
+             wts::AbstractVector{<:Real}=fweights(Float64[]),
```
devmotion:
```suggestion
             wts::AbstractVector{<:Real}=fweights(similar(y, 0)),
```

src/GLM.jl
```diff
@@ -66,6 +67,9 @@ module GLM
         predict,        # make predictions
         ftest           # compare models with an F test
 
+        fweights         # frequency weights short constructor
+        FrequencyWeights # frequency weights type
```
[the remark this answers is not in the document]
nalimilan: https://github.com/JuliaStats/GLM.jl/pull/487 also reexports them. This makes sense to me as otherwise people will have to load StatsBase manually every time they want to use this argument, which seems weird and annoying.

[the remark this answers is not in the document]
nalimilan: I've dropped the export.

src/glmfit.jl
```diff
@@ -579,6 +579,12 @@ function fit(::Type{M},
         throw(DimensionMismatch("number of rows in X and y must match"))
     end
 
+    wts isa FrequencyWeights ||
+        Base.depwarn("Passing weights as vector is deprecated in favor of explicitly using " *
+                    "`FrequencyWeights` from StatsBase. " *
+                    "Proceeding by coercing `wts` to `FrequencyWeights`.",
+                    :fit)
```
devmotion: Maybe enforce the deprecation warning?
```suggestion
                    :fit; force = true)
```

src/lm.jl
```diff
@@ -123,31 +123,38 @@ const FIT_LM_DOC = """
...
         @warn "Positional argument `allowrankdeficient` is deprecated, use keyword " *
               "argument `dropcollinear` instead. Proceeding with positional argument value: $allowrankdeficient_dep"
         dropcollinear = allowrankdeficient_dep
     end
+
+    wts isa FrequencyWeights ||
+        Base.depwarn("Passing weights as vector is deprecated in favor of explicitly using " *
+                     "`FrequencyWeights` from StatsBase. " *
+                     "Proceeding by coercing `wts` to `FrequencyWeights`.",
+                     :fit)
```
devmotion:
```suggestion
                     :fit; force = true)
```

src/glmfit.jl
```diff
@@ -464,7 +464,7 @@ end
 
 function StatsBase.fit!(m::AbstractGLM,
                         y;
-                        wts=nothing,
+                        wts::Union{AbstractVector{<:Real}, Nothing}=nothing,
```
devmotion: Maybe this should be restricted to

```suggestion
                        wts::Union{AbstractVector{<:Integer}, Nothing}=nothing,
```

(see https://github.com/JuliaStats/StatsBase.jl/pull/250#discussion_r2649569917)?

src/glmfit.jl
nalimilan: That deprecation isn't particularly problematic (if you missed it, we will print an explicit error message for some time in 2.x I think), so I wouldn't force it. Users who want to keep using GLM 1.x without changing their code immediately should be able to do that IMO.

devmotion (approved):

Issue #617, opened 2025-12-15
`dof()` incorrect for `NegativeBinomial()` and `Geometric()`


gragusa:
The `dof()` method for GeneralizedLinearModel incorrectly counts the dispersion parameter θ for NegativeBinomial(θ) and Geometric() distributions, even though θ is fixed (not estimated). This causes `dof()` to be off by 1 (too high) and the AIC/BIC  to be inflated.

For distributions where dispersion IS estimated (Normal, Gamma, InverseGaussian), Julia's dof() correctly matches `R`, and AIC agrees. The bug only affects `NegativeBinomial` and `Geometric`.

Here a comparison between `GLM` and `R` (using `MASS` package)

| Distribution    | dof (Jl) | AIC (Jl) | loglik_df (R) | AIC (R) |
  |-----------------|----------|----------|---------------|---------|
  | Geometric       | 8        | 1112.74  | 7             | 1110.74 |
  | NegBinomial(2)  | 8        | 1122.52  | 7             | 1120.52 |

The cause is in `src/glmfit.jl`
```julia
function dof(x::GeneralizedLinearModel)
    modelrank = linpred_rank(x.pp)
    dispersion_parameter(x.rr.d) ? modelrank + 1 : modelrank
end
```
The issue is that `dispersion_parameter()` returns `true` for `NegativeBinomial` and `Geometric`, but their θ parameter is fixed by the user, not estimated from the data. The +1 should only apply when dispersion is actually being estimated.

Since #485 is almost ready, I propose tackling this issue after it is merged.  In #485, `dispersion_parameter()` is used in the calculation of the dispersion parameter for weighted models, and the best option is a special handling in dof() to not count fixed θ for these distributions.

nalimilan:
See [#624](https://github.com/JuliaStats/GLM.jl/pull/624).

andreasnoack:
As mentioned in #624, I think the two cases here are separate. `Geometric` is just wrong, but I don't think our negative binomial is. The `dof` value should just be the number of estimated parameters. That also seems to be the definition of `MASS::glm.nb`. I suppose R values in the table above are based on a fixed, and not an estimated, shape paramter?
```R
> ft.fixed <- glm(Days ~ Eth + Sex + Age + Lrn, quine, family=negative.binomial(2, link="log"))
> ft.estimated <- glm.nb(Days ~ Eth + Sex + Age + Lrn, quine, link="log")
> logLik(ft.fixed)
[...]

PR #624, opened 2025-12-24
Stop considering that Negative Binomial has a dispersion parameter

### LINKED ISSUE #564: `dispersion_parameter(::Geometric)` incorrectly returns true
https://github.com/JuliaStats/GLM.jl/blob/0e145bd3e7b9f2db5a5137c57b1d82efee8d5d95/src/glmtools.jl#L484-L500

The only parameter of the geometric distribution is the success probability.

nalimilan:
This is just incorrect. Fix tests to reflect correct degrees of freedom and therefore AIC, BIC and standard errors.

This requires adapting how `negbin` works as we need to count one additional degree of freedom due to the shape parameter (distinct from the dispersion, which is 1) being estimated from the data. Do this by wrapping the shape parameter in an `Estimated` type that we use to detect that degrees of freedom need adjusting.

Also avoid defining `dispersion_parameter` fallback for any distribution: it's clearer for users to get a `MethodError` so that they have to define the appropriate method themselves, than silently returning a wrong value.

Fixes #564, #617.

--------

I have tentatively implemented the first of three possible approaches I see to the `negbin` issue to illustrate how it looks:
1. Wrap the shape parameter of `NegativeBinomial` in an `Estimated` type: I find this is somewhat elegant. The disadvantage is that users can't really work with that parameter if they extract it. We could define more methods on this type but that would be weird anyway. We'd better provide an accessor which would unwrap it, which is needed anyway.
2. Add a field to `GeneralizedLinearModel` (or one of its component) indicating that one parameter was estimated. This has the advantage of being super simple. The drawback is that its a bit ad-hoc but in practice it probably doesn't matter.
3. Make `negbin` return a `NegativeBinomialModel <: GLMLikeModel <: LinPredModel`. This has the advantage of clarity (in R also `glm.nb` returns an object of a different class) since these are not really GLMs. In terms of implementation I'm not sure whether it would be easy (just replace `AbstractGLM` with `GLMLikeModel` for most methods) or not.

Opinions on the best approach?

Note that R's glm seems broken with `negative.binomial` as [it estimates a dispersion parameter](https://stats.stackexchange.com/questions/70619/dispersion-in-summary-glm) even if it's 1 by theory. You have to do `summary(m, dispersion=1)` to get the correct standard errors. Yet `?negative.binomial` claims it's meant do be used with `glm` with a fixed shape.


andreasnoack:
The two cases here are different, and I think it is best to handle them in separate PRs. `Geometric` only has a single parameter while `NegativeBinomial` has two parameters where the second isn't a dispersion parameter in the GLM sense. I agree that we should not have a fallback but only define `dispersion_parameter` for the cases we want to support. I think the better way to handle the NegativeBinomial case is to introduce a new type just for NegativeBinomial, somewhat like what MASS does, and I think it should just be a direct subtype of `LinPredModel`, i.e. no need for a new abstract type.

nalimilan:
It's hard to fix the two cases completely separately as we have tests checking that geometric and negative binomial give the same results, but we can leave some tests as broken temporarily. Let's start by fixing `Geometric`, which is the simplest: #633.

src/glmfit.jl
```diff
@@ -668,14 +676,20 @@ function glm(formula::FormulaTerm, data, d::UnivariateDistribution,
...
 
 """
+    dispersion(m::LinearModel, sqr::Bool=false)
     dispersion(m::AbstractGLM, sqr::Bool=false)
 
 Return the estimated dispersion (or scale) parameter for a model's distribution,
 generally written σ for linear models and ϕ for generalized linear models.
-It is, by definition, equal to 1 for the Bernoulli, Binomial, and Poisson families.
+
+It is, by definition, equal to 1 for the Bernoulli, Binomial, Geometric, Negative
```
andreasnoack: I don't think it is correct to say that it is one for the negative binomial. As I understand McCullagh and Nelder, the proper dispersion parameter is only defined in the case when the variance can be written
```math
\mathrm{Var}(Y) = \frac{\phi}{m}\mathrm{Var}(\mu)
```
in which case $\phi$ is called the dispersion parameter. For the negative binomial, the variance is
```math
\mathrm{Var}(Y) = \mu + \mu^2/k
```
so $k$ isn't a *dispersion* parameter according to their definition, but I wouldn't say it is one either. It happens to be the case the Fisher information is $\phi(X^T W X)^{-1}$ for the proper GLMs and for negative binomial, it is just $(X^T W X)^{-1}$ so you can reuse the code for the GLMs if you pass $\phi=1$ but that is just a computational convenience and not because the negative binomial fits into the framework.

src/lm.jl
```diff
@@ -208,6 +208,13 @@ function lm(f::FormulaTerm, data;
     return fit(LinearModel, f, data; wts, dropcollinear, method, contrasts)
 end
 
+"""
+    dof(model::GeneralizedLinearModel)
```
andreasnoack:
```suggestion
    dof(model::LinearModel)
```

src/glmfit.jl
```diff
@@ -394,7 +394,15 @@ function nullloglikelihood(m::GeneralizedLinearModel)
...
 end
 
-dof(obj::GeneralizedLinearModel) = linpred_rank(obj) + dispersion_parameter(obj.rr.d)
+"""
+    dof(model::AbstractGLM)
+
+For generalized linear models, consumed degrees of freedom correspond
+to the number of estimated coefficients, plus one for the estimated
+dispersion parameter if the distribution is `Gamma`, `InverseGaussian` or `Normal`
+(see [GLM.dispersion_parameter](@ref)).
```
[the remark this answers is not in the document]
andreasnoack: Are you sure it is useful to loosen the signature from GeneralizedLinearModel to AbstractGLM? Isn't the only type that is an `AbstractGLM` but not a `GeneralizedLinearModel` the new `NegativeBinomialModel` for which there is a separate model?

src/glmfit.jl
nalimilan: So say "Negative Binomial (with fixed θ)"?

src/negbinfit.jl
```diff
@@ -1,3 +1,25 @@
+mutable struct NegativeBinomialModel{G<:GlmResp,L<:LinPred} <: AbstractGLM
```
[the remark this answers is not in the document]
nalimilan: With a `getproperty` override to delegate field accesses? Doesn't really seem simpler in the end?

PR #625, opened 2025-12-26
Print model information above coefficient table in `show`

### LINKED ISSUE #623: Print more information in `show`
I suggest we give more information when printing models than just the formula and coefficient table. Changing the result of `show` may not be considered as breaking, but 2.0 would be a good time to do this. I would suggest printing at least these indicators which [are present](https://stats.oarc.ucla.edu/other/dae/) in R, Stata and SAS:
- number of observations (super important in particular to check that you didn't lose too many observations due to missing values)
- R² and adjusted R² (for linear models -- it's more controversial to choose a pseudo-R² for GLMs)
- log-likelihood (for GLMs)
- F-test or LR test
- dispersion parameters for distributions that have one

Other possible indicators to include:
- AIC and AICc
- null log-likelihood

Econometrics.jl seems like a good inspiration:
```julia
Continuous Response Model
Number of observations: 223174
Null Loglikelihood: -6746726.41
Loglikelihood: -345253.18
R-squared: 0.0466
Wald: 860.46 ∼ F(3, 223158) ⟹ Pr > F = 0.0000
Formula: log(Earnings) ~ 1 + Educatn + Age + (Age ^ 2) + absorb(Kids)
Variance Covariance Estimator: OIM
──────────────────────────────────────────────────────────────────────────────────────
                    PE           SE      t-value  Pr > |t|         2.50%        97.50%
──────────────────────────────────────────────────────────────────────────────────────
(Intercept)   8.23837      0.13135      62.721      <1e-99   7.98093       8.49581
Educatn       0.000538588  0.000138739   3.88203    0.0001   0.000266663   0.000810512
Age           0.0389003    0.0068679     5.66408    <1e-07   0.0254394     0.0523613
Age ^ 2      -0.000191192  8.86411e-5   -2.15692    0.0310  -0.000364926  -1.74575e-5
──────────────────────────────────────────────────────────────────────────────────────
```

nalimilan:
This basic information is printed by most other implementations.

Comments welcome.

I haven't included the LR test for now since we don't have a single-model function like `ftest` at the moment and it seems less common to print it than the F-test.

After #624 I'll also show the shape parameter of `NegativeBinomial` distributions.

Fixes #623.

### DIFF
--- a/src/linpred.jl
+++ b/src/linpred.jl
@@ -378,7 +378,8 @@ function show(io::IO, obj::LinPredModel)
     end
     print(io, "Observations: ", lpad(nobs(obj), colwidth-14), sep)
     println(io, "Degrees of freedom: ", lpad(dof(obj), colwidth-20))
-    if wts isa ProbabilityWeights
+    if wts isa ProbabilityWeights ||
+        (wts isa AnalyticWeights && obj isa AbstractGLM && obj.rr.d isa Union{Bernoulli, Binomial})
         print(io, "Log-likelihood: ", lpad("not supported", colwidth-16), sep)
     else
         print(io, "Log-likelihood: ",
--- a/test/analytic_weights.jl
+++ b/test/analytic_weights.jl
@@ -30,6 +30,8 @@ itr = Iterators.product((:qr, :cholesky), (true, false))
                  method=dmethod, dropcollinear=drop)
     X = hcat(ones(length(df.y)), df.x1, df.x2)
     model_2 = lm(X, y; wts=aweights(df.w), method=dmethod, dropcollinear=drop)
+    test_show(model_0)
+    test_show(model_1)
     @test ftest(model_1).fstat ≈ 1.551275 rtol = 1e-05
     @test ftest(model_2) === ftest(model_1)
     @test ftest(model_0, model_1).fstat[2] ≈ 1.7860438 rtol = 1e-05
@@ -42,6 +44,7 @@ end
 
     model = glm(@formula(y ~ 1 + x1 + x2), df, Binomial(), LogitLink(), wts=aweights(df.w),
                 method=dmethod, dropcollinear=drop, atol=1e-08, rtol=1e-08)
+    test_show(model)
     @test deviance(model) ≈ 39.58120350785813 rtol = 1e-06
     @test coef(model) ≈ [0.6333582770515337, 1.8861277804531265, 18.61281712203539] rtol = 1e-06
     @test stderror(model) ≈ [0.9021013750843575, 2.063002891039618, 2337.217357530545] rtol = 1e-07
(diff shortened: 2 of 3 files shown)


nalimilan:
But then we would use exactly the same space as the current output, we would just refuse to show classic Wald standard errors in that space. That would be weird IMO, especially if we expect people to call `coeftable` manually right after fitting the model to get the inference they want (just imagine how many examples would look in the docs).

andreasnoack:
The design of the package suggests that `coeftable` *should* be called on the fitted object when inference is needed. The relevant arguments for inference are exposed in the `coeftable` function, not in `fit`, so, in my view, the current behavior is weird. As explained above, if your position is that the current output should be part of the show method for the fit itself, then the struct for the model fit should store these quantities, and the `fit` method should expose all keyword arguments for inference. I prefer the current design, and it is too late to change for 2.0 anyway.

nalimilan:
I mean, have you ever used coefficients without their inference? People are going to call `coeftable` right after fitting anyway (to see Wald tests or different tests once we support that).

If we print coefficients in `show` and not their standard erros, the typical use case would look like this, right?
```julia
julia> m = lm(@formula(SR ~ Pop15 + Pop75 + DPI + DDPI), LifeCycleSavings)
LinearModel:
SR ~ 1 + Pop15 + Pop75 + DPI + DDPI

Observations:                        50  Degrees of freedom:                   6
Log-likelihood:                 -135.10  RSS:                             650.71
RSE:                               3.80  F-test:                          0.0008
R²:                              0.3385  Adjusted R²:                     0.2797

Coefficients:
─────────────────────────
                    Coef.
─────────────────────────
(Intercept)  28.5661     
Pop15        -0.461193  
Pop75        -1.6915    
DPI          -0.000336902
DDPI          0.409695  
─────────────────────────

julia> coeftable(m)
─────────────────────────────────────────────────────────────────────────────────
                    Coef.   Std. Error      t  Pr(>|t|)    Lower 95%    Upper 95%
─────────────────────────────────────────────────────────────────────────────────
(Intercept)  28.5661       7.35452       3.88    0.0003  13.7533      43.3788
Pop15        -0.461193     0.144642     -3.19    0.0026  -0.752518    -0.169869
Pop75        -1.6915       1.0836       -1.56    0.1255  -3.87398      0.490983
DPI          -0.000336902  0.000931107  -0.36    0.7192  -0.00221225   0.00153844
DDPI          0.409695     0.196197      2.09    0.0425   0.0145336    0.804856
─────────────────────────────────────────────────────────────────────────────────
```

Seems quite redundant to me. That's even worse when you have ten or twenty variables, which is common in my field. Even if we print several coefficients on a single line, that takes space.

AFAICT the only implementation that prints coefficients separate from inference is R's `lm` and `glm`. And even in R, modeling packages generally don't follow that example and do print inference by default.

PR #627, opened 2025-12-26
Allow choosing Student distribution for Wald tests with GLMs

nalimilan:
This is useful in particular because that's what R's `glm` does by default. Python's statsmodels supports this via a `use_t=true` argument.

I'm even tempted to use the Student distribution by default for models with a dispersion parameter as it's less surprising for people trying to compare results with R. @dmbates Is there any particular reason why you chose to always use a Normal distribution (apart from consistency and simplicity)?

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -2615,6 +2615,7 @@ end
     df = DataFrame(; x, y)
     m1 = glm(@formula(y ~ x), df, Normal())
     @test repr(coeftable(m1, test=:t)) == repr(coeftable(lm(@formula(y ~ x), df)))
+    @test_throws ArgumentError coeftable(m1, test=:xx)
 
     m2 = glm(@formula(y ~ x), df, Gamma())
     @test repr(coeftable(m2, test=:t)) == """


nalimilan:
Yeah there's always the tension between providing what people are used to and trying to improve on old habits... But here I was just assuming we keep the current approach with Wald tests for each coefficient in the alternative model. Do you have a preference between the t-statistic and the z-statistic in that case?

nalimilan:
I'm not sure what makes people happy. Maybe what makes R users happy. :sweat_smile:

src/glmfit.jl
```diff
@@ -262,16 +262,36 @@ function GeneralizedLinearModel(rr::GlmResp, pp::LinPred,
...
-    p = 2 * ccdf.(Ref(Normal()), abs.(zz))
-    ci = se * quantile(Normal(), (1 - level) / 2)
+    if test === :z
+        distr = Normal()
+        tstr, pstr = "z", "Pr(>|z|)"
+    elseif test === :t
+        dispersion_parameter(mm.rr.d) ||
+            throw(ArgumentError("`test=:t` is not appropriate for models with " *
+                                "$(nameof(typeof(mm.rr.d))) distribution as " *
+                                "dispersion parameter is not estimated"))
```
andreasnoack: The support for this claim seems rather weak to me. I could only find https://stat.ethz.ch/pipermail/r-help/2006-March/100878.html. Even in the case of known dispersion, the estimator is only asymptotically normal. I know that it is often claimed that the t/F distribution often has better small-sample properties, but I don't see why that argument shouldn't apply equally in the case of known dispersion parameter (with the exception of the Gaussian case where the small-sample distribution is known to be t/F).

nalimilan: Yeah, I was just being cautious because it's not clear why the t distribution would be appropriate here. It seems the only established result is that the ratio asymptotically follows a Normal distribution. The Student distribution only applies to a linear model with the normality assumption. McCullagh & Nelder don't seem to talk about this. I've pushed a commit to allow `test=:t` for all models.

nalimilan:
> I think it might be fine to allow users to get p-values based on the t distribution, but see my inline comment. Personally, I'd be more interested in other ways to compute coefficient tables based on e.g. LRT, profiling, and bootstrapping. That is, of course, for another day, but maybe it has implications for how we should add options to `coeftable` or coetable-like functionality.

Yes it would be nice to allow getting the same table but with p-values from LRT or bootstrap. Maybe `coeftable(m, test=:lr)` and `coeftable(m, test=BasicSampling(1000))` from Bootstrap.jl would be a nice API? Ideally we would allow passing any function which would get passed the model and return statistics, p-values and/or confidence intervals.

nalimilan:
For the reference, previous discussion about covariance matrices happened at #42.

PR #628, opened 2025-12-28
Adjust ftest

andreasnoack:
This commit changes the degrees of freedom shown in the table from the number of estimated parameters (dof) to be the number of degrees of freedom for the model (residual_dof). This makes it possible to easily calculate the F-test statistic from the other quantities in the table with the usual formulas.

Since nobs now returns floats, residual_dof also generally returns floats, so the degrees of freedom parameters are now stored as floats instead of integers.

The commit removes the R-squared quantities. They are not needed and I don't think they add value.

The commit also removes numbering of the rows in the output. I don't think it is helpful.

The example from the docstring changes from 
```julia
────────────────────────────────────────────────────────────────
     DOF  ΔDOF     SSR    ΔSSR      R²     ΔR²        F*   p(>F)
────────────────────────────────────────────────────────────────
[1]    2        3.2292          0.0000
[2]    3     1  0.1283  3.1008  0.9603  0.9603  241.6234  <1e-07
────────────────────────────────────────────────────────────────
```
to
```julia
───────────────────────────────────────────
   SSR  dof    ΔSSR  Δdof        F*   p(>F)
───────────────────────────────────────────
3.2292   11
0.1283   10  3.1008     1  241.6234  <1e-07
───────────────────────────────────────────
```

The test statistic can be calculated as
```julia
julia> (3.2292 - 0.1283)/1/(0.1283/10)
241.6913484021824
```


nalimilan:
Funny, I had made the opposite choice regarding DOF at https://github.com/JuliaStats/GLM.jl/pull/337#discussion_r334281149. I'm not a fan of residual DOF because the number can be very high with large samples, but OK if you prefer that. Though I wouldn't call the column with residual degrees of freedom "dof", as it's the name of the function that returns residual DOF. Confusion is already quite common. Maybe "Res. DOF"? Using upper case would seem more consistent with SSR.

Also I'd rather keep the R², it's interesting to compare models and its magnitude more interpretable than SSR. It's not needed for the F-test, but we probably don't want to add another function just to compare models with their R² (Stata has [`nestreg`](https://www.stata.com/manuals/rnestreg.pdf) which does all of this). At any rate it doesn't hurt.

nalimilan:
Ah, and something to take into account is that we should use the same output for `lrtest` in StatsModels. Not sure whether changing the output would require a breaking release or not...

andreasnoack:
> I'm not a fan of residual DOF because the number can be very high with large samples, but OK if you prefer that.

The idea is that the numbers in the table are the ones that are used in the test statistic. The residual degrees of freedom are the relevant quantities for that.

> Though I wouldn't call the column with residual degrees of freedom "dof", as it's the name of the function that returns residual DOF. Confusion is already quite common.

I don't think `dof` would cause confusion with the change here, since the `dof` column is associated with the SSR, i.e. "Residual". 

> Using upper case would seem more consistent with SSR.

Sum of squares are almost always upper case, but degrees of freedom is a mixed bag. SAS writes "DF" (but uses upper case a lot), Stata writes "df", R writes "Df", and two books in English on my shelf that cover ANOVA write "df" and "d.f." respectively (and SSR/SSE). Hence, I'd actually be in favor of just "df".

> Also I'd rather keep the R², it's interesting to compare models and its magnitude more interpretable than SSR. It's not needed for the F-test

I think it is odd to include in an ANOVA table. The name of the function here is `ftest` after all. No doubt it can be useful to consider multiple quantities when comparing models, but just including the coefficient of determination seems arbitrary. Why not the adjusted version or some information criteria. Hence, I'd actually be in favor of a separate function for more general model comparisons.

> Ah, and something to take into account is that we should use the same output for lrtest in StatsModels.

Hadn't considered that so I'll take a look.

src/ftest.jl
```diff
@@ -5,11 +5,10 @@ struct SingleFTestResult
     pval::Float64
 end
 
-mutable struct FTestResult{N}
+struct FTestResult{N}
     nobs::Int
     ssr::NTuple{N,Float64}
-    dof::NTuple{N,Int}
-    r2::NTuple{N,Float64}
+    dof::NTuple{N,Float64}
```
nalimilan, line 11: Better rename the field, if users access it it will be clearer to get an error than silently get the wrong value.
```suggestion
    resdof::NTuple{N,Float64}
```

PR #638, opened 2026-01-11
Fix handling of zero weights

nalimilan:
Observations with zero weights have to be skipped for probability and analytic weights. Values have been checked against R. Avoid duplication by using a single method for all `LinPredModel`s (though this means we cannot call `nobs` from functions which don't get passed the full model).

Also fix name of internal function `link`, which actually returns the distribution.

Improves on #487. Cc: @gragusa

### DIFF
--- a/src/glmfit.jl
+++ b/src/glmfit.jl
@@ -25,7 +25,7 @@ struct GlmResp{V<:FPVector,D<:UnivariateDistribution,L<:Link,W<:AbstractWeights}
     wrkresid::V
 end
 
-link(rr::GlmResp) = rr.d
+distr(rr::GlmResp) = rr.d
 
 function GlmResp(y::V, d::D, l::L, η::V, μ::V, off::V, wts::W) where {V<:FPVector,D,L,W}
     n = length(y)
@@ -79,7 +79,7 @@ end
 function deviance(r::GlmResp)
     wts = weights(r)
     d = sum(r.devresid)
-    return wts isa ProbabilityWeights ? d * nobs(r) / sum(wts) : d
+    return wts isa ProbabilityWeights ? d * sum(!iszero, wts) / sum(wts) : d
 end
 
 weights(r::GlmResp) = r.wts
@@ -342,7 +342,7 @@ function loglikelihood(m::AbstractGLM)
     y = r.y
     mu = r.mu
     wts = weights(r)
-    d = link(r)
+    d = distr(r)
     ll = zero(eltype(mu))
     N = length(y)
     δ = deviance(r)
@@ -860,7 +860,8 @@ function checky(y, d::Binomial)
 end
 
 function nobs(r::GlmResp{V,D,L,W}) where {V,D,L,W<:AbstractWeights}
-    return oftype(sum(one(eltype(weights(r)))), length(r.y))
+    n = W<:ProbabilityWeights ? sum(!iszero, weights(r)) : length(r.y)
+    return oftype(sum(one(eltype(weights(r)))), n)
 end
 nobs(r::GlmResp{V,D,L,W}) where {V,D,L,W<:FrequencyWeights} = sum(r.wts)
 
@@ -884,7 +885,7 @@ end
 function momentmatrix(m::GeneralizedLinearModel)
     X = modelmatrix(m; weighted=false)
     r = varstruct(m)
-    if link(m) isa Union{Gamma,InverseGaussian}
+    if distr(m) isa Union{Gamma,InverseGaussian}
         r .*= sum(working_weights(m)) / sum(abs2, r)
     end
     return Diagonal(r) * X
--- a/src/glmfit.jl
+++ b/src/glmfit.jl
@@ -79,7 +79,7 @@ end
 function deviance(r::GlmResp)
     wts = weights(r)
     d = sum(r.devresid)
-    return wts isa ProbabilityWeights ? d * sum(!iszero, wts) / sum(wts) : d
+    return wts isa ProbabilityWeights ? d * nobs(r) / sum(wts) : d
 end
 
 weights(r::GlmResp) = r.wts
@@ -858,12 +858,6 @@ function checky(y, d::Binomial)
     end
     return nothing
 end
-
-function nobs(r::GlmResp{V,D,L,W}) where {V,D,L,W<:AbstractWeights}
-    n = W<:ProbabilityWeights ? sum(!iszero, weights(r)) : length(r.y)
-    return oftype(sum(one(eltype(weights(r)))), n)
-end
-nobs(r::GlmResp{V,D,L,W}) where {V,D,L,W<:FrequencyWeights} = sum(r.wts)
 
 function residuals(r::GlmResp; weighted::Bool=false)
     y, η, μ = r.y, r.eta, r.mu
--- a/src/linpred.jl
+++ b/src/linpred.jl
@@ -316,7 +316,7 @@ function vcov(x::LinPredModel)
         s = nobs(x) / (nobs(x) - 1)
         mm = momentmatrix(x)
         A = invloglikhessian(x)
-        if link(x) isa Union{Gamma,InverseGaussian}
+        if distr(x) isa Union{Gamma,InverseGaussian}
             r = varstruct(x)
             A ./= sum(working_weights(x)) / sum(abs2, r)
         end
@@ -326,7 +326,7 @@ function vcov(x::LinPredModel)
     end
 end
 
-link(x::LinPredModel) = link(x.rr)
+distr(x::LinPredModel) = distr(x.rr)
 
 function _vcov(pp::LinPred, Z::AbstractMatrix, A::AbstractMatrix)
     if linpred_rank(pp) < size(Z, 2)
@@ -434,8 +434,19 @@ end
 For linear and generalized linear models, return the number of rows when
 the model is unweighted or uses analytical or probability weights.
 If the model uses frequency weights, return the sum of weights.
+Rows with zero weights are not counted.
 """
-nobs(obj::LinPredModel) = nobs(obj.rr)
+function nobs(m::LinPredModel)
+    wts = weights(m)
+    n = if wts isa UnitWeights
+        length(wts)
+    elseif wts isa FrequencyWeights
+        sum(wts)
+    elseif wts isa Union{ProbabilityWeights,AnalyticWeights}
+        sum(!iszero, wts)
+    end
+    return oftype(sum(one(eltype(wts))), n)
+end
 
 weights(m::LinPredModel) = weights(m.rr)
 weights(pp::LinPred) = pp.wts
--- a/src/linpred.jl
+++ b/src/linpred.jl
@@ -436,14 +436,16 @@ the model is unweighted or uses analytical or probability weights.
 If the model uses frequency weights, return the sum of weights.
 Rows with zero weights are not counted.
 """
-function nobs(m::LinPredModel)
-    wts = weights(m)
+nobs(obj::LinPredModel) = nobs(obj.rr)
+
+function nobs(rr::ModResp)
+    wts = weights(rr)
     n = if wts isa UnitWeights
         length(wts)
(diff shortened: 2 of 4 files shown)


test/runtests.jl
```diff
@@ -2659,3 +2659,49 @@ end
...
+    @test coef(lmod) ≈ coef(glmod) ≈ [3.6707317073170724, 0.20731707317073195]
+    @test stderror(lmod) ≈ stderror(glmod) ≈ [2.060504744176091, 0.2870314608599428]
+
+    lmod = lm(X, y, wts=pweights(wts))
+    glmod = glm(X, y, Normal(), wts=pweights(wts))
+    @test nobs(lmod) == nobs(glmod) == sum(!iszero, wts)
+    @test dof_residual(lmod) == dof_residual(glmod) == sum(!iszero, wts) - 2
+    @test deviance(lmod) ≈ deviance(glmod) ≈ 24.500813008130084
+    @test coef(lmod) ≈ coef(glmod) ≈ [3.6707317073170724, 0.20731707317073195]
+    @test stderror(lmod) ≈ stderror(glmod) ≈ [1.7617126628715203, 0.23455696986842048]
```
nalimilan: @gragusa I get slightly different standard errors with `svyglm`. Is this expected? Generally in my tests I got exactly the same values up to a least four decimals.

```R
> X = 1:10
> y = c(1, 4, 6, 2, 3, 5, 6, 7, 1, 6)
> wts = c(1, 0, 4, 0, 5, 2, 6, 4, 2, 6)
> svyd <- svydesign(~1, weights=wts, data=data.frame(X=X, y=y, wts=wts))

> summary(svyglm(y ~ X, svyd))

Call:
svyglm(formula = y ~ X, design = svyd)

Survey design:
svydesign(~1, weights = wts, data = data.frame(X = X, y = y, 
    wts = wts))

Coefficients:
            Estimate Std. Error t value Pr(>|t|)  
(Intercept)   3.6707     1.7371   2.113    0.079 .
X             0.2073     0.2313   0.896    0.405  
---

(Dispersion parameter for gaussian family taken to be 3.500116)

```

src/glmfit.jl
```diff
@@ -860,7 +860,8 @@ function checky(y, d::Binomial)
 end
 
 function nobs(r::GlmResp{V,D,L,W}) where {V,D,L,W<:AbstractWeights}
-    return oftype(sum(one(eltype(weights(r)))), length(r.y))
+    n = W<:ProbabilityWeights ? sum(!iszero, weights(r)) : length(r.y)
```
devmotion:
```suggestion
    n = W<:ProbabilityWeights ? count(!iszero, weights(r)) : length(r.y)
```

src/linpred.jl
```diff
@@ -434,8 +434,19 @@ end
...
 """
-nobs(obj::LinPredModel) = nobs(obj.rr)
+function nobs(m::LinPredModel)
+    wts = weights(m)
+    n = if wts isa UnitWeights
+        length(wts)
+    elseif wts isa FrequencyWeights
+        sum(wts)
+    elseif wts isa Union{ProbabilityWeights,AnalyticWeights}
+        sum(!iszero, wts)
```
devmotion:
```suggestion
        count(!iszero, wts)
```

src/lm.jl
```diff
@@ -67,23 +67,18 @@ function deviance(r::LmResp)
             v += abs2(y[i] - mu[i]) * wts[i]
         end
     end
-    return wts isa ProbabilityWeights ? v ./ (sum(wts) / length(y)) : v
+    return wts isa ProbabilityWeights ? v ./ (sum(wts) / sum(!iszero, wts)) : v
```
devmotion:
```suggestion
    return wts isa ProbabilityWeights ? v ./ (sum(wts) / count(!iszero, wts)) : v
```

andreasnoack:
There are some conflicts here that need to be resolved

test/runtests.jl
nalimilan: @gragusa I'd appreciate your input on this one too.

nalimilan:
Done.

test/runtests.jl
gragusa: Sorry for the late reply - I don't get the notifications from this repo (although I am subscribed). 

I think they match to 4 digits (the one used by R to print the coefficient):

```julia
## GLM is master/main
using GLM, DataFrames
X = 1:10
y = [1, 4, 6, 2, 3, 5, 6, 7, 1, 6]
wts = [1, 0, 4, 0, 5, 2, 6, 4, 2, 6]
df = DataFrame(y=y, X=X, wts=wts)
julia> lm1 = lm(@formula(y~X), df, weights=GLM.pweights(df.wts))
LinearModel

y ~ 1 + X

Coefficients:
───────────────────────────────────────────────────────────────────────
                Coef.  Std. Error     t  Pr(>|t|)  Lower 95%  Upper 95%
───────────────────────────────────────────────────────────────────────
(Intercept)  3.67073     1.73707   2.11    0.0675  -0.334964   7.67643
X            0.207317    0.231276  0.90    0.3962  -0.326007   0.740641
───────────────────────────────────────────────────────────────────────
```

The variance returned by `R` on this example is
```r
R> ss$cov.scaled
            (Intercept)           X
(Intercept)   3.0174195 -0.37802382
X            -0.3780238  0.05348872
```
which matches Julia's GMM:
```Julia
julia> @show vcov(ll);
vcov(lm1) = [3.0174195202295895 -0.378023819721369; -0.3780238197213689 0.05348872288847028]
```

gragusa: Actually, the problem is that the deviance reported by Julia and R differs. 
```
- `GLM`: deviance = 30.626, dispersion = 3.828
- `svyglm`: deviance = 24.501, dispersion = 3.500
```
I'm unable to view the code for syvglm for the license. However, the ratio between the two is exactly `n_all/n_nz = 10/8`. The difference seems to be that `syvglm` normalizes the weights by the mean over non-zero observations, while GLM normalizes by the mean over all observations.
  
Let me try to fix this.

nalimilan: Thanks. The "problem" is that this PR changes the result. I get this:
```julia
LinearModel

y ~ 1 + X

Coefficients:
───────────────────────────────────────────────────────────────────────
                Coef.  Std. Error     t  Pr(>|t|)  Lower 95%  Upper 95%
───────────────────────────────────────────────────────────────────────
(Intercept)  3.67073     1.76171   2.08    0.0823  -0.640024   7.98149
X            0.207317    0.234557  0.88    0.4108  -0.366623   0.781257
───────────────────────────────────────────────────────────────────────

julia> vcov(lm1)
2×2 Matrix{Float64}:
  3.10363   -0.388825
 -0.388825   0.055017

julia> deviance(lm1)
24.50081300813008
```

This PR seems more correct to me as `vcov` uses the correct number of observations (skipping zero weights). But R's survey doesn't seem to do the same?

nalimilan: BTW, I had two other questions at https://github.com/JuliaStats/GLM.jl/pull/632#discussion_r2659163598 and https://github.com/JuliaStats/GLM.jl/pull/632#discussion_r2659164865. :-)

gragusa:
I have a PR that fixes these cases against the nl/zerowts branch. What should I do? I do not think I can push it.

test/runtests.jl
nalimilan: Ah I hadn't seen your last post. With this PR I do get the same deviance as `svyglm` (see my post above).

gragusa: The fix to make everything (SE and deviance/dispersion parameter) is to actually use the full sample size `length(y)` instead of `nobs == count(!iszero, w)` and add a missing normalization to `nulldeviance(::LinearModel)`. As mentioned, I have the change locally. What should I do with them?

nalimilan: I think you can make a PR against this PR?

gragusa: I'll try .... 🤞🏽

gragusa: PR open: #648

nalimilan: Hmm so this works but are we sure it's right? :-) Do you think `svyglm` does not exclude observations with zero weights on purpose?

EDIT: and `svyglm` prints a warning when fitting a model with zero-weight observations, so it doesn't seem results are supposed to be trusted in that case.
```R
1: In summary.glm(g) :
  observations with zero weight not used for calculating dispersion
```

gragusa: Again, difficult to say what `syvglm` does internally. This is GPL code, so I'm not able to review it. Having weights equal to zero is a pathological case (observations with zero weights should not exist). The fix now handles this case by giving the same SE and deviance/dispersion as one would obtain by running the same model, while omitting observations with zero weights, which I think is the right thing to do. Throwing a warning is something we should do, since 0 weights may indicate an issue on the model's side.

nalimilan: > The fix now handles this case by giving the same SE and deviance/dispersion as one would obtain by running the same model, while omitting observations with zero weights, which I think is the right thing to do.

This is was my goal with this PR, but I don't think that's the case with https://github.com/JuliaStats/GLM.jl/pull/648 at the moment, right?
Current state of this PR:
```julia
julia> lm1 = lm(@formula(y~X), df, weights=GLM.pweights(df.wts))
[ Warning: pweights is defined in StatsBase and is not public in GLM
LinearModel

y ~ 1 + X

Coefficients:
───────────────────────────────────────────────────────────────────────
                Coef.  Std. Error     t  Pr(>|t|)  Lower 95%  Upper 95%
───────────────────────────────────────────────────────────────────────
(Intercept)  3.67073     1.76171   2.08    0.0823  -0.640024   7.98149
X            0.207317    0.234557  0.88    0.4108  -0.366623   0.781257
───────────────────────────────────────────────────────────────────────

julia> df2 = df[df.wts .> 0,:];

julia> lm1 = lm(@formula(y~X), df2, weights=GLM.pweights(df2.wts))
LinearModel

y ~ 1 + X

Coefficients:
───────────────────────────────────────────────────────────────────────
                Coef.  Std. Error     t  Pr(>|t|)  Lower 95%  Upper 95%
───────────────────────────────────────────────────────────────────────
(Intercept)  3.67073     1.76171   2.08    0.0823  -0.640024   7.98149
X            0.207317    0.234557  0.88    0.4108  -0.366623   0.781257
───────────────────────────────────────────────────────────────────────
```
Current state of #648:
```julia
julia> lm1 = lm(@formula(y~X), df, weights=GLM.pweights(df.wts))
[ Warning: pweights is defined in StatsBase and is not public in GLM
LinearModel

y ~ 1 + X

Coefficients:
───────────────────────────────────────────────────────────────────────
                Coef.  Std. Error     t  Pr(>|t|)  Lower 95%  Upper 95%
───────────────────────────────────────────────────────────────────────
(Intercept)  3.67073     1.73707   2.11    0.0790  -0.579731    7.92119
X            0.207317    0.231276  0.90    0.4046  -0.358596    0.77323
───────────────────────────────────────────────────────────────────────

julia> df2 = df[df.wts .> 0,:];

julia> lm1 = lm(@formula(y~X), df2, weights=GLM.pweights(df2.wts))
LinearModel

y ~ 1 + X

Coefficients:
───────────────────────────────────────────────────────────────────────
                Coef.  Std. Error     t  Pr(>|t|)  Lower 95%  Upper 95%
───────────────────────────────────────────────────────────────────────
(Intercept)  3.67073     1.76171   2.08    0.0823  -0.640024   7.98149
X            0.207317    0.234557  0.88    0.4108  -0.366623   0.781257
───────────────────────────────────────────────────────────────────────
```

If that's useful I can look at the svyglm code for you and/or write to Thomas Lumley (I think he's responsive).

Zero weights should be supported without a warning IMO as they can arise sometimes. For example I have a survey where we have two different definitions of who is in scope, and it's convenient to have two different weighting columns, with zero weights for those who are out of scope in a given definition. We simply need to skip observations with zero weights everywhere and everything should be fine.

